PR #1, opened 2019-02-15
Sample unit test

carlosmunoz:
Add a sample unit test for servlets using Junit 5, Mockito, and Sling's own mock libraries.


carlosmunoz:
Merging per no additional comments.

PR #2, opened 2019-03-01
Ccs 2702

benradey:
Code's not pretty but it works. You can upload a single-file asciidoc and get a preview. (Imports won't work).


src/main/java/com/redhat/pantheon/servlet/AsciidocRenderingServlet.java
```diff
@@ -44,40 +43,25 @@
...
-        service = Servlet.class,
-        property = {
-                "sling.servlet.resourceTypes=pantheon/modules",
-                "sling.servlet.extensions=preview",
-                Constants.SERVICE_DESCRIPTION+"=Servlet which transforms asciidoc content into html",
-                Constants.SERVICE_VENDOR+"=Red Hat Content Tooling team"
-        })
+
+@Component(service = { Servlet.class })
+@SlingServletResourceTypes(
```
carlosmunoz: :+1: here. I was surprised to see how this annotation had not been deprecated as a lot of them were when moving over to the OSGI-specific ones.

src/main/java/com/redhat/pantheon/servlet/AsciidocRenderingServlet.java
```diff
@@ -44,40 +43,25 @@
...
+        extensions="preview")
 @SuppressWarnings("serial")
 public class AsciidocRenderingServlet extends SlingSafeMethodsServlet {
+
+    private static final String ADOC_NODE_NAME = "asciidoc";
+    private static final String CONTENT_NODE_NAME = "jcr:content";
+    private static final String CACHE_NODE_NAME = "cachedContent";
     
-    private final Logger log = LoggerFactory.getLogger(AsciidocRenderingServlet.class);
+    private final Logger log = LoggerFactory.getLogger("com.redhat");
```
carlosmunoz: Why the change to a package name? The class name gives you far more specificity on the log entries.

src/main/java/com/redhat/pantheon/servlet/AsciidocRenderingServlet.java
```diff
@@ -44,40 +43,25 @@
...
- *    @Property(name="sling.servlet.selectors", value="hello"),
- *    @Property(name="sling.servlet.extensions", value="html")
- * })
- */
-@Component(
-        service = Servlet.class,
-        property = {
-                "sling.servlet.resourceTypes=pantheon/modules",
-                "sling.servlet.extensions=preview",
-                Constants.SERVICE_DESCRIPTION+"=Servlet which transforms asciidoc content into html",
```
carlosmunoz: We lost these two properties which will be useful when keeping track of all servlets. I say we add them back.

src/main/java/com/redhat/pantheon/servlet/AsciidocRenderingServlet.java
```diff
@@ -105,16 +88,18 @@ protected void doGet(SlingHttpServletRequest request,
         w.write(html);
     }
 
-    private boolean generatedContentHashDoesntMatch(Resource resource) {
-        String srcContent = resource.getValueMap().get("jcr:content/jcr:data", String.class);
-        String existingHash = resource.getValueMap().get("pant:cachedContent/pant:hash", String.class);
+    private boolean generatedContentHashMatches(Resource resource) {
+        String srcContent = resource.getValueMap().get("asciidoc/jcr:content/jcr:data", String.class);
```
carlosmunoz: I guess this assumes the resource itself is not a file, but instead has an `asciidoc` node under it which is of type `nt:file`?

src/main/java/com/redhat/pantheon/servlet/AsciidocRenderingServlet.java
benradey: Which two properties? The Service Description/Vendor? I wasn't sure what they were used for - i.e., whether they populate anything in a Sling admin console anywhere - so I took them out for simplicity's sake. Do you know if the values actually get used anywhere?

src/main/java/com/redhat/pantheon/servlet/AsciidocRenderingServlet.java
benradey: https://sling.apache.org/documentation/the-sling-engine/servlets.html

I'm not sure how to read your comment about deprecation, but far from being deprecated, this is actually the preferred way to register the servlet according to the doc I linked.

src/main/java/com/redhat/pantheon/servlet/AsciidocRenderingServlet.java
benradey: My logic was that during the course of development, I may need to destroy my sling settings folder as part of each deployment, including loggers I have set up. Rather than needing to re-set up 5 or so loggers each time, it's easier to just do 1.

Having said that, I haven't actually had that problem yet, and I'm also a big fan of "don't solve problems you don't have and may never" so, we can put it back the way it was.

src/main/java/com/redhat/pantheon/servlet/AsciidocRenderingServlet.java
benradey: Yeah, I had to rearrange the data structure a little bit due to the way file upload works in sling+HTL. The OOTB widget creates a node to hold most of the properties, with another child node underneath for the file. You can hint at what some of the more core property values should be, but sling can get confused about what node to apply the hint to.

For example, I tried to specify pant:module instead of the default nt:file, but then both the parent node and the file node became pant:module, which was bad. I didn't have a good reason to fight with the OOTB behavior, so I chose to make some simple changes to our data structure instead.

src/main/java/com/redhat/pantheon/servlet/AsciidocRenderingServlet.java
carlosmunoz: When I first generated the project, it came with a few servlet examples. These examples used a handful of now-deprecated annotations. I replaced all those with standard osgi annotations without realizing there were some still-active Sling-specific ones. So, the comment just some praise as I learned something new :smiley:

src/main/java/com/redhat/pantheon/servlet/AsciidocRenderingServlet.java
carlosmunoz: I'd agree with that.

src/main/java/com/redhat/pantheon/servlet/AsciidocRenderingServlet.java
carlosmunoz: Those values are used on the OSGI management screens. For example, there is a screen where you see all the deployed components pre Bundle, and these values show up. I think it's a good documentation practice to have them, even if the code to do so is rather ugly (concatenating strings with property lines)

src/main/java/com/redhat/pantheon/servlet/AsciidocRenderingServlet.java
carlosmunoz: I had the exact same problem, so I know what you are referring to. I finally made it work so I can share those findings with you. Having said that, this is a good approach and I can see the possibilities. Lets catchup offline on the pros/cons of this structure.

src/main/java/com/redhat/pantheon/servlet/AsciidocRenderingServlet.java
benradey:
```suggestion
    private final Logger log = LoggerFactory.getLogger(AsciidocRenderingServlet.class);
```

carlosmunoz: :+1:

src/main/java/com/redhat/pantheon/servlet/AsciidocRenderingServlet.java
benradey: Ok, I put them back.

PR #4, opened 2019-03-06
add a travis-ci config and a maven wrapper

carlosmunoz:


benradey:
Carlos, I have no objection to merging this in. I know Ivan had some thoughts on whether this will ultimately be beneficial for us, but let's give it a try for now (if for no other reason, at least because I'd like to learn more about Travis). We can always change our minds later, right?

PR #6, opened 2019-03-08
Integrated Mark's templates and CSS

benradey:
The bulk of this code is the new TemplateDirectory class, which is a workaround for AsciidoctorJ not really being designed to work in an OSGI context. It wants templates as File objects, which can't refer to bundle resources, so unfortunately we have to write them out as temporary files so Asciidoctor can use them.

Also, getting a handle on bundle resources requires use of the BundleContext object which we had been ignoring up until this point. Our newfound usage of it breaks our test class. I'm open to suggestions on how best to address that. I haven't done much dependency injection before, but I'm pretty sure that model is designed to solve this sort of problem.


src/main/resources/SLING-INF/content/static
```diff
@@ -0,0 +1 @@
+../../../../../../rhdocs-styles/stylesheets
```
carlosmunoz: Is this a symlink to files outside of the project? If so, we might need a better way to do this.

src/main/java/com/redhat/pantheon/servlet/AsciidocRenderingServlet.java
```diff
@@ -110,8 +115,14 @@ private Content generateHtml(SlingHttpServletRequest request, Resource resource)
         RubyInstanceConfig config = new RubyInstanceConfig();
         config.setLoader(Thread.currentThread().getContextClassLoader());
 
-        Asciidoctor instance = Asciidoctor.Factory.create(
-                singletonList("uri:classloader:/gems/asciidoctor-1.5.8/lib"));
+        Enumeration<URL> gems = PantheonBundleActivator.getContext().getBundle().findEntries("gems", "*", false);
```
carlosmunoz: Why did we have to add these lines? I mean, this is about finding the ruby gems for asciidoctorj to work, which we weren't having a problem with before.

src/main/java/com/redhat/pantheon/asciidoctor/extension/TemplateDirectory.java
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Created by ben on 3/8/19.
+ */
+public class TemplateDirectory {
+
+    private static File tmplDir;
+
+    public static File get() {
+        if (tmplDir == null) {
+            Enumeration<URL> urls = PantheonBundleActivator.getContext().getBundle().findEntries("apps/pantheon/templates/haml/html5", "*", false);
```
carlosmunoz: I feel like we should do this in the Bundle activator itself.

Also, this class looks an awful lot like a singleton, and we should try to avoid singletons in clustered environments.

src/main/resources/apps/pantheon/templates/haml
```diff
@@ -0,0 +1 @@
+../../../../../../../rhdocs-templates/haml
```
carlosmunoz: same comment as before about symlinks.

These templates and stylesheets are a dependency for the project, but how to manage them?

src/main/resources/SLING-INF/content/static
benradey: It sure is. I'm open to suggestions. I had briefly discussed the idea of a git submodule with Ivan but he was pretty opposed, calling it more trouble than it was worth.

The templates & stylesheets are stored in Mark's personal repository for right now... how should we handle that?

src/main/java/com/redhat/pantheon/servlet/AsciidocRenderingServlet.java
benradey: It's because the templating functionality required the presence of additional gems. In particular, we needed tilt, haml, and thread_safe, which luckily were all already included in the jar (but we hadn't needed to point at them before, so they were being ignored).

So rather than hardcoding a bunch of directories which are likely going to change if we move to a newer version of asciidoctorj, I tried to make it somewhat dynamic. (There are actually 10 different gems directories in the JAR! Now we include them all.)

But there's still the assumption that the directories will take the form of /gems/xxxxx/lib. Hopefully that's safe but I'm open to better ideas.

src/main/java/com/redhat/pantheon/asciidoctor/extension/TemplateDirectory.java
benradey: Agree with the bundle activator idea. I will implement that.

Why was the singleton pattern bad in this case though? My logic was that someday this will be deployed on Openshift, and yes the pods won't be able to share this instance, but each pod will need its own set of /tmp template files, so that falls into the "feature not bug" category. Were you thinking of a different angle?

src/main/resources/apps/pantheon/templates/haml
benradey: I suggest we resolve this thread with the understanding that the original one applies to both symlinks.

src/main/resources/SLING-INF/content/static
carlosmunoz: Yeah, git submodules are a bad idea imo too. I'll resolve the discussion for now, but we need to figure out how to manage these types of dependencies.

We've used webjars (https://www.webjars.org) in the past, so we could look at it.

@markcaron any ideas?

src/main/java/com/redhat/pantheon/servlet/AsciidocRenderingServlet.java
carlosmunoz: Got it. 
We just need to take it out of this class. Sounds like a one time setup thing for asciidoctorj.

src/main/java/com/redhat/pantheon/asciidoctor/extension/TemplateDirectory.java
carlosmunoz: You're right; in this case it won't harm us. I am just being cautious that the moment we put mutable shared memory data on a singleton is where we fail. I guess we could store it in the JCR repository itself.

PR #8, opened 2019-03-21
various fixes

carlosmunoz:
reinstate pom.xml options to deploy to sling using the PostServlet
add upload parameter to specify the mimetype for asciidoc modules

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -73,8 +73,6 @@
                 <version>2.3.4</version>
                 <configuration>
                     <slingUrl>${sling.url}</slingUrl>
-                    <user>admin</user>
-                    <password>admin</password>
                     <slingUrlSuffix>/apps/pantheon/install</slingUrlSuffix>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -73,7 +73,10 @@
                 <version>2.3.4</version>
                 <configuration>
                     <slingUrl>${sling.url}</slingUrl>
-                    <password>ccsadmin</password>
+                    <user>admin</user>
+                    <password>admin</password>
```
benradey: Would you mind changing this to our non-default ccsadmin? Or, I wonder if we could handle this a bit better, perhaps using the profiles that we talked about earlier today. Certainly open to suggestions.

carlosmunoz: Done. Removed the explicit defaults (they are still defaults per the plugin). For now the way to customize this will be to provide the `-Dsling.user` and `-Dsling.password` properties if they differ.

PR #11, opened 2019-03-27
various refactorings

carlosmunoz:
Remove the need for the dependency provider classes, and instead move that logic to a component which can now be activated as soon as the bundle starts.
Also return for now to the per-request asciidoctor instance for thread safety.


pom.xml
```diff
@@ -252,6 +274,12 @@
             <version>2.16.3</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <artifactId>commons-io</artifactId>
+            <groupId>commons-io</groupId>
+            <version>2.6</version>
+            <scope>provided</scope>
```
benradey: I was going to leave a comment asking about bundle bloat - is it really worth it to include this when it really only saves us 4ish lines of code - but then I see that it's got a scope of provided. If we already get it for free, heck yeah we should use it! (Makes sense since it's all in the Apache family). Nice.

src/main/java/com/redhat/pantheon/sling/PantheonRepositoryInitializer.java
```diff
@@ -26,10 +28,16 @@
...
+public class PantheonRepositoryInitializer implements SlingRepositoryInitializer {
+
+    private static final Logger log = LoggerFactory.getLogger(SlingRepositoryInitializer.class);
 
     @Override
     public void processRepository(SlingRepository slingRepository) throws Exception {
+        initializeRepositoryACLs(slingRepository);
+    }
+
+    private void initializeRepositoryACLs(SlingRepository slingRepository) throws RepositoryException {
```
benradey: What was the purpose of pulling this out into its own method? Just for readability's sake?

pom.xml
carlosmunoz: yup, first thing I check is whether the necessary classes are already provided in an existing bundle.

src/main/java/com/redhat/pantheon/sling/PantheonRepositoryInitializer.java
carlosmunoz: to separate the different tasks that we would be doing when initializing the repository. Right now it's just creating some ACLs, but there may be more in the future. I did this refactoring originally when I had the code to load gem paths in this class.

PR #12, opened 2019-03-28
Ccs 2701

benradey:
Adding the bulk upload script and tweaking ACLs to accommodate the repository node structure.


src/main/java/com/redhat/pantheon/servlet/AsciidocRenderingServlet.java
```diff
@@ -140,9 +140,11 @@ private Content generateHtml(SlingHttpServletRequest request, Resource resource)
             ob = ob.templateDir(dependencyProvider.getTemplateDir());
         }
 
+        long start = System.currentTimeMillis();
         c.html = getDependencyProvider().getAsciidoctor().convert(
                 c.asciidoc,
                 ob.get());
+        System.out.println("Rendering finished in " + (System.currentTimeMillis() - start) + " ms.");
```
carlosmunoz: Lets use the `log` variable here, instead of `System.out`.

benradey: Boooooo! System.out FOREVER! :)

benradey:
> I think we need to bring the whole Sling content management bundle into a sub-directory, at the same level as the uploader (maybe as a different pull request).

Can you clarify what you mean by this?

Also, please re-review, I added a little bit of logging configuration. (Not strictly related but seemed like a good time).

carlosmunoz:
This branch needs conflict resolution. Easier to review once these have been applied.

uploader/push_to_sling.py
```diff
@@ -0,0 +1,98 @@
...
+parser.add_argument('--user', '-u', help='Username for authentication, default \'demo\'', default='demo')
+parser.add_argument('--password', '-p', help='Password for authentication, default \'demo\'. If \'-\' is supplied, the script will prompt for the password.', default='demo')
+parser.add_argument('--directory', '-d', help='Directory to upload, default is current working directory. (' + os.getcwd() + ')', default=os.getcwd())
+args = parser.parse_args()
+
+pw = args.password
+if pw == '-':
+    pw = getpass.getpass()
+
+print('Using server: ' + args.server)
```
ivanhorvath: I'm strongly against using print for any purpose, but especially for an app that can potentially be like a regular unix tool, thus emits nothing upon successful run. I like to use logging.console for this purpose so the output can be suppressed, or redirected to a log file without needing stdout/err redirection of the shell it's running under. But as a first version this is fine, just want to start a discussion on the subject.

uploader/push_to_sling.py
```diff
@@ -0,0 +1,98 @@
...
+    if parent_dir_str == '.':
+        parent_dir_str = ''
+    # file becomes a/file/name (no extension)
+    base_name = path.stem
+
+    url = args.server + "/content/repositories/" + args.repository
+    if parent_dir_str:
+        url += '/' + parent_dir_str
+
+    # print('base name: ' + base_name)
```
ivanhorvath: Let's try to clean up commented out debug lines or convert them to logging.debug, so they can be used later for troubleshooting. Again, not critical, but it's good practice.

uploader/push_to_sling.py
```diff
@@ -0,0 +1,98 @@
...
+            files = {'asciidoc': ('asciidoc', open(path, 'rb'), 'text/x-asciidoc')}
+
+            # Minor question: which is correct, text/asciidoc or text/x-asciidoc?
+            # It is text/x-asciidoc. Here's why:
+            # https://tools.ietf.org/html/rfc2045#section-6.3
+            # Paraphrased: "If it's not an IANA standard, use the 'x-' prefix.
+            # Here's the list of standards; text/asciidoc isn't in it.
+            # https://www.iana.org/assignments/media-types/media-types.xhtml#text
+
+            r = requests.post(url, headers=HEADERS, data=data, files=files, auth=(args.user, pw))
```
ivanhorvath: It's good practice to wrap most http calls with some exception handling, since response codes are in the very high numbers and growing, so some of the more critical ones can be handled or ignored. Let's discuss this with Ankit about what we want our python guidelines to be. But we need to handle when 1 out of 100 files comes back with a network error or 400 response. Should the rest be deleted? Marked incomplete? All uploaded in a state where nobody has viewing rights to them until all arrive as a cohesive unit? Let's brainstorm one morning about the outcomes of various use cases.

src/main/java/com/redhat/pantheon/servlet/AsciidocRenderingServlet.java
```diff
@@ -140,9 +140,11 @@ private Content generateHtml(SlingHttpServletRequest request, Resource resource)
             ob = ob.templateDir(dependencyProvider.getTemplateDir());
         }
 
+        long start = System.currentTimeMillis();
         c.html = getDependencyProvider().getAsciidoctor().convert(
                 c.asciidoc,
                 ob.get());
+        log.info("Rendering finished in {} ms.", System.currentTimeMillis() - start);
```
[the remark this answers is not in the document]
benradey: I prefer .info here, since with the caching in place we shouldn't hit this all that often - I'd like to know when we do hit it so I can make sure that 1.) it's not unexpected, and 2.) we get a feel for how long it takes on average. (And by extension, if a request IS taking a long time, is it asciidoctor or something else?)

uploader/push_to_sling.py
benradey: I went and looked up some info about the logging module and this page contained this information:
https://docs.python.org/3/howto/logging.html
Task you want to perform: Display console output for ordinary usage of a command line script or program
The best tool for the task: print()

For this particular output, I think print() is correct - the user would always want to make sure they didn't accidentally flub the server or repo. But for the other print's in the script, you're correct, and I'll switch those over.

uploader/push_to_sling.py
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/python3
+import os
+import requests
+import getpass
+import argparse
+import getpass
```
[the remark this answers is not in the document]
benradey: Good catch.

uploader/push_to_sling.py
benradey: Agreed.

uploader/push_to_sling.py
benradey: Good thinking, Ivan. I'll leave this as-is for now, but you're absolutely right, we should implement some form of error-handling as a future story.

src/main/java/com/redhat/pantheon/servlet/AsciidocRenderingServlet.java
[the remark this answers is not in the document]
carlosmunoz: :+1:

PR #13, opened 2019-04-01
Ccs 2706

benradey:
Significantly simplified module html in order to facilitate easy client-side sorting. Pretty CSS can be a future story.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -57,7 +57,8 @@
                             SLING-INF/oak:index;overwrite:=true;uninstall:=true;path:=/oak:index
                         ]]></Sling-Initial-Content>
                         <Sling-Model-Packages>
-                            com.redhat.pantheon.model
+                            com.redhat.pantheon.model,
+                            com.redhat.pantheon.use
                         </Sling-Model-Packages>
                         <Import-Package>
                             !jnr.a64asm,!jnr.x86asm,!org.apache.bsf,!org.apache.bsf.util,
--- a/src/main/java/com/redhat/pantheon/use/DateFormatter.java
+++ b/src/main/java/com/redhat/pantheon/use/DateFormatter.java
@@ -1,0 +1,25 @@
+package com.redhat.pantheon.use;
+
+import org.apache.sling.api.SlingHttpServletRequest;
+import org.apache.sling.models.annotations.Model;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+import java.text.SimpleDateFormat;
+import java.util.Calendar;
+
+/**
+ * Created by ben on 4/1/19.
+ * http://blogs.adobe.com/experiencedelivers/experience-management/htl-date-formatting/
+ * https://sling.apache.org/documentation/bundles/models.html#basic-usage
+ */
+@Model(adaptables = SlingHttpServletRequest.class)
+public class DateFormatter {
+
+    public String value;
+
+    @Inject
+    public DateFormatter(@Named("format") String format, @Named("date") Calendar date) {
+        value = new SimpleDateFormat(format).format(date.getTime());
+    }
+}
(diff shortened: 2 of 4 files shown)


src/main/java/com/redhat/pantheon/use/DateFormatter.java
```diff
@@ -0,0 +1,25 @@
...
+@Model(adaptables = SlingHttpServletRequest.class)
+public class DateFormatter {
+
+    public String value;
+
+    @Inject
+    public DateFormatter(@Named("format") String format, @Named("date") Calendar date) {
+        value = new SimpleDateFormat(format).format(date.getTime());
+    }
+}
```
carlosmunoz: Wow, this I want to hear more about! :smile:

pom.xml
```diff
@@ -57,7 +57,8 @@
                             SLING-INF/oak:index;overwrite:=true;uninstall:=true;path:=/oak:index
                         ]]></Sling-Initial-Content>
                         <Sling-Model-Packages>
-                            com.redhat.pantheon.model
+                            com.redhat.pantheon.model,
+                            com.redhat.pantheon.use
```
carlosmunoz: Mmm... lets talk about our package structure. We need to start thinking about how we want to structure our code... there are different options.

src/main/resources/apps/pantheon/modules/all/html.html
```diff
@@ -10,35 +10,89 @@ <h1>All Modules</h1>
...
+        </sly>
+    </table>
 </div>
 
 <div data-sly-include="${'/apps/pantheon/partials/footer.html'}">
 </div>
+
+<!-- https://www.w3schools.com/howto/howto_js_sort_table.asp -->
+<script>
+function sortTable(n) {
```
carlosmunoz: This is the type of code we need to have unit tested and inlining it doesn't help with that. This is where a good front end framework helps (e.g. Angular, etc). We might need to discover how to do this while continuing to take advantage of HTL.

src/main/java/com/redhat/pantheon/use/DateFormatter.java
benradey: What do you want to know? :)

This enables the passing of parameters via the data-sly-use property on src/main/resources/apps/pantheon/modules/all/html.html:29.

One note... Injection via constructor annotations works. Injection via field annotations, followed by a @PostConstruct method annotation, did not. No idea why. All I can say is that my favorite phrase "too much magic" applies here.

pom.xml
benradey: Sure, open to suggestions. There is also a `<Sling-Model-Classes>` tag we can use to be more granular.

src/main/resources/apps/pantheon/modules/all/html.html
benradey: Yeah, I knew this wasn't the greatest solution, but my thought was I'll get something quick and dirty in now, and we can implement a proper UI in a future story.

src/main/java/com/redhat/pantheon/use/DateFormatter.java
benradey: IMO, sorting is a good candidate for being done client-side. The server doesn't have any info the client doesn't have and waiting for the round-trip takes time. That said, I could imagine at least one way to accomplish a server-side sort using URL params. (It's clumsy, and I'm sure there are better ways, but I don't know them offhand).

benradey: You know, when I had been thinking about sorting, it was sorting titles - and I had thought it was reasonable to always send the full list to the client. Now I realize we're talking about sorting modules, which could be much more numerous. In that case perhaps server-side sorting is the best approach.

PR #14, opened 2019-04-04
Ccs 2749

benradey:
Implementing a basic thread-safe asciidoctor pool

### DIFF
--- a/src/main/java/com/redhat/pantheon/conf/AsciidoctorPoolService.java
+++ b/src/main/java/com/redhat/pantheon/conf/AsciidoctorPoolService.java
@@ -1,0 +1,55 @@
+package com.redhat.pantheon.conf;
+
+import com.redhat.pantheon.asciidoctor.extension.SlingResourceIncludeProcessor;
+import org.apache.sling.api.resource.Resource;
+import org.asciidoctor.Asciidoctor;
+import org.osgi.service.component.annotations.Activate;
+import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.Reference;
+import org.osgi.service.component.annotations.ServiceScope;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.concurrent.ConcurrentLinkedQueue;
+
+/**
+ * Created by ben on 4/2/19.
+ */
+@Component(
+        service = AsciidoctorPoolService.class,
+        scope = ServiceScope.SINGLETON
+)
+public class AsciidoctorPoolService {
+
+    // Purposefully specifying the ConcurrentLinkedQueue here because it is thread-safe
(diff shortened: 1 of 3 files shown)


jmarrero:
1. Code makes sense to me
2. Tests were modified for the change

src/main/java/com/redhat/pantheon/conf/AsciidoctorPoolService.java
```diff
@@ -0,0 +1,57 @@
...
+        queue.offer(asciidoctor);
+        log.trace("Returned Asciidoctor instance to pool, new pool size: {}", queue.size());
+    }
+
+    public Asciidoctor requestInstance(Resource resource) throws IOException {
+        Asciidoctor ret = queue.poll();
+        if (ret == null) {
+            log.trace("No Asciidoctor instances available from pool, creating...");
+            synchronized (this) {
+                ret = Asciidoctor.Factory.create(localFileManagementService.getGemPaths());
```
carlosmunoz: We are not limiting the number of instances created in the pool, which might lead to memory issues. Not probable, but possible.

src/main/java/com/redhat/pantheon/servlet/AsciidocRenderingServlet.java
```diff
@@ -145,9 +149,11 @@ private Content generateHtml(SlingHttpServletRequest request, Resource resource)
         }
 
         long start = System.currentTimeMillis();
-        c.html = getAsciidoctorInstance(resource).convert(
+        Asciidoctor asciidoctor = asciidoctorPoolService.requestInstance(resource);
+        c.html = asciidoctor.convert(
                 c.asciidoc,
                 ob.get());
+        asciidoctorPoolService.releaseInstance(asciidoctor);
```
carlosmunoz: I think this should be in a `try - finally` block, to guarantee the instance gets returned even in the case of runtime exceptions.

src/main/java/com/redhat/pantheon/conf/AsciidoctorPoolService.java
benradey: Yeah, I thought about that. What would we set the limit to? 5? 500? I don't know what's reasonable, so I thought 'better to just not solve a problem we don't have yet.' It would be trivial for us to add a limit later if we find that we need one.

src/main/java/com/redhat/pantheon/servlet/AsciidocRenderingServlet.java
benradey: Good idea!

src/main/java/com/redhat/pantheon/conf/AsciidoctorPoolService.java
benradey: @carlosmunoz do you object to merging this in with no limit?

PR #18, opened 2019-04-08
Removing use of Optional class and adding appropriate null checks

benradey:
This remedies the mysterious 404 issue (a NPE that Sling was just swallowing) that Joseph uncovered when missing Red Hat templates and css.

### DIFF
--- a/src/main/java/com/redhat/pantheon/conf/LocalFileManagementService.java
+++ b/src/main/java/com/redhat/pantheon/conf/LocalFileManagementService.java
@@ -15,9 +15,9 @@ import java.net.URL;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Enumeration;
 import java.util.List;
-import java.util.Optional;
 
 @Component(
         service = LocalFileManagementService.class,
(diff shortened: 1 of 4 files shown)


carlosmunoz:
But why remove the optionals? :cry: 
Personally, I think they are way more elegant than null checks.

For reference: https://www.deadcoderising.com/2015-10-06-java-8-removing-null-checks-with-optional/

benradey:
> But why remove the optionals? cry
> Personally, I think they are way more elegant than null checks.
> 
> For reference: https://www.deadcoderising.com/2015-10-06-java-8-removing-null-checks-with-optional/

Ok, we brought them back (as you already know)... mostly. I don't think an Optional is appropriate for gemPaths - since if that's null, we have big problems. Otherwise though, how do you feel about this PR now?

src/main/java/com/redhat/pantheon/conf/LocalFileManagementService.java
```diff
@@ -38,54 +40,78 @@ public void activate() throws IOException {
...
-
-            FileUtils.copyInputStreamToFile(is, f);
-        }
+//        if (urls != null && urls.hasMoreElements()) {
+//            templateDirectory = Optional.of(p.toFile());
+//        }
+        // This .filter(a -> a) is subtle. I believe it is necessary - without the filter, if the Enumeration is not
+        // null BUT it has no elements, the ifPresent call will execute. WITH the filter, the ifPresent call would NOT
+        // execute in the same scenario.
+        urls.map(Enumeration::hasMoreElements).filter(a -> a).ifPresent(a -> templateDirectory = Optional.of(p.toFile()));
```
carlosmunoz: Per our discussion lets change this.

PR #19, opened 2019-04-12
Python script no longer includes/excludes files based on filename...

benradey:
...but does include hidden (dot-prefixed) files and directories.

Also made an unrelated change to pom.xml, I discovered an HTL validation plugin that seemed like a good idea to include generally.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -123,6 +123,24 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>org.apache.sling</groupId>
+                <artifactId>htl-maven-plugin</artifactId>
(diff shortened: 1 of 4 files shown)


ivanhorvath:
LGTM, however in the future let's move towards having different commits at least for different features. That way they can be reverted or partially merged individually and not disturb the rest of the code.

benradey:
Please re-review, I uploaded new changes to ModuleData.java to 1.) show files that were uploaded using the python script, and 2.) identify the source of the listings in the html. Both of those things were accomplished by adding new pant:transient fields in the ModuleData code. This is a new technique, feedback is welcome. @redhataccess/pantheon-developers

PR #20, opened 2019-04-12
Ccs 2753

benradey:
Continuing from #18 which accidentally closed due to a commit snafu.

### DIFF
--- a/src/main/java/com/redhat/pantheon/conf/LocalFileManagementService.java
+++ b/src/main/java/com/redhat/pantheon/conf/LocalFileManagementService.java
@@ -46,7 +46,9 @@ public class LocalFileManagementService {
 
         log.info("Initializing template directories at " + p.toString());
 
-        urls.ifPresent(a -> templateDirectory = Optional.of(p.toFile()));
+        if (urls.isPresent()) {
+            templateDirectory = Optional.of(p.toFile());
+        }
 
         Collections.list(urls.orElse(Collections.emptyEnumeration()))
                 .stream()
--- a/src/main/java/com/redhat/pantheon/conf/LocalFileManagementService.java
+++ b/src/main/java/com/redhat/pantheon/conf/LocalFileManagementService.java
@@ -51,10 +51,9 @@ public class LocalFileManagementService {
 //        if (urls != null && urls.hasMoreElements()) {
 //            templateDirectory = Optional.of(p.toFile());
 //        }
-        // This .filter(a -> a) is subtle. I believe it is necessary - without the filter, if the Enumeration is not
-        // null BUT it has no elements, the ifPresent call will execute. WITH the filter, the ifPresent call would NOT
-        // execute in the same scenario.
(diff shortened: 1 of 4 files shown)


src/main/java/com/redhat/pantheon/servlet/AsciidocRenderingServlet.java
```diff
@@ -144,9 +143,7 @@ private Content generateHtml(SlingHttpServletRequest request, Resource resource)
                 // Generate the html header and footer
                 .headerFooter(true)
                 .attributes(atts);
-        if (localFileManagementService.getTemplateDirectory() != null) {
-            ob = ob.templateDir(localFileManagementService.getTemplateDirectory());
-        }
+        localFileManagementService.getTemplateDirectory().ifPresent(ob::templateDir);
```
carlosmunoz: nice :+1:

src/main/java/com/redhat/pantheon/conf/LocalFileManagementService.java
```diff
@@ -38,54 +40,77 @@ public void activate() throws IOException {
...
+//            String filename = url.toString();
+//            filename = filename.substring(filename.lastIndexOf("/") + 1);
+//
+//            File f = new File(templateDirectory.get(), filename);
+//            f.deleteOnExit();
+//
+//            InputStream is = url.openConnection().getInputStream();
+//
+//            FileUtils.copyInputStreamToFile(is, f);
+//        }
```
carlosmunoz: Lets get rid of commented out code.

benradey: I left it in place as a good example between old-school and new-school, but ok, I removed it. I understand we don't want to clutter our codebase. I also took the opportunity to change the if-statement in the middle too, let me know if you have thoughts on lambdas in the form of "a -> (something unrelated to 'a')"

src/main/java/com/redhat/pantheon/conf/LocalFileManagementService.java
```diff
@@ -38,54 +40,58 @@ public void activate() throws IOException {
...
-            filename = filename.substring(filename.lastIndexOf("/") + 1);
-
-            File f = new File(templateDirectory.get(), filename);
-            f.deleteOnExit();
-
-            InputStream is = url.openConnection().getInputStream();
-
-            FileUtils.copyInputStreamToFile(is, f);
-        }
+        urls.ifPresent(a -> templateDirectory = Optional.of(p.toFile()));
```
benradey: @carlosmunoz This is what I'm talking about in the other comment.

carlosmunoz: I think a traditional if statement makes this more readable. The fact that `a` is not consumed is a bit of a distraction.

benradey: Ok, I understand. I put the if-statement back (using isPresent instead of orElse).

PR #22, opened 2019-04-15
Initial backend search in html

jmarrero:
Initial backend search, will build on top of this with a front-end framework next.

Note, I changed a few variables and methods to static to be able to grab the data from sly/htl at any point on  src/main/java/com/redhat/pantheon/use/ModuleData.java .

The javax.script.Bindings don't get passed down to the class when I create a new instance of the class, it looks like it they get populated at app start by sling. By making it static I make sure the bindings that populate the resolver variable at init() are are accessible when I do each search.

Simple test case added to build on top of it later too.

### DIFF
--- a/src/main/java/com/redhat/pantheon/use/ModuleData.java
+++ b/src/main/java/com/redhat/pantheon/use/ModuleData.java
@@ -1,6 +1,5 @@
 package com.redhat.pantheon.use;
 
-import com.google.common.collect.Lists;
 import org.apache.sling.api.resource.Resource;
 import org.apache.sling.api.resource.ResourceResolver;
 import org.apache.sling.scripting.sightly.pojo.Use;
@@ -10,12 +9,10 @@ import org.slf4j.LoggerFactory;
 import javax.jcr.query.Query;
 import javax.script.Bindings;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
-import java.util.stream.Collectors;
 
 /**
  * Created by ben on 3/5/19.
@@ -26,6 +23,7 @@ public class ModuleData implements Use {
 
     private ResourceResolver resolver;
     private Resource currentResource;
+    private String searchTerm;
 
     private final Logger log = LoggerFactory.getLogger(ModuleData.class);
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/redhat/pantheon/use/ModuleData.java
```diff
@@ -8,40 +8,51 @@
...
+import javax.inject.Inject;
+
 
 /**
  * Created by ben on 3/5/19.
  */
 public class ModuleData implements Use {
 
-    private ResourceResolver resolver;
+    private static ResourceResolver resolver;
```
carlosmunoz: Not sure making this variable `static` is a good idea here. If multiple `ModuleData` objects are used simultaneously (which is likely the case), sling will just overwrite the resolver every time.

src/main/resources/apps/pantheon/modules/all/search.html
```diff
@@ -0,0 +1,111 @@
...
+
+    <table id="myTable">
+        <tr>
+            <th width="200px"><a href="#" onclick="sortTable(0)">Name</a></th>
+            <th width="325px"><a href="#" onclick="sortTable(1)">Description</th>
+            <th><a href="#" onclick="sortTable(2)">Upload Time</a></th>
+        </tr>
+
+        <sly data-sly-use.resultList="${'com.redhat.pantheon.use.DataRetriever' @ query=request.requestParameterMap['search'][0].toString}">
+        <sly data-sly-list=${resultList.data}>
```
benradey: Can we please get rid of search.html and put all the logic back into html.html? I remember you told me that the reason for search.html was so that you could guarantee that the 'search' parameter was specified. I came up with another technique that mitigates that need:
`        <sly data-sly-test.searchIsNull="${!request.requestParameterMap['search']}"></sly>
        <sly data-sly-use.resultList="${'com.redhat.pantheon.use.DataRetriever' @ query=(searchIsNull ? '*' : request.requestParameterMap['search'][0].toString )}">`

benradey: Can we please get rid of search.html and put all the logic back into html.html? I remember you told me that the reason for search.html was so that you could guarantee that the 'search' parameter was specified. I came up with another technique that mitigates that need:
`        <sly data-sly-test.searchIsNull="${!request.requestParameterMap['search']}"></sly>
        <sly data-sly-use.resultList="${'com.redhat.pantheon.use.DataRetriever' @ query=(searchIsNull ? '*' : request.requestParameterMap['search'][0].toString )}">`

src/main/resources/apps/pantheon/modules/all/html.html
```diff
@@ -1,86 +1,9 @@
 <html>
-<head data-sly-include="${'/apps/pantheon/partials/header.html'}">
+<head>
 </head>
 <body>
-
-<div class="container-fluid" data-sly-use.moduleData="com.redhat.pantheon.use.ModuleData">
```
benradey: I am pretty sure we can get rid of your DataRetriever class and bring the query functionality into the existing ModuleData class. To do that, change this line to something like this:
`<div class="container-fluid" data-sly-use.moduleData="${'com.redhat.pantheon.use.ModuleData' @ benTestVar='Hello world'}">`
...and then in the ModuleData class, you can use code like this to retrieve the value in the init() method:
`log.info("Ben Test Var: " + bindings.get("benTestVar"));`

benradey: My intention here is to combine this technique with my other comment. Use data-sly-test to figure out whether the user specified a 'search' param on the URL, and depending on that, pass the correct value into the ModuleData class via this. Then we can eliminate both the search.html and the DataRetriever.java files.

src/main/java/com/redhat/pantheon/use/ModuleData.java
jmarrero: I think my understanding of the resolver is was mistaken, I was thinking that I did not care if it got replaced because I would get the new state of the data instead. And given that I mean just looking to see what modules are available It does not matter if the data was updated just as I was searching.

jmarrero: However, I have updated the code with Ben's suggestions and this takes care of my initial need for a static resolver.

src/main/resources/apps/pantheon/modules/all/search.html
jmarrero: thanks, this is much nicer.

src/main/resources/apps/pantheon/modules/all/html.html
jmarrero: awesome I get it, I was guiding my decision on the searches I did looking on ways to pass parameters to a method thru sly which ended up with finding that I could only pass parameters down to the constructor of the class. But this is passing it via the bindings... This is way better than refactoring the code like I was doing.

anknite:
Just a thought, would it be a good idea to merge the commits, for a cleaner commit history?

jmarrero:
Thanks, I will squash and merge in sec.

src/main/java/com/redhat/pantheon/use/ModuleData.java
```diff
@@ -20,28 +20,38 @@
 
     private ResourceResolver resolver;
     private Resource currentResource;
-
+    private String searchTerm;
+    
     private final Logger log = LoggerFactory.getLogger(ModuleData.class);
-
+
```
anknite: is this an extra space :p

jmarrero: thanks, fixed in the updated code.

jmarrero:
rebased on top of Ben's latest changes, and addressed Ankit's comments.

PR #28, opened 2019-05-03
Uploader updates (CCS-2824)

benradey:
Reworking the uploader for a few new features:

- Config file support via pantheon2.yml
- Whitelist-based file pattern globbing (defined in the config file)
- Support for optionally following symlinks in a directory


uploader/push_to_sling.py
```diff
@@ -4,28 +4,60 @@
 import argparse
 import getpass
 import logging
-from pathlib import Path
+import yaml
 from pathlib import PurePath
 from datetime import datetime
 
+DEFAULT_SERVER = 'http://localhost:8080'
+DEFAULT_REPOSITORY = getpass.getuser() + '_' + datetime.now().strftime('%Y-%m-%d_%H-%M-%S')
```
carlosmunoz: I would leave the default repository as just the user name (without the current date component). That way it's ensured that they are working on their own local repository every time and not create spurious ones.

uploader/push_to_sling.py
```diff
@@ -4,28 +4,60 @@
...
-parser.add_argument('--repository', '-r', help='The name of the repository; this is a convenience for the user. If not specified, one will be autogenerated', default=getpass.getuser() + '_' + datetime.now().strftime('%Y-%m-%d_%H-%M-%S'))
-parser.add_argument('--user', '-u', help='Username for authentication, default \'demo\'', default='demo')
-parser.add_argument('--password', '-p', help='Password for authentication, default \'demo\'. If \'-\' is supplied, the script will prompt for the password.', default='demo')
+parser.add_argument('--server', '-s', help='The Pantheon server to upload modules to, default ' + DEFAULT_SERVER)
+parser.add_argument('--repository', '-r', help='The name of the repository; this is a convenience for the user. If not specified, one will be autogenerated')
+parser.add_argument('--user', '-u', help='Username for authentication, default \'' + DEFAULT_USER + '\'', default=DEFAULT_USER)
+parser.add_argument('--password', '-p', help='Password for authentication, default \'' + DEFAULT_PASSWORD + '\'. If \'-\' is supplied, the script will prompt for the password.', default=DEFAULT_PASSWORD)
```
carlosmunoz: Consider removing the default username/password combination. Once in production, users will need to specify credentials every time, so might as well enforce the requirement early on.

uploader/push_to_sling.py
```diff
@@ -4,28 +4,60 @@
...
 parser.add_argument('--directory', '-d', help='Directory to upload, default is current working directory. (' + os.getcwd() + ')', default=os.getcwd())
-parser.add_argument('--log', '-l', help='Verbosity of logging, default is \'warning\'', default='warning')
+parser.add_argument('--links', '-l', help='Resolve symlinks when searching for files to upload', action='store_const', const=True)
+parser.add_argument('--verbose', '-v', help='Print information that may be helpful for debugging', action='store_const', const=True)
+parser.add_argument('--dry', '-D', help='Dry run; print information about what would be uploaded, but don\'t actually upload', action='store_const', const=True)
```
carlosmunoz: Nice one :+1:

uploader/push_to_sling.py
```diff
@@ -4,28 +4,60 @@
...
+parser.add_argument('--sample', '-S', help='Print a sample pantheon2.yml file to stdout (which you may want to redirect to a file).', action='version', version='''\
+server: http://localhost:8080
+repository: pantheonSampleRepo
+followlinks: true
+
+titles:
```
carlosmunoz: I know the intention here, and I'm all for it, just making sure we understand the differences between titles and modules.

uploader/push_to_sling.py
```diff
@@ -4,28 +4,60 @@
...
+ - master.adoc
+
+modules:
+ - shared/*.module.adoc
+ - shared/*.internal.adoc
```
carlosmunoz: Question: there's no longer any need for the `.module` and `.internal` naming conventions, right?

uploader/push_to_sling.py
```diff
@@ -36,74 +68,114 @@
...
     pw = getpass.getpass()
 
-print('Using server: ' + args.server)
-print('Using repository: ' + args.repository)
-print('--------------')
+config = None
+try:
+    config = yaml.safe_load(open(args.directory + '/pantheon2.yml'))
+except FileNotFoundError:
+    logger.warning('Could not find a valid config file in this directory; all files will be treated as resource uploads.')
```
carlosmunoz: Would a better default be perhaps to stop the upload altogether if no config file is found?

uploader/push_to_sling.py
```diff
@@ -36,74 +68,114 @@
...
+
+server = resolveOption(args.server, 'server', DEFAULT_SERVER)
+repository = resolveOption(args.repository, 'repository', DEFAULT_REPOSITORY)
+links = resolveOption(args.links, 'followlinks', DEFAULT_LINKS)
+
+print('Using server: ' + server)
+print('Using repository: ' + repository)
+print('--------------')
+
+# pathlist = Path(args.directory).glob('**/*')
```
carlosmunoz: lets get rid of commented out code

uploader/push_to_sling.py
```diff
@@ -36,74 +68,114 @@
...
+titleGlobs = config['titles'] if config is not None else ()
+moduleGlobs = config['modules'] if config is not None else ()
+resourceGlobs = config['resources'] if config is not None else '*'
+
+# for path in pathlist:
```
carlosmunoz: same here... lets remove commented out code

uploader/push_to_sling.py
```diff
@@ -36,74 +68,114 @@
...
+            # file becomes a/file/name (no extension)
+
+            url = server + "/content/repositories/" + repository
+            if parent_dir_str:
+                url += '/' + parent_dir_str
+
+            logger.debug('base name: %s', base_name)
+
+            # Asciidoc content (treat as a module)
+            if path.suffix == '.adoc' or path.suffix == '.asciidoc':
```
carlosmunoz: I think this condition should be replaced with:
```python
if isModule:
```
This way we upload modules as modules.

uploader/push_to_sling.py
```diff
@@ -36,74 +68,114 @@
...
+                # Paraphrased: "If it's not an IANA standard, use the 'x-' prefix.
+                # Here's the list of standards; text/asciidoc isn't in it.
+                # https://www.iana.org/assignments/media-types/media-types.xhtml#text
+
+                if not args.dry:
+                    r = requests.post(url, headers=HEADERS, data=data, files=files, auth=(args.user, pw))
+                    print(r.status_code, r.reason)
+                logger.debug('')
+            # Otherwise just upload as a regular file
+            else:
```
carlosmunoz: Similarly, this should be:
```python
elif isResource:
```
We can deal with titles later.

uploader/push_to_sling.py
benradey: The reason that I added the date component was in case we have 2 testing users that happen to have the same username. Remember that we are grabbing the local machine's user, not the kerberos ID, so we could easily have 2 people working in a "jon" repository (for example). If you are trying to avoid repository soup, perhaps we could come up some other pseudo-unique naming convention - maybe something including the machine hostname? Still not perfect, but better.

uploader/push_to_sling.py
benradey: Yeah, we should have a conversation about credentials. I would be open to storing them in a separate config file, maybe ~/.config/pantheon2.cfg, and we check that the file is chmod 600 or else we throw an error, similar to ssh. As an advanced feature, we could also assign per-user tokens/keys, although I'm not sure there's enough payoff there, but it's a possibility.

For now, since the demo credentials are the only ones that are available, I think leaving them here as defaults makes sense. When we get real auth working, let's revisit this.

uploader/push_to_sling.py
benradey: Not sure what you're referring to, but I'll take it! :)

uploader/push_to_sling.py
benradey: Well right now there is no difference, I just want to set us up for success in the future when the distinction becomes important.

uploader/push_to_sling.py
benradey: That's right, I was just reusing my old sandbox for testing. The naming conventions no longer apply, so if this is confusing, we could use any other example.

uploader/push_to_sling.py
carlosmunoz: Ah. As long as we understand this is not the full solution I'm good. The only problem is that should a test user want to change a few lines on a module and re-upload, they will get a completely new namespace.

uploader/push_to_sling.py
carlosmunoz: :+1:

uploader/push_to_sling.py
carlosmunoz: I was referring to the dry run mode :smile:  ... it'll be useful for users to validate what they are pushing before actually doing the push.

uploader/push_to_sling.py
carlosmunoz: ok :+1:

uploader/push_to_sling.py
carlosmunoz: ok good. Simple enough change to remove the convention from the example so I'll leave this comment open.

uploader/push_to_sling.py
benradey: I thought about that and I could be convinced that that's the way to go, but if we're going to be requiring the config file anyway, then we may as well drop half of the command line arguments. (Maybe that's a good thing for simplicity's sake?)

I had thought that maybe some users would want to script their uploads and perhaps having the options exposed as arguments rather than a file-based config might be more convenient for that, but maybe that's not the case.

uploader/push_to_sling.py
benradey: Good catch! I was just setting up `isModule` for future use, but I think you have a good point, we might be able to key off that here. However... I have a few thoughts.
- I'm not sure it's wise to key off of `isModule` alone... the user could easily accidentally write a wildcard glob that flags some .jpg's as modules. (Not that the file extension is much more reliable, but any tomfoolery there is more deliberate).
- What happens if we upload a .adoc file as a resource (not a module)? Probably the system would not play nicely with that in its current state. We'd need to test this pretty thoroughly.

uploader/push_to_sling.py
benradey: Yeah, I'm thinking that for the time being, rather than keying off of `isXxxx` flags, we keep the filename convention - BUT we use the `isXxxx` flag to set some new property on the .adoc uploads, like `pant:renderable` or something to that effect. We set it to true for `isModule` (and `isTitle`, hopefully?), and false for `isResource`.

That way, we still have asciidoc content where we expect it for JCR nodes regardless of whether it's a module/title/resource, but we're also not trying to render content that isn't standalone.

uploader/push_to_sling.py
benradey: Yeah, the long-term plan is to have the repo identified permanently in the pantheon2.cfg file. I'll change the default repo name to username_hostname, and the user will still have the option to specify a different one if they so choose.

carlosmunoz: :+1:

uploader/push_to_sling.py
carlosmunoz: that makes sense. so maybe what we need to make sure is that all parameters can be provided either via config file *or* command line, and which one precedes the other.

uploader/push_to_sling.py
carlosmunoz: So...

- the code computes the `isModule` value for each file. Why not just add to the current condition something like (pseudo code):
```python
isModule = matches(path, moduleGlobs, 'module') AND path.endsWith('.adoc') if not isTitle else False
```
There is no way stop all tomfoolery, but at least we can have some basic intelligence

- If a user uploads an adoc file as a non-module (totally valid use case btw), then it won't list as a module and it won't be previewable. But there is no problem there, as fixing the config file and re-uploading should do the trick.

benradey: I think I'd prefer not to look at file extension at all, since we're on the topic. Perhaps some author names their files .ascii or something and we're not set up to handle that. Let's just trust the authors to get the globs right - we can always tweak it later if need be. (Sounds like that was your original suggestion anyway)

>     * If a user uploads an adoc file as a non-module (totally valid use case btw), then it won't list as a module and it won't be previewable. But there is no problem there, as fixing the config file and re-uploading should do the trick.

The [possible] problem I'm worried about is in the includeProcessor. Right now it's set up to assuming that it can find the adoc content in jcrNode -> asciidocNode -> jcr:content (I think). Modules fit that pattern, but resources (like .jpgs) do not - there is no asciidoc node for .jpgs, for example. So if we upload a .adoc resource the same way we upload a .jpg, the includeProcessor won't handle it correctly.

uploader/push_to_sling.py
carlosmunoz: We should discuss this one. I might see the value of the `renderable` field, but at this point we know the file naming convention is a no-go, so why not remove it?

uploader/push_to_sling.py
benradey: My thought for priority was: command line > config file > hardcoded default - for all options.

Technically we have the ability to specify arguments more than once and capture them in an array (I really like this library!), so we COULD let users specify the globs on the command line as well. That seems like gold-plating though - what do you think?

uploader/push_to_sling.py
benradey: Now that I thought about it for a few minutes, I don't like the `pant:renderable` suggestion. We already have the `sling:resourceType` of `pantheon/modules`, now it probably makes sense to introduce `pantheon/titles` and `pantheon/resources` as well, and start keying off that.

uploader/push_to_sling.py
carlosmunoz: makes sense to me :+1:

uploader/push_to_sling.py
carlosmunoz: +1 on not using the extension and trusting users to get it right (or making it very clear why they failed).

But I think our include processor needs to be smart about how to include different types of content (e.g. a `pant:module` vs a `pant:title` vs a regular `nt:file`

uploader/push_to_sling.py
carlosmunoz: Remember there is sling's resource type, which is used for rendering; and JCRs node type, which is used more for hard node types. In this scenario, we should be referring to node types. So `pant:module` and `pant:title`.

uploader/push_to_sling.py
benradey: I just checked and right now it's checking for the existence of an "asciidoc" child node before deciding where to ultimately pull the content from. So it may just work as-is, perhaps not the best implementation, but I think it's good enough for now so I retract my concern.

uploader/push_to_sling.py
benradey: Ok, good thinking, we can use `jcr:nodeType` for this. Agreed.

src/main/resources/SLING-INF/nodetypes/nodetypes.cnd
```diff
@@ -28,6 +28,11 @@
...
 // Base type for modules
-[pant:module] > nt:unstructured, sling:Folder
+[pant:module] > nt:unstructured
 	- sling:resourceType (string) = 'pantheon/modules' mandatory autocreated
     + cachedContent (nt:unstructured) = nt:unstructured mandatory autocreated
+
+// Base type for titles
+[pant:title] > nt:unstructured
+	- sling:resourceType (string) = 'pantheon/titles' mandatory autocreated
+    + cachedContent (nt:unstructured) = nt:unstructured mandatory autocreated
```
benradey: Looks good! Can we also have an entry for `[pant:resource]`?

uploader/pantheon.py
```diff
@@ -0,0 +1,214 @@
...
+
+            logger.debug('base name: %s', base_name)
+
+            # Asciidoc content (treat as a module)
+            if path.suffix == '.adoc' or path.suffix == '.asciidoc':
+                print(path)
+                url += '/' + path.name
+                logger.debug('url: %s', url)
+                jcr_primary_type = "pant:module" if isModule else "pant:title"
+                sling_resource_type = "pantheon/modules" if isModule else "pantheon/titles"
```
benradey: Two comments:
1. Can we also add logic for "pant:resource"?
2. Are we sure it's necessary to specify the sling_resource_type here? It's not hurting anything to have it, but if we can just leave it at it's default value (which we specify via nodetypes.cnd), then I'd rather take that approach.

uploader/pantheon.py
```diff
@@ -0,0 +1,214 @@
...
+                    r = requests.post(url, headers=HEADERS, files=files, auth=(args.user, pw))
+                    print(r.status_code, r.reason)
+                logger.debug('')
+        else:
+            # Ignore the files are not specified in .yml file.
+            unspecified_files += 1
+
+
+if unspecified_files > 0:
+    print (f'{unspecified_files} additional files detected. Only files specified in ' + CONFIG_FILE +' are handled for upload.')
```
benradey: Ah, this is a good thought!

src/main/java/com/redhat/pantheon/servlet/BulkDeleteServlet.java
```diff
@@ -64,42 +64,42 @@ public BulkDeleteServlet() {
...
-			resourceResolver.revert();
+		if (request.getResourceResolver().hasChanges()) {
+			request.getResourceResolver().revert();
 			String commitError = "Something unexpected happened. Message was: " + e.getMessage();
 			response.sendError(HttpServletResponse.SC_BAD_REQUEST, commitError);
 		}
         } finally {
         	// When done, close the ResourceResolver.
-        	resourceResolver.close();
+        	request.getResourceResolver().close();
```
benradey: I'm confused why these changes are part of this branch. I went through your commits and I get the impression that this code was throwing exceptions when you ran the python script? If that's so, that's a bit of a concern... I would rather find some mechanism where the delete servlet isn't invoked at all for anything other than a delete operation. However that must have been an existing issue, so for this branch specifically, we'll leave it as-is.

src/main/resources/SLING-INF/nodetypes/nodetypes.cnd
```diff
@@ -31,3 +31,8 @@
 [pant:module] > nt:unstructured
 	- sling:resourceType (string) = 'pantheon/modules' mandatory autocreated
     + cachedContent (nt:unstructured) = nt:unstructured mandatory autocreated
+
+// Base type for titles
+[pant:title] > nt:unstructured
```
benradey: Please look at my most recent changes for `pant:module` in the master branch and reflect them here for `pant:title` also.

uploader/pantheon.py
benradey: Same comment as before here - can we remove `sling_resource_type` here? This should get autocreated for us via `nodetypes.cnd`.

src/main/java/com/redhat/pantheon/servlet/BulkDeleteServlet.java
xdavidson: bulkDeleteServlet get registered when uploader pushes the same repo over and over. I scheduled a a discussion session on this topic this afternoon.

uploader/pantheon.py
xdavidson: done

src/main/java/com/redhat/pantheon/servlet/BulkDeleteServlet.java
benradey: Cool, I saw that. That's going to be a significant change though, so let's create a separate branch if we're going to be reworking the delete servlet.

benradey:
Lisa, looks good to me! Consider this "approved," but github won't let me approve it since technically this is my PR. Someone else will have to actually hit the button. Perhaps @jmarrero would be good enough to weigh in and (hopefully) add his approval?

benradey:
Yeah Lisa and I had the same discussion back when she first started work on the delete servlet. I'm not sure that it's possible to get the browser to do anything other than GET and POST. I'd like to know for sure, but FWIW, even the sling ootb admin repo browser accomplishes its deletes via POST, so that's pretty telling.

PR #29, opened 2019-05-06
Initial frontend import

jmarrero:
I think this is a good starting point for the react front end.

The frontend is accessible on localhost:8080/pantheon

Going to create a few issues to work on top of this to achieve:
1) Parity with features on htl/html frontend (sort and delete)
2) Change tables for datalist and enable sort.
3) Separate frontend into it's own bundle or deployment script. 
4) Move the header/router to a sidebar component

but I believe this pull request accomplishes CCS-2705 requirements.


frontend/.prettierrc
```diff
@@ -0,0 +1,4 @@
+{
+  "singleQuote": true,
+  "printWidth": 120
+}
```
benradey: Should these 3 `frontend/.*` files be included in .gitignore?

jmarrero: Great questions, these come from https://github.com/patternfly/patternfly-react-seed which I used as a base. I did not remove anything used by them I just built on top of it.

frontend/src/app/app.test.tsx
```diff
@@ -0,0 +1,25 @@
...
+  });
+
+  it('should hide the alert when clicking the dismiss button', () => {
+    const wrapper = mount(<App />);
+    const button = wrapper.find(Button);
+    expect(wrapper.find(Alert).exists()).toBe(true);
+    button.simulate('click');
+    expect(wrapper.find(Alert).exists()).toBe(false);
+  });
+});
```
benradey: Just making sure, this is an ootb test example, right? Not specific to our code at all?

jmarrero: Yes

frontend/.prettierrc
jmarrero: It looks like they are not essential to the code itself. I don't have a preference on removing them or not.

frontend/src/app/module.tsx
```diff
@@ -0,0 +1,103 @@
...
+    console.log("My name is: " + this.state.moduleName + " and my desc is " + this.state.moduleDescription + " and my files are " + this.state.moduleFile)
+
+
+    const hdrs = {
+      'cache-control': 'no-cache',
+      'Accept': 'application/json'
+    }
+    console.log("The file is: " + this.state.moduleFile)
+    const blob = new Blob([this.state.moduleFile[0]])
+    const formData = new FormData();
```
benradey: Heyyyy! At least this piece I recognize... :)

frontend/yarn.lock
```diff
@@ -0,0 +1,8938 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
benradey: I'm betting this should also be in .gitignore

frontend/src/app/app.test.tsx
jmarrero: Updated the tests to check for some of the components we are using. Thanks for pointing this out.

frontend/yarn.lock
jmarrero: https://yarnpkg.com/blog/2016/11/24/lockfiles-for-all/

.lock files in for npm and yarn are recommended to be committed as they tell you what was the versions of the dependencies your were using. This will track all dependencies not just the ones we declare on package.json but the "dependencies of the dependencies".

benradey: I stand corrected - thanks for clarifying!

frontend/.prettierrc
jmarrero: Updated the gitignore and removed these files.

jmarrero:
Thanks @benradey I updated the PR based on your comments and like we talked about. We will surely will continue to build on top of this and I know there are a lot of improvements to do, but this should accomplish 2705. BTW thanks for the help on the upload and backend.

@ivanhorvath Thanks :)

PR #31, opened 2019-05-07
add unit tests and move others

carlosmunoz:
Move a unit test to a different class as it was testing methods for the wrong class.
Also create a new unit test class for the `ModuleDataRetriever` class with some basic unit tests.


carlosmunoz:
Agreed. The tests in their current form are not providing a lot of value. But two things: 1. writing tests along with the classes will provide some enforcement of good practices as the team writes the code for testability; and 2. We should get into the habit of writing unit tests for any newly created class.

PR #35, opened 2019-05-09
Address QA concerns

jmarrero:
-Search by pressing enter.
-Get source name from transientSourceName.
-Fix reference problem for Firefox.
-Make date human readable.
-Inform the user when there are no results from the search.
-Validate that name and file are not empty.

### DIFF
--- a/frontend/src/app/index.tsx
+++ b/frontend/src/app/index.tsx
@@ -1,5 +1,5 @@
 import React, { Component } from 'react';
-import { TextInput, Label, BackgroundImage, BackgroundImageSrc } from '@patternfly/react-core';
+import { Alert, AlertActionCloseButton, TextInput, Label } from '@patternfly/react-core';
 import { Table, TableHeader, TableBody } from '@patternfly/react-table';
 import '@app/app.css';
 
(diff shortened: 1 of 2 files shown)


frontend/src/app/index.tsx
```diff
@@ -36,31 +47,60 @@ export default class Index extends Component {
...
   };
 
+  private formatDate(date: Date) {
+    //2019/05/07 14:21:36
+    var dateStr = date.getFullYear().toString() + "/" +
+      date.getMonth().toString() + "/" +
+      date.getDay().toString() + " " +
+      date.getHours().toString() + ":" +
+      date.getMinutes().toString() + ":" +
+      date.getSeconds().toString()
```
benradey: Curious if there's a react equivalent to java's SimpleDateFormatter, but this certainly works.

jmarrero: There are external libraries. That do date formatting, but for this simple format I tried to avoid adding a dependency. If we need something more complex maybe we can look at something like:
https://www.npmjs.com/package/react-moment
https://stackoverflow.com/questions/34590369/formatting-a-date-string-in-react-native

PR #37, opened 2019-05-10
add a user group with restricted access

carlosmunoz:
The access to the new group is restricted to the content/repositories
directory.
Also remove unnecesary entries in the cnd file as well as the corresponding
content entries, which were more experimental.


src/main/java/com/redhat/pantheon/sling/PantheonRepositoryInitializer.java
```diff
@@ -41,22 +41,29 @@ public void processRepository(SlingRepository slingRepository) throws Exception
     private void initializeRepositoryACLs(SlingRepository slingRepository) throws RepositoryException {
         JackrabbitSession s = (JackrabbitSession) slingRepository.loginAdministrative(null);
         try {
-            User admin = (User) s.getUserManager().getAuthorizable("admin");
-            admin.changePassword("ccsadmin"); // FIXME - hardcoding admin passwords is a Bad Thing
```
benradey: Did you remove this because you reimplemented it elsewhere, or did you just want it gone?

src/main/java/com/redhat/pantheon/sling/PantheonRepositoryInitializer.java
```diff
@@ -41,22 +41,29 @@ public void processRepository(SlingRepository slingRepository) throws Exception
...
-
+            // Create and give the pantheon service user permissions to the whole /content path
             try {
                 s.getUserManager().createSystemUser("pantheon", null);
+                s.save();
```
benradey: Do you need this s.save() here, since we've already got an s.save() coming later in the method?

benradey: Ah, perhaps it's so that you can catch the `AuthorizableExistsException` separately from the rest of the logic?

src/main/java/com/redhat/pantheon/sling/PantheonRepositoryInitializer.java
```diff
@@ -41,22 +41,29 @@ public void processRepository(SlingRepository slingRepository) throws Exception
...
             }
+            assignPermissionToPrincipal(s, "pantheon", "/content", "*", Privilege.JCR_ALL);
+
+            // Create and grant permissions to the pantheon-users group
+            try {
+                s.getUserManager().createGroup("pantheon-users");
+                s.save();
+                log.info("Created pantheon-users group");
+            } catch (AuthorizableExistsException aeex) {
+                log.info("pantheon-users group already exists");
```
benradey: I see this is likely the same logic as above - try/catch and s.save() this separately - right?

src/main/java/com/redhat/pantheon/sling/PantheonRepositoryInitializer.java
carlosmunoz: I just wanted it gone :blush:  (I just think we shouldn't be hardcoding passwords in the code)

src/main/java/com/redhat/pantheon/sling/PantheonRepositoryInitializer.java
carlosmunoz: I need the call to save here so the call in line 52 is able to "see" the changes. Otherwise, that call would end up throwing an error because it doesn't recognize the 'pantheon' principal. It acts like a bit of a flush for the operations on the JCR repository.

src/main/java/com/redhat/pantheon/sling/PantheonRepositoryInitializer.java
carlosmunoz: Yes, the only reason to separate them is to be able to log specific entries for each operation attempted. Should give us better insight into what the system is doing.

src/main/java/com/redhat/pantheon/sling/PantheonRepositoryInitializer.java
carlosmunoz: ...as you also realized in your FIXME comment.

benradey: Yes, hardcoding it was Bad, but it was Less Bad than leaving it as the default!

Well, it's a small issue, especially if we're sending the users through our custom login form that doesn't suggest admin:admin right next to the login fields...

I'm sure there is a "proper" solution that we'll figure out someday, but perhaps we should record that as a jira issue so we don't forget about it.

src/main/java/com/redhat/pantheon/sling/PantheonRepositoryInitializer.java
benradey: Ah, that also makes sense, thanks!

PR #39, opened 2019-05-14
React login

benradey:


jmarrero:
Looks good, bulds and deploys ok, new UI page looks good and I can see the login text change to logout when I logged in.

However modules.tsx is still pointing to old sling login UI:
`      return window.location.assign("/system/sling/login.html");
`
 Also when I clicked logout I did not get the login option I had to manually go back to /pantheon. 

I also want to understand if we need this page for the login to work from react or you just added it for the old UI sake.
```
src/main/resources/apps/pantheon/welcome.html
```

benradey:
> I also want to understand if we need this page for the login to work from react or you just added it for the old UI sake.
> 
> ```
> src/main/resources/apps/pantheon/welcome.html
> ```

That was just me playing around with a custom login form in a less-foreign (i.e., not React) environment. Technically it's throwaway and doesn't need to be part of the PR.

PR #41, opened 2019-05-16
[CCS-2790] hide delete button when none of the modules is selected

xdavidson:
Address QA comments on Delete button submission exception when none of the modules is selected.

### DIFF
--- a/src/main/resources/apps/pantheon/modules/all/html.html
+++ b/src/main/resources/apps/pantheon/modules/all/html.html
@@ -43,7 +43,7 @@
                     ${item.pant:transientSourceName}
                 </td>
                 <td>
-                    <sly data-sly-use.date="${'com.redhat.pantheon.use.DateFormatter' @ format='yyyy/MM/dd HH:mm:ss', date=item.jcr:created}">${date.value}</sly>
+                    <!-- <sly data-sly-use.date="${'com.redhat.pantheon.use.DateFormatter' @ format='yyyy/MM/dd HH:mm:ss', date=item.jcr:created}">${date.value}</sly> -->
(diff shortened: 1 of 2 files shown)


src/main/resources/apps/pantheon/modules/all/html.html
```diff
@@ -43,7 +43,7 @@ <h1>All Modules</h1>
                     ${item.pant:transientSourceName}
                 </td>
                 <td>
-                    <sly data-sly-use.date="${'com.redhat.pantheon.use.DateFormatter' @ format='yyyy/MM/dd HH:mm:ss', date=item.jcr:created}">${date.value}</sly>
+                    <!-- <sly data-sly-use.date="${'com.redhat.pantheon.use.DateFormatter' @ format='yyyy/MM/dd HH:mm:ss', date=item.jcr:created}">${date.value}</sly> -->
```
benradey: Why is this commented? If this was just a mistake then can we bring it back?

xdavidson: for some reason, modules.html throws 500 error on com.redhat.pantheon.use.DateFormatter. This branch is based off master. I commented it out so that I can see the modules.html page for the delete button. Let me uncomment it.

xdavidson: done

benradey: Hmm, well that's worrisome, but we can troubleshoot it later if necessary. Thanks for explaining!

jmarrero:
I built the latest changes and tested them, and it looks good to me.

PR #44, opened 2019-05-21
Upload/preview fix

benradey:
Moved creation of "pantheon-users" group into users.json and included "demo" user for group membership, fixing upload issue. Re-added "inline=true" for maven dependency handling, fixing preview issue.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -83,6 +83,7 @@
                             !org.apache.tools.ant,!org.joda.convert,!org.objectweb.asm,!sun.misc,*
                         </Import-Package>
                         <!-- Embed the necessary dependencies -->
+                        <!-- The `inline=true` is important, don't remove it unless you want cryptic problems such as servlets that suddenly stop resolving without explanation -->
                         <Embed-Dependency>*;scope=compile;inline=true</Embed-Dependency>
                         <Embed-Transitive>true</Embed-Transitive>
                         <Require-Capability>osgi.ee;filter:="(&amp;(osgi.ee=JavaSE)(version=1.8))"</Require-Capability>
--- a/pom.xml
+++ b/pom.xml
@@ -83,7 +83,7 @@
                             !org.apache.tools.ant,!org.joda.convert,!org.objectweb.asm,!sun.misc,*
                         </Import-Package>
                         <!-- Embed the necessary dependencies -->
-                        <Embed-Dependency>*;scope=compile</Embed-Dependency>
+                        <Embed-Dependency>*;scope=compile;inline=true</Embed-Dependency>
                         <Embed-Transitive>true</Embed-Transitive>
                         <Require-Capability>osgi.ee;filter:="(&amp;(osgi.ee=JavaSE)(version=1.8))"</Require-Capability>
                     </instructions>
--- a/src/main/java/com/redhat/pantheon/sling/PantheonRepositoryInitializer.java
+++ b/src/main/java/com/redhat/pantheon/sling/PantheonRepositoryInitializer.java
@@ -50,16 +50,8 @@ public class PantheonRepositoryInitializer implements SlingRepositoryInitializer
                 log.info("Pantheon service account already exists");
             }
             assignPermissionToPrincipal(s, "pantheon", "/content", "*", Privilege.JCR_ALL);
-
-            // Create and grant permissions to the pantheon-users group
-            try {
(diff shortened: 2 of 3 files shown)


pom.xml
```diff
@@ -83,7 +83,8 @@
                             !org.apache.tools.ant,!org.joda.convert,!org.objectweb.asm,!sun.misc,*
                         </Import-Package>
                         <!-- Embed the necessary dependencies -->
-                        <Embed-Dependency>*;scope=compile</Embed-Dependency>
+                        <!-- The `inline=true` is important, don't remove it unless you want cryptic problems such as servlets that suddenly stop resolving without explanation -->
+                        <Embed-Dependency>*;scope=compile;inline=true</Embed-Dependency>
```
carlosmunoz: I'm curious about this. The only noticeable difference for me is the way the dependencies are packaged. With `inline=true` everything is just expanded in the jar file, while the opposite makes it neatly arranged in jar files (except for our own code) with the appropriate entries in the manifest files. 

Are we sure this was causing servlets to stop resolving? If so, I don't mind keeping `inline=true` as long as we assume it as tech debt to fix down the line.

src/main/java/com/redhat/pantheon/sling/PantheonRepositoryInitializer.java
```diff
@@ -50,16 +50,8 @@ private void initializeRepositoryACLs(SlingRepository slingRepository) throws Re
...
-            try {
-                s.getUserManager().createGroup("pantheon-users");
-                s.save();
-                log.info("Created pantheon-users group");
-            } catch (AuthorizableExistsException aeex) {
-                log.info("pantheon-users group already exists");
-            }
-            assignPermissionToPrincipal(s,"pantheon-users","/content/repositories", null, Privilege.JCR_ALL);
+            assignPermissionToPrincipal(s,"pantheon-users","/content/repositories", null, Privilege.JCR_WRITE, Privilege.JCR_NODE_TYPE_MANAGEMENT);
+            assignPermissionToPrincipal(s,"pantheon-users","/content/modules", null, Privilege.JCR_WRITE, Privilege.JCR_NODE_TYPE_MANAGEMENT);
```
carlosmunoz: I think we need to decide where we want to keep our content. Either `repositories` or `modules` (personally think it should be the former).

src/main/resources/SLING-INF/users/users.json
```diff
@@ -1,5 +1,6 @@
 {
   "security:principals": [
-    { "name": "demo", "password": "demo" }
+    { "name": "demo", "password": "demo" },
+    { "name": "pantheon-users", "isgroup": true, "members": ["demo"] }
```
carlosmunoz: nice :+1: it's a pity service users cannot be created using this mechanism.

pom.xml
benradey: I am 100% sure this is the cause.

I am 30% sure that I know the reason why, with a lot of speculation and hand-waving involved.

Remember that we needed to do a lot of work to manually locate the ruby gems necessary in order to instantiate an Asciidoctor instance. Because of the way Asciidoctorj is written, we can only do that successfully with `inline=true`.

My guess is that when Sling is trying to figure out what servlets exist, it has to instantiate a copy of the servlet in order to inspect it. (This is where things get reeeally hand-wavy). When it tries to do that without `inline=true`, the servlet fails to initialize due to Asciidoctor throwing some exception and as a result Sling pretends it doesn't even exist.

Thus when the user requests anything.preview, Sling sees it has 0 servlets registered to handle .preview requests and throws a 404.

Giant hole in this theory: none of our rendering servlet loggers were printing anything, not even the ones before we try and get an Asciidoctor instance.

I'd love to know what's really going on here, but I can only guess for now. "Too much magic" and all that.

src/main/java/com/redhat/pantheon/sling/PantheonRepositoryInitializer.java
benradey: Yes, we need to decide that. My concern for this PR was getting the codebase working again with as few changes as possible, not to finalize our folder structure.

pom.xml
carlosmunoz: I think you are right... it's exactly that. I wonder if we can either a. extract the files from a jar file, or 2. just inline that specific dependency. As I said, for the time being we should take this one as tech debt, and try to figure it out later.

PR #45, opened 2019-05-22
introduce a default query servlet for any node

carlosmunoz:
Introduce a default query servlet. This means all users can make use of this servlet provided the user has JCR repository access to the individual resource. The servlet allows to query resources **under** the selected node, providing filters, ordering clauses and pagination. The maximum page size is set to 1000.

How to use it?
For any resource in the sling repository (e.g `/content/folder/node`) do a GET request with selector `query` and extension `json`. Example: `GET /content/folder/node.query.json` 

Available (and optional) query parameters include:

`where` - A JCR where clause to filter our results
`orderBy` - A JCR order by clause to order results
`nodeType` - A JCR node type name (only resources with this primary type will be returned)
`limit` - Limits the number of results to be returned (max is 1000)
`offset` - A starting position to be able to do paging

The result will return a json object of the type:

```json
{"results": [...], "size": 10, "nextOffset": 20}
```


src/test/java/com/redhat/pantheon/servlet/DefaultQueryServletTest.java
```diff
@@ -0,0 +1,203 @@
...
+        slingContext.request().setResource(slingContext.resourceResolver().getResource("/content/test"));
+
+        // When
+        servlet.doGet(slingContext.request(), slingContext.response());
+
+        // Then
+        JsonNode jsonNode = new ObjectMapper().readTree(slingContext.response().getOutputAsString());
+        assertEquals(10L, jsonNode.get("size").asLong());
+        assertEquals(10L, jsonNode.get("nextOffset").asLong());
+        assertTrue(jsonNode.get("results").isArray());
```
benradey: I really like this Given-When-Then pattern. It makes the tests very easy to read and understand. Thanks for setting this good example!

src/test/java/com/redhat/pantheon/servlet/DefaultQueryServletTest.java
```diff
@@ -0,0 +1,203 @@
...
+        for (int i = 0; i < 10; i++) {
+            builder.resource("/content/test/node-" + i,
+                    "name", "node-" + i,
+                    "number", i);
+        }
+        builder.commit();
+
+        DefaultQueryServlet servlet = new DefaultQueryServlet();
+        servlet.init();
+        slingContext.request().setResource(slingContext.resourceResolver().getResource("/content/test"));
```
benradey: We seem to be repeating the above dozen lines; is it worth it to pull it out into some kind of testSetup method?

src/test/java/com/redhat/pantheon/servlet/DefaultQueryServletTest.java
```diff
@@ -0,0 +1,203 @@
...
+        // When
+        servlet.doGet(slingContext.request(), slingContext.response());
+
+        // Then
+        JsonNode jsonNode = new ObjectMapper().readTree(slingContext.response().getOutputAsString());
+        assertEquals(5L, jsonNode.get("size").asLong());
+        assertEquals(5L, jsonNode.get("nextOffset").asLong());
+        assertTrue(jsonNode.get("results").isArray());
+        for (int i = 0; i < 5; i++) {
+            assertTrue(slingContext.response().getOutputAsString().contains("node-" + i));
```
benradey: I would change this to just verify that the output contains 5 nodes. If you want to be specific about which 5 nodes it contains, then I would specify a sort order as well.

src/test/java/com/redhat/pantheon/servlet/DefaultQueryServletTest.java
```diff
@@ -0,0 +1,203 @@
...
+        // When
+        servlet.doGet(slingContext.request(), slingContext.response());
+
+        // Then
+        JsonNode jsonNode = new ObjectMapper().readTree(slingContext.response().getOutputAsString());
+        assertEquals(5L, jsonNode.get("size").asLong());
+        assertEquals(10L, jsonNode.get("nextOffset").asLong());
+        assertTrue(jsonNode.get("results").isArray());
+        for (int i = 5; i < 10; i++) {
+            assertTrue(slingContext.response().getOutputAsString().contains("node-" + i));
```
benradey: Same comment here, probably wise to specify a sort order.

src/main/java/com/redhat/pantheon/servlet/DefaultQueryServlet.java
```diff
@@ -0,0 +1,107 @@
...
+        String nodeType = paramValue(request, "nodeType","nt:base");
+        String where = paramValue(request, "where");
+        String orderBy = paramValue(request, "orderBy");
+        long limit = paramValueAsLong(request, "limit", RESULT_SIZE_LIMIT);
+        long offset = paramValueAsLong(request, "offset", 0L);
+
+        StringBuilder query = new StringBuilder("select * from ")
+                .append("[").append(nodeType).append("]")
+                .append(" as node ")
+                .append(" where ISDESCENDANTNODE(\"" + request.getResource().getPath() + "\") ");
```
benradey: How is the user prevented from querying for nodes that are outside of the /content path? It looks like this might list all child nodes for any resource in the system, not just those that are in /content. If we're relying on Sling ACLs for that, then we should be diligent about knowing exactly what nodes users are able to query for using this tool.

(On that note, it's probably worth it to add some negative unit tests as well - to ensure that users can't get a listing of nodes that we know to be sensitive - underneath /home/users for example...)

src/main/java/com/redhat/pantheon/servlet/ServletUtils.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    /**
+     * Returns a string servlet request parameter, or the provided default value if the parameter was not present.
+     * @param request
+     * @param paramName
+     * @param defaultValue
+     * @return
+     */
+    public static @Nonnull String paramValue(final HttpServletRequest request, final String paramName,
```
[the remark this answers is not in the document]
carlosmunoz: The compiler and IDEs can do a few checks (in the form of warnings I believe) based on these annotations. Overall I've always found them useful to specify method contracts.

src/main/java/com/redhat/pantheon/servlet/ServletUtils.java
```diff
@@ -0,0 +1,94 @@
...
+    /**
+     * Returns a servlet request parameter as a Long, or null if the parameter was not present or
+     * unparseable to a Long value.
+     * @param request
+     * @param paramName
+     * @return
+     */
+    public static @Nullable
+    Long paramValueAsLong(final HttpServletRequest request, final String paramName) {
+        return paramValueAsLong(request, paramName, null);
```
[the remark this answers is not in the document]
carlosmunoz: you are absolutely right. :+1: for a good code review :smile: 
This method should also specify nullability. I'll fix that.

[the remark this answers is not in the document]
carlosmunoz: Good catch! :+1: I think this forces the other method to be `@Nullable` too

src/test/java/com/redhat/pantheon/servlet/DefaultQueryServletTest.java
carlosmunoz: Yes it is. I'll fix this.

src/main/java/com/redhat/pantheon/servlet/DefaultQueryServlet.java
carlosmunoz: The access to the resource itself is in fact controlled by the sling ACLs. So only users with read access to the node will be able to use this servlet. Additionally, the `ISDESCENDANTNODE` clause (which is hard coded) restricts all returned results to nodes under the requested one. So, in your example, because `/home/users` is a restricted node (by ALCs) no user would be able to get to this servlet.

The only problem I foresee is if for some reason the child nodes are restricted (as in not even viewable) by the user, but not the parent node, and hence said restricted nodes could be returned by this servlet.

src/test/java/com/redhat/pantheon/servlet/DefaultQueryServletTest.java
carlosmunoz: I will take out the common JCR repository prep code, but I'll leave some individual lines just to keep the tests independent.

src/test/java/com/redhat/pantheon/servlet/DefaultQueryServletTest.java
carlosmunoz: I agree this is unnecessary (and not really apt to this test), but I'm testing sorting in a different test so I'll keep that logic there.

carlosmunoz:
> At a high level, what do you see this code being used for? (Not saying we won't find a need for it sooner or later - I'm sure we will - but curious if you have anything specific/immediate in mind.)

The idea is that instead of building specific servlets to query for data, this generic servlet will be able to serve most of the use cases where we would need from the UI. As an example, our module search screen would be able to use this to search for modules, and tweak the query itself if needed without any changes to the back end.

carlosmunoz:
ready for another review

src/test/java/com/redhat/pantheon/servlet/DefaultQueryServletTest.java
carlosmunoz: :+1: removed the check

src/main/java/com/redhat/pantheon/servlet/ServletUtils.java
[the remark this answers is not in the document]
benradey: Good to know. Thanks!

src/main/java/com/redhat/pantheon/servlet/DefaultQueryServlet.java
benradey: I would imagine that the query is still running in the context of the user, so it wouldn't return any restricted child nodes.

I'm still uneasy about this because I don't know if we're opening up a security concern, but it's mostly because I'm unfamiliar with 1.) how sling's acls are set up ootb, and 2.) what data even exists that might be considered sensitive. I'll ignore these concerns for now - my favorite phrase "don't solve problems you don't have yet" comes to mind.

benradey:
Yep, I also wondered if we could replace the default search servlet with this. Now that I'm thinking about it though, the search servlet does some extra work that involves slapping a few extra fields onto the objects before returning them (i.e. identifying the repository name they are contained in), so this wouldn't be a drop-in replacement.

It's a good idea to have though, I'm sure we'll find a use for it!

carlosmunoz:
OK, I'll merge and we can revisit later if needed.
As a solution to the security issue, we could set up a blacklist mechanism to restrict which nodes (if any) need to be left out of the results. For example: having a `pant:queryRestricted` property on blacklisted nodes. The servlet would filter these out. It's a sledgehammer approach (as in not controlled on a per-user basis), but it could provide a basic mechanism hide private data.

PR #48, opened 2019-05-23
Initial sidebar implementation

jmarrero:
- changed app/index to app/search
- Added sidebar, header and route components using https://github.com/seanforyou23/draftcab as starting point. 
- Using router-react-dom 5.0
- Now the search loads with the application.
- Will create different PR for tracking the user logged in work that Ben originally did for the previous header.
- removed unused deps.

### DIFF
--- a/frontend/src/app/app.test.tsx
+++ b/frontend/src/app/app.test.tsx
@@ -1,17 +1,18 @@
 import React from 'react';
-import Index from '@app/index';
+import Search from '@app/search';
+import "isomorphic-fetch"
 
 import { mount, shallow } from 'enzyme';
 import { Label } from '@patternfly/react-core';
 
 describe('App tests', () => {
   test('should render default App component', () => {
-    const view = shallow(<Index />);
+    const view = shallow(<Search />);
     expect(view).toMatchSnapshot();
   });
 
   it('should render a search Label', () => {
-    const wrapper = mount(<Index />);
+    const wrapper = mount(<Search />);
     const button = wrapper.find(Label);
     expect(button.exists()).toBe(true)
   });
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -23,6 +23,7 @@
     "@types/node": "^11.9.6",
     "@types/react": "^16.8.6",
     "@types/react-dom": "^16.8.2",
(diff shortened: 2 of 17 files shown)


frontend/src/app/app.test.tsx
```diff
@@ -1,17 +1,18 @@
 import React from 'react';
-import Index from '@app/index';
+import Search from '@app/search';
+import "isomorphic-fetch"
 
 import { mount, shallow } from 'enzyme';
 import { Label } from '@patternfly/react-core';
 
 describe('App tests', () => {
```
carlosmunoz: @jmarrero I just saw that we have tests in the react code :open_mouth: :+1: ... how do we run them? I'd like to get them incorporated into our CI build and even our coverage report if possible.

jmarrero: We run it by running "yarn test" on the frontend folder. 

adding:

```
      <execution>
            <id>yarn-test</id>
            <goals>
                <goal>yarn</goal>
            </goals>
            <configuration>
                <arguments>test</arguments>
            </configuration>
        </execution>
```

to line 192 of the pom.xml should be enough to run it as part of the build.

jmarrero: Let me know if you want me to add this on this PR or if you wanted to do the yarn test in the CI script

frontend/src/app/app.test.tsx
```diff
@@ -1,17 +1,18 @@
...
   test('should render default App component', () => {
-    const view = shallow(<Index />);
+    const view = shallow(<Search />);
     expect(view).toMatchSnapshot();
   });
 
   it('should render a search Label', () => {
-    const wrapper = mount(<Index />);
+    const wrapper = mount(<Search />);
```
benradey: Nice, the Index thing had kind of been bugging me as badly named. I know this is a small thing to call out but I appreciate the small details!

frontend/src/app/app.test.tsx
carlosmunoz: I don't want to increase the scope of the PR, so lets tackle this separately.
Thanks for the info on how to run it. I will take a look at integrating it with our CI builds.

PR #50, opened 2019-05-24
checkboxes added with delete functionality

ankkit24:


frontend/src/app/__snapshots__/app.test.tsx.snap
```diff
@@ -85,6 +85,20 @@ exports[`App tests should render default App component 1`] = `
               className=""
               dataListCells={
                 Array [
+                  <DataListCell
+                    alignRight={false}
+                    className=""
+                    isFilled={true}
+                    isIcon={false}
+                    width={1}
```
benradey: I'm curious about this name={value} syntax. I haven't seen it anywhere else before, but I think I have only worked with name="stringvalue" so far. Are the brackets necessary for non-string types?

frontend/src/app/__snapshots__/app.test.tsx.snap
```diff
@@ -219,6 +254,10 @@ exports[`App tests should render default App component 1`] = `
               }
             />
           </DataListItemRow>
+          <DataListItemRow
+            className=""
+            id="data-rows"
+          />
```
benradey: Does this empty row serve a purpose?

frontend/src/app/search.tsx
```diff
@@ -17,8 +18,15 @@ export default class Search extends Component {
     sortKey: '',
     pageOffset: 1,
     pageCount: 5,
-    initialLoad: true
+    initialLoad: true,
+    value: false,
+    show: false,
+    count: 0,
+    flag: false
```
benradey: Can we please name these variables more specifically? Instead of `value`, perhaps `checkedForDelete` (or something along those lines). Same thing for `show` - how about `deleteButtonVisible`? Same for `count` and `flag` - whatever's appropriate.

frontend/src/app/search.tsx
```diff
@@ -98,6 +125,60 @@ export default class Search extends Component {
     );
   }
 
+  private handleChange = (checked, event) => {
```
benradey: Another good example of specific naming - can we rename this to "handleDeleteCheckboxChange" or similar?

frontend/src/app/search.tsx
```diff
@@ -46,44 +54,63 @@ export default class Search extends Component {
...
+                          <DataListCell width={4} key={data["pant:transientPath"]} onClick={() => this.setPreview(data["pant:transientPath"])}>
+                            {data["jcr:description"]}
+                          </DataListCell>,
+                          <DataListCell key={data["pant:transientPath"]} onClick={() => this.setPreview(data["pant:transientPath"])}>
+                            {data["pant:transientSource"]}
+                          </DataListCell>,
+                          <DataListCell key={data["pant:transientPath"]} onClick={() => this.setPreview(data["pant:transientPath"])}>
+                            {data["pant:transientSourceName"]}
+                          </DataListCell>,
+                          <DataListCell key={data["pant:transientPath"]} onClick={() => this.setPreview(data["pant:transientPath"])}>
```
benradey: There is probably a way for us to not have to repeat `onClick={() => this.setPreview(data["pant:transientPath"])}` 5 times. Please find a way to pull that out into a common method.

frontend/src/app/search.tsx
```diff
@@ -98,6 +125,60 @@ export default class Search extends Component {
...
+    });
+  };
+
+  delete = (event, keydata) => {
+    console.log('in the delete function')
+    for(var i=0;i<keydata.length;i++){
+      // console.log(keydata[i])
+      const formData = new FormData();
+      formData.append('module', keydata[i])
+      fetch('/content/'+keydata[i], {
```
benradey: As we discussed this morning, this performs one POST per item deleted, but the goal is to perform one single POST with a multivalued parameter. I know you have fixed this in your latest code, so this is just a reminder for me to look at it in your next PR. :)

frontend/src/app/search.tsx
```diff
@@ -98,6 +125,60 @@ export default class Search extends Component {
...
+        method: 'post',
+        body: formData
+      }).then(response => {
+        if (response.status == 200) {
+          console.log(" Delete Works " + response.status)
+          this.state.flag = true
+          //window.location.href = "/pantheon"
+        } else if (response.status == 403) {
+          this.setState({ authMessage: "Delete failed." })
```
benradey: Again, we already talked about it, but just to write it here... 1.) don't use authMessage, that's for a separate feature, and 2.) use a patternfly widget instead of a javascript alert or console.log.

benradey:
Ankit looks good so far! I noticed a few things, but I know you have another PR on the way here so I'm looking forward to seeing the new stuff!

frontend/src/app/__snapshots__/app.test.tsx.snap
ankkit24: This seems to have been generated automatically by the tests. I have not made the changes.

frontend/src/app/__snapshots__/app.test.tsx.snap
ankkit24: This is the same file created by the test.

frontend/src/app/search.tsx
ankkit24: Acknowledged.
Thank you for catching this. I will make the changes to the names of the variables specific to their functionality.

frontend/src/app/search.tsx
ankkit24: Acknowledged.
Yes, this function needs renaming to the task it performs.

frontend/src/app/search.tsx
ankkit24: Acknowledged.
I will investigate on this and will work on simplifying it and putting it into a common set of cells., so that the onclick() function could be called just once.

frontend/src/app/search.tsx
ankkit24: This has been fixed in the current commit.

frontend/src/app/search.tsx
ankkit24: Acknowledged.
Thanks for the suggestions, and going forward, I will take note of the mentioned points.

jmarrero:
To rebase this is the process I follow:

1) git fetch upstream
2) git rebase upstream/master

If this says that you have conflicts(it will likely do...) you have to resolve them manually with your editor.

3) git add /path/to/the/file/with/conflcts
4) git rebase --continue

Test your changes and finally update your branch. If you have doubts let me know.

PR #52, opened 2019-05-24
per team discussion, we are going to use this instead: https://sling.…

xdavidson:
…apache.org/documentation/bundles/manipulating-content-the-slingpostservlet-servlets-post.html#deleting-multiple-items


jmarrero:
This is ok with me as the initial change, but we need to discuss if we are maintaining the htl UI too or not. 
If we are, this PR needs to include the change to handle the delete with the HTL page too.

Currently we are using the HTL page to do quick tests and to compare behaviors, however keeping it adds an overhead of maintaining it.
Personally, I don't like the idea of maintaining it, as it adds extra work and we are against a tight schedule.

benradey:
I agree with Joseph. Maintaining the HTL is not worth the effort. This PR is good as-is, but it's probably worth getting rid of the HTL in a different PR.

PR #53, opened 2019-05-28
Update Brand with login logic

jmarrero:
- Re-adds the login login that Ben created but into the Brand component in the Header.
- Added some labels for the pagination
- Some random linter fixes


jmarrero:
Ahh, no I was not trying to do that. Just to do the login logic outside the router and on it's own component. 
I think I can do that in a similar way. I'll try that next

PR #55, opened 2019-05-28
Ccs 2790 delete modules using sling post servlet

xdavidson:
bulk delete modules using SlingPostServlet


xdavidson:
We are going to remove the modules.html page and using react-based implementation.

PR #57, opened 2019-05-29
hide module link for unauth users

jmarrero:
- Added logic to show module creation to only authorized users
- Removed subheader that was not used.
- Added some linter fixes.


frontend/src/app/components/Chrome/Header/Brand/Brand.tsx
```diff
@@ -12,8 +13,8 @@ export default class Brand extends Component {
             fetch("/system/sling/info.sessionInfo.json")
                 .then(response => response.json())
                 .then(responseJSON => {
-                    if (responseJSON['userID'] !== 'anonymous') {
-                        this.setState({ linkText: 'Pantheon | Log Out [' + responseJSON['userID'] + ']' })
+                    if (responseJSON[this.state.id] !== 'anonymous') {
+                        this.setState({ linkText: 'Pantheon | Log Out [' + responseJSON[this.state.id] + ']' })
```
benradey: So it looks like you've pulled 'userID' out as a constant, but you've pulled it out into the state object. Was that intentional?

frontend/src/app/components/Chrome/Sidebar/NavLinks.tsx
```diff
@@ -0,0 +1,41 @@
...
+export default class NavLinks extends Component {
+    public state = {
+        id: 'userID',
+        isLoggedIn: false,
+        moduleText: ''
+    };
+
+    public render() {
+        if (!this.state.isLoggedIn) {
+            fetch("/system/sling/info.sessionInfo.json")
```
benradey: I'm sure this works and it's fine for now, but I'm betting we make this fetch call once in Brand and again here. It would be cleaner if we could just make the call once and share the data somehow.

frontend/src/app/components/Chrome/Sidebar/Sidebar.tsx
```diff
@@ -3,28 +3,18 @@ import { NavLink } from 'react-router-dom';
...
-      );
-    });
-  }
-
   const nav = () => (
     <aside className="pf-c-page__sidebar">
       <nav className="pf-c-nav" id="page-layout-default-nav-primary-nav" aria-label="Primary Nav Default Example">
         <ul className="pf-c-nav__list">
-          {outputNavLinks()}
+          <NavLinks />
```
benradey: :+1:

frontend/src/app/components/Chrome/Sidebar/NavLinks.tsx
```diff
@@ -0,0 +1,39 @@
...
+        }
+
+        return (
+            <React.Fragment>
+                <li>
+                    <Link to='/search'>
+                        {this.state.searchText}
+                    </Link>
+                </li>
+                <li>
```
benradey: Would it make sense to change this to `{this.state.moduleText.length > 0 && <li>...`?

frontend/src/app/components/Chrome/Header/Brand/Brand.tsx
jmarrero: https://stackoverflow.com/questions/33387090/how-to-rewrite-code-to-avoid-tslint-object-access-via-string-literals

This is a change to handle an tslint rule without disabling it.

frontend/src/app/components/Chrome/Sidebar/NavLinks.tsx
jmarrero: maybe I am not sure if it makes a difference performance wise, but I don't think it would hurt us either way. So I'll add it.

frontend/src/app/components/Chrome/Sidebar/NavLinks.tsx
jmarrero: Investigating how to do this.

frontend/src/app/components/Chrome/Header/Brand/Brand.tsx
jmarrero: I changed it to a const in the method. I overlooked your concern before.

PR #62, opened 2019-06-04
Testing maven yarn script

benradey:
Please ignore for now

### DIFF
--- a/frontend/src/app/__snapshots__/app.test.tsx.snap
+++ b/frontend/src/app/__snapshots__/app.test.tsx.snap
@@ -64,7 +64,7 @@ exports[`App tests should render default App component 1`] = `
           onKeyDown={[Function]}
           pattern="[0-9]*"
           type="text"
-          value={55}
+          value={5}
         />
         <Button
           aria-label={null}
--- a/frontend/src/app/__snapshots__/app.test.tsx.snap
+++ b/frontend/src/app/__snapshots__/app.test.tsx.snap
@@ -64,7 +64,7 @@ exports[`App tests should render default App component 1`] = `
           onKeyDown={[Function]}
           pattern="[0-9]*"
           type="text"
-          value={5}
+          value={55}
         />
         <Button
           aria-label={null}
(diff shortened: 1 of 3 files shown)


benradey:
Well that was easy.

benradey:
@jmarrero Ready for review :)

jmarrero:
This works to make the life easier for now, but it will always create a snapshot file when someone makes a change to the UI. 
The idea of the test failing is to flag when someone made a change in the UI which is ok for now to remove but we need to make sure to remove the "-u" later when Pantheon 2 matures into a production project. This is done to make sure we know when someone is altering the UI workflow. 

If we are ok with that, this change should also have an entry to the gitignore like:
frontend/src/app/__snapshots__/*

and 

make sure you run 
```
git rm frontend/src/app/__snapshots__/*
```

and commit the removal of the snapshot.


Reference:
https://jestjs.io/docs/en/snapshot-testing
 which states:  
```
The snapshot artifact should be committed alongside code changes, and reviewed as part of your code review process. Jest uses pretty-format to make snapshots human-readable during code review. On subsequent test runs Jest will simply compare the rendered output with the previous snapshot. If they match, the test will pass. If they don't match, either the test runner found a bug in your code (in this case, it's <Link> component) that should be fixed, or the implementation has changed and the snapshot needs to be updated.
```

jmarrero:
This looks good for me, this still forces people to commit the snap or the CI build will fail.

benradey:
I think I'm happy with this change, this way we can either build the project with our traditional command which will NOT update the snaps:
`mvn clean package sling:install`
or now we can use this new command which WILL update the snaps:
`mvn clean compile frontend:yarn@yarn-update-test package sling:install`

benradey:
@jmarrero Did you have another PR open that was supposed to go along with this one - updating .gitignore or something? Do we need to revert a change from that if we go with this "new target" approach?

jmarrero:
I did not commit and I deleted those changes from my PR after you showed me your new approach, this looks good to me.

PR #65, opened 2019-06-04
Add cursor pointer to list and sorted columns, fix firefox 63 bug and move login link to the right.

jmarrero:


jmarrero:
yep, committed.

jmarrero:
Also added a fix for the delete functionality on older versions of firefox, like the one QA uses (Firefox 63).

jmarrero:
Per CCS-2909 I moved the Login to the right.

frontend/src/app/components/Chrome/Header/User.tsx
```diff
@@ -0,0 +1,39 @@
+import React, { Component } from 'react';
+import { Link } from "react-router-dom";
+
+export default class Brand extends Component {
```
benradey: Should this be `User` instead of `Brand`?

benradey:
I notice also that we only get the finger pointer when hovering over the text of the row now, but we can still click the areas that give the normal pointer and get to the preview anyway. That is a trivial issue that is not worth fixing right now, but maybe it's something we can revisit in the future.

frontend/src/app/components/Chrome/Header/User.tsx
jmarrero: yep this is what Michael talked about the explicit exports, I have changed it now to be explicit so this kind of typos don't happen in these two classes at least. Will work to do the same in all files eventually.

jmarrero:
yeah the css parent child relationship is not strait forward for me yet. Ill try to fix this in another PR.

jmarrero:
Fixed the checkbox thing. Now this includes changes for all your comments.

PR #66, opened 2019-06-04
update PANTHEON_BIN and avoid using sudo

xdavidson:
1. change PANTHEON_BIN to ~/bin
2. avoid using sudo in symlinking

### DIFF
--- a/uploader/setup.sh
+++ b/uploader/setup.sh
@@ -2,7 +2,7 @@
 
 # Variables
 UPLOADER_HOME=~/.pantheon
-PANTHEON_BIN=/usr/local/bin/pantheon
+PANTHEON_BIN=~/bin/pantheon
 
 # Create UPLOADER_HOME directory if not exists
 if [ ! -d "$UPLOADER_HOME" ]; then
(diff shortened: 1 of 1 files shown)


uploader/setup.sh
```diff
@@ -18,11 +18,16 @@ curl -o ${UPLOADER_HOME}/pantheon.py https://raw.githubusercontent.com/redhatacc
 echo "Ensure uploader script is executble"
 chmod +x ${UPLOADER_HOME}/pantheon.py 
 
-# Create a symlink to /usr/local/bin if not exists
+# Create a symlink to ~/bin/pantheon if not exists
 if [ ! -h "$PANTHEON_BIN" ]; then
   echo "Create a symlink to $PANTHEON_BIN"
-  sudo ln -s ${UPLOADER_HOME}/pantheon.py $PANTHEON_BIN
+  ln -s ${UPLOADER_HOME}/pantheon.py $PANTHEON_BIN
```
benradey: I'm not sure it's necessary to put the uploader in `~/.pantheon` and then symlink it, instead of just putting the script itself into `~/bin`, but this doesn't hurt anything. Was there a reason you chose to do it this way?

uploader/setup.sh
```diff
@@ -18,11 +18,16 @@ curl -o ${UPLOADER_HOME}/pantheon.py https://raw.githubusercontent.com/redhatacc
...
 fi
+# Add a new directory into PATH variable
+echo "Adding ${PANTHEON_BIN} to PATH variable"
+PATH=$PATH:$PANTHEON_BIN
+export PATH
+echo $PATH | grep $PANTHEON_BIN
```
benradey: Can you please delete these 5 lines? Here's my thought:

- I'm 90% sure that an `export` all by itself will only last for that terminal session. It won't affect any future terminals nor persist through a system restart, so this does nothing permanent.
- `~/bin` is already in the default path, at least on my machine:
>$ echo $PATH
>/usr/local/bin:/usr/bin:/bin:/usr/local/sbin:/usr/sbin:/home/ben/.local/bin:/home/ben/bin

So I don't believe that these lines do anything meaningful.

xdavidson: good point. removed!

uploader/setup.sh
xdavidson: ~/.pantheon today only has pantheon.py, in the future we can also place other helpful files in there. E.g a README with helpful info that's not covered in `pantheon -h`

benradey: Ok, that makes sense!

PR #67, opened 2019-06-07
Add: pagination functionality

ankkit24:


src/main/java/com/redhat/pantheon/servlet/ModuleListingServlet.java
```diff
@@ -39,16 +39,27 @@
     protected void doGet(@Nonnull SlingHttpServletRequest request, @Nonnull SlingHttpServletResponse response) throws ServletException, IOException {
         ModuleDataRetriever mdr = new ModuleDataRetriever(request.getResourceResolver());
         String searchParam = getParam(request, "search");
+        String countParam = getParam(request, "search");
```
benradey: I suspect this is a typo and that you intended to get the param of `count` instead, but I think we shouldn't need a count param at all. I see that this variable is unused, so please remove it.

src/main/java/com/redhat/pantheon/servlet/ModuleListingServlet.java
```diff
@@ -39,16 +39,27 @@
...
         String directionParam = getParam(request, "direction");
         String offset = getParam(request, "offset");
         String limit = getParam(request, "limit");
 
         try {
-            List<Map<String, Object>> payload = mdr.getModulesSort(searchParam, keyParam, directionParam, offset, limit);
             response.setContentType("application/json");
             Writer w = response.getWriter();
-            w.write(new ObjectMapper().writer().withDefaultPrettyPrinter().writeValueAsString(payload));
+            if(searchParam != ""){
```
benradey: I'm not quite sure what's happening with this `if` statement, but it doesn't seem right. We shouldn't be only returning the row count depending on whether the user supplied a search parameter. Let's talk on Monday about how this ought to look.

src/main/java/com/redhat/pantheon/data/ModuleDataRetriever.java
```diff
@@ -59,6 +59,20 @@ public ModuleDataRetriever(ResourceResolver resolver) {
...
+    public int getRowCount() throws Exception{
+            // String query = "AND (a.[jcr:title] like " + "'%" + query + "%' " +
+            //         "OR a.[jcr:description] like " + "'%" + query + "%') ";
+
+            StringBuilder countQuery = new StringBuilder()
+                .append("select * from [nt:base] as a ")
+                .append("where [sling:resourceType] = 'pantheon/modules' ")
+                .append("and (isdescendantnode(a, '/content/repositories') ")
+                .append("or isdescendantnode(a, '/content/modules') ")
+                .append("or isdescendantnode(a, '/content/sandboxes')) ");
```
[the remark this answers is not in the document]
carlosmunoz: Just a quick comment on this: I agree that `queryRaw` is not the right name for this method, and that we need the where clause in the count query.
As for the `select count()` part, unfortunately there is no way to do that in JCR-QL. This approach (calling `getRowCount()`) is the only way we have to get the total count. I *think* getting the row count does not imply fetching everything in memory. If I'm mistaken, we might need to think of a smart way to do the pagination.

benradey:
Holy smokes, you're right!

https://stackoverflow.com/questions/26400206/how-to-apply-aggregate-functionslike-min-max-count-in-jcr-sql2

Looks like the only way to get the count is manually in the java code. That
stinks, but if it's what we have to do, then so be it...

src/main/java/com/redhat/pantheon/data/ModuleDataRetriever.java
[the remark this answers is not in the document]
benradey: Holy smokes, you're right!

https://stackoverflow.com/questions/26400206/how-to-apply-aggregate-functionslike-min-max-count-in-jcr-sql2

Looks like the only way to get the count is manually in the java code. That stinks, but if it's what we have to do, then so be it...

src/main/java/com/redhat/pantheon/jcr/JcrQueryHelper.java
```diff
@@ -51,6 +53,23 @@ public JcrQueryHelper(ResourceResolver resourceResolver) {
...
+        Query queryObj = queryManager.createQuery(query, Query.JCR_SQL2);
+        QueryResult result = queryObj.execute();
+        NodeIterator nodes = result.getNodes();
+        
+        int rowCount = 0;
+        while (nodes.hasNext()) {
+            Node node = nodes.nextNode();
+            rowCount+=1;
+        }
+        return rowCount;
```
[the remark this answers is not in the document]
benradey: Oops! There is no `select count()`, nevermind... :(

jmarrero:
@ankkit24 this needs to be rebased on top of upstream/master.

frontend/src/app/Pagination.tsx
```diff
@@ -0,0 +1,52 @@
+import React, { Fragment } from 'react';
+import {Button} from '@patternfly/react-core';
+import { Badge } from '@patternfly/react-core';
+
+export interface IProps{
```
jmarrero: Is  "newPage: number" just the page number? If so I would rename that to pageNumber instead 

and "isNextPageRequied: number" is named like a boolean but it's expecting a number. I would either turn this into a boolean or rename it to represent what it is really doing.

frontend/src/app/search.tsx
```diff
@@ -178,41 +203,42 @@ export default class Search extends Component {
...
     console.log('handleSelectAll')
     this.setState({check: !this.state.check}, () => {
       this.setState(prevState => {
+        this.state.allPaths=[]
+        this.transientPaths=[]
         const selectAllcheck = this.state.data.map(dataitem => {
               dataitem["checkedItem"] = this.state.check
               console.log(dataitem["pant:transientPath"]+":"+dataitem["checkedItem"])
-              this.state.check?this.state.allPaths.push(dataitem["pant:transientPath"]):delete this.state.allPaths[this.state.allPaths.indexOf(dataitem["pant:transientPath"])]
+              this.state.check?this.state.allPaths.push(dataitem["pant:transientPath"]):this.state.allPaths.splice(this.state.allPaths.indexOf(dataitem["pant:transientPath"]),1)
```
jmarrero: I see some of the checkbox changes here, make sure you don't overwrite your checkboxes changes that are already on Master. When doing the rebase the new change set should only show the stuff for the pagination.

jmarrero:
It builds
basic functionality works
functions added verified
I can follow the code
test coverage went up

frontend/src/app/Pagination.tsx
ankkit24: Hey Joseph!
I have made the change of name for this variable to 'nextPageRecordCount' which indicates the number of records in the next page.

PR #71, opened 2019-06-17
CCS-2833-only logged in users can use the delete functionality

ankkit24:


frontend/src/app/search.tsx
```diff
@@ -183,36 +203,33 @@ export default class Search extends Component {
     console.log('handleSelectAll')
     this.setState({check: !this.state.check}, () => {
       this.setState(prevState => {
+        this.state.allPaths=[]
+        this.transientPaths=[]
         const selectAllcheck = this.state.data.map(dataitem => {
               dataitem["checkedItem"] = this.state.check
               console.log(dataitem["pant:transientPath"]+":"+dataitem["checkedItem"])
-              this.state.check?this.state.allPaths.push(dataitem["pant:transientPath"]):delete this.state.allPaths[this.state.allPaths.indexOf(dataitem["pant:transientPath"])]
+              this.state.check?this.state.allPaths.push(dataitem["pant:transientPath"]):this.state.allPaths.splice(this.state.allPaths.indexOf(dataitem["pant:transientPath"]),1)
```
benradey: It's hard for me to read what's going on in this line - it looks like if `this.state.check` is true, then we push the transientPath into allPaths. Makes sense. But if `this.state.check` is false, then we splice a value out of allPaths... except we just set `this.state.allPaths` to `[]` 5 lines above, didn't we? So is this really necessary?

frontend/src/app/search.tsx
```diff
@@ -183,36 +203,33 @@ export default class Search extends Component {
...
           return dataitem
         })
         this.transientPaths=this.state.allPaths
-        this.transientPaths.map(e => e === "" ? delete this.transientPaths[this.transientPaths.indexOf(e)] : e)
-        console.log('final transientPaths:'+this.transientPaths)
-        console.log('all paths out:'+this.state.allPaths)
+        this.transientPaths.map(e => e === "" ? this.transientPaths.splice(this.transientPaths.indexOf(e),1) : e)
```
benradey: Curious what this is about. Looks like you're trying to remove elements that have a blank `transientPath`. I would say that we should have no such elements and that this is unnecessary. If we DO have such elements, I'd like to know if there's a good reason why, otherwise I'd say that's it's own bug and that this is masking the problem.

frontend/src/app/search.tsx
```diff
@@ -183,36 +203,33 @@ export default class Search extends Component {
...
-                  this.setState({deleteButtonVisible: false})
-                }
-              })
-              this.transientPaths = []
-              console.log('transientPaths:'+this.transientPaths)
+          this.setState({countOfCheckedBoxes: 0}, () => {
+            console.log('countOfCheckedBoxes: '+this.state.countOfCheckedBoxes)
+            this.transientPaths = []
+            console.log('transientPaths:'+this.transientPaths)
+            this.setState({deleteButtonVisible: false})
```
benradey: I remember we had talked about this block. Looks good. :)

frontend/src/app/search.tsx
```diff
@@ -25,14 +25,25 @@ export default class Search extends Component {
...
+    const id = 'userID';
+    if (!this.state.loggedinStatus) {
+      fetch("/system/sling/info.sessionInfo.json")
+        .then(response => response.json())
+        .then(responseJSON => {
+          if (responseJSON[id] !== 'anonymous') {
+            this.setState({ loggedinStatus: true })
+          }
+        })
+    }
```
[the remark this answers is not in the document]
ankkit24: Good observation Ben!
As far as I understand the virtual DOM concept in react, only the node/elements that are updated, get re-rendered in the actual DOM. So, if the login status is not changing while navigating through pages, then the whole things wont be rendered I suppose. So, if this is true, then the piece of code should only be rendered in the actual DOM if the state of 'loggedinStatus' changes.

frontend/src/app/search.tsx
```diff
@@ -93,16 +105,24 @@ export default class Search extends Component {
                         ]} 
                   />
                 </DataListItemRow>
+                <DataListItemRow id="data-rows" key={this.state.data["pant:transientPath"]}>
```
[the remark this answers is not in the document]
ankkit24: Sure Ben!
I will put comments for the widgets that do not look clear.
Should be visible in the code for the next PR.

frontend/src/app/search.tsx
ankkit24: Yes, this was covered.
Thanks Ben!

frontend/src/app/search.tsx
ankkit24: The splice is required if we an uncheck the select all box.
But, the below 2 lines of code-
        this.state.allPaths=[]
        this.transientPaths=[]
will clear the transientPaths[] and allPaths[] when we try to use the select all feature for the second time in a row.
If we dont do that, the transientPaths[] and allPaths[], will carry forward the values which have been deleted, and in the next delete operations, these values are not found.
Hence, they need to be cleared.

frontend/src/app/search.tsx
ankkit24: Yes Ben!
This was added as the checkbox for selectall was visible initially even when no rows were fetched.
That condition has been handled now, so this line of code seems unnecessary.
Will remove this.

frontend/src/app/search.tsx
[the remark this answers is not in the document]
benradey: Hmm - you might be on to something - but I couldn't resist testing this myself.

Unfortunately, the problem is worse than I thought... :)

Every time you type a character into the search box, the `render()` method is called, which makes a new `fetch()` with each and every keystroke when the user is not logged in. 

Given this behavior, this has gone from "theoretical annoyance" to "actual misbehavior" and I would like to see it fixed. Sorry to be a pain. :)

image

frontend/src/app/search.tsx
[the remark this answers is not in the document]
benradey: I see them in this PR. Looks good! We should follow this pattern generally. :)

frontend/src/app/search.tsx
ankkit24: This has been taken care of after the discussion we had on this.

frontend/src/app/search.tsx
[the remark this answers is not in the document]
ankkit24: Hey Ben!
I have made the code changes so that the fetch() is not called for each keystroke on any text input box.

PR #72, opened 2019-06-18
Add MIT License file to the main project

jmarrero:


jmarrero:
I am proposing using the MIT license for the whole project. The frontend already has this license as I forked the project from https://github.com/patternfly/patternfly-react-seed

jmarrero:
While I love the GPL and what it stands for, a lot of companies refuse to use GPL software or even contribute to projects with similar licenses. Many companies make teams go thru a lengthy legal process to even use GPL licensed software internally. Ideally I would love for everyone to jump on the opensource all the things bandwaggon but that does not seem to be happening. The reverse seems to be happening with big projects like reddis and mongodb changing their licenses to more proprietary stuff.

An article that talk about it a little bit is: https://opensource.com/article/17/2/decline-gpl

Thinking of the type of tool we are developing I think making it easy to adopt to enterprise teams is something that might help the adoption of this project in the long term.

ivanhorvath:
After reading that article, my vote is still the MIT one. Trying to force others to do what we want them to do with our free bits of code is why GPL is frowned upon in the last few years.

benradey:
I don't see an issue with using GPL. Companies can still take our code and do what they want with it, internally. All the GPL asks is that IF they make their version of the software available, they make it open with the same license. Am I wrong about that?

Practically speaking, I don't know that anybody will actually release their fork of our software, but a man can dream. My point is this doesn't stop companies from adopting it and tweaking it for themselves, so I don't think it should harm adoption.

jmarrero:
While a lot of people in the open source community think like this, many lawyers don't seem too. I am talking from experience working on various places where GPL software was banned unless you had legal clearance and legal clearance was not always granted.

See an example of exactly that:
https://www.r-bloggers.com/how-gpl-makes-me-leave-r-for-python/

Here is someone in the community with those views from wayback in 2002:
https://www.linuxjournal.com/article/5935

Furthermore there are exceptions like the ones used for JAVA to have a more clear statement on what derivative works mean:
https://resources.whitesourcesoftware.com/blog-whitesource/top-9-gpl-with-the-classpath-exception-questions-answered

Many people take this portion of the GPLv2: https://www.gnu.org/licenses/old-licenses/gpl-2.0.en.html

> b) You must cause any work that you distribute or publish, that in whole or in part contains or is derived from the Program or any part thereof, to be licensed as a whole at no charge to all third parties under the terms of this License.

to be too open for interpretation.

benradey:
One of the articles you linked pointed out that people are more likely to contribute their changes back upstream if they don't think they need to avoid being caught not honoring the GPL. That is a really good point.

Does the LGPL suffer the same criticisms as the GPL? If not, we should consider that. The case I'm trying to avoid is someone slapping their label on our project and selling it as their own. If the LGPL avoids that while also avoiding the legal scorn of the GPL, then that's a win in my book.

jmarrero:
I have not studied the LGPL before as I have not had to discuss it in previous projects. But from a little search The LGPL is similar to GPL in the sense that any code you modify must be contributed back.

The difference is that the implication of "all the project" that touches GPL code does not need to be contributed back. Just the modified part.

This is also seems to be seen as something to be cautious about.
https://softwareengineering.stackexchange.com/questions/47323/can-i-use-an-lgpl-licenced-library-in-my-commercial-app

That las link points to this nice thread that raises a lot of questions for me:
http://answers.google.com/answers/threadview/id/439136.html

I would also argue that the wording of the license might not be clear for us as a whole project, we are using Frontend code in react, code in python for the client and java backend. It gives people many ways to interpret the dynamic linking part of the LGPL.

If someone takes our code and wraps around their own code is it violating the LGPL? What if they make an application that calls the client programmatically? What about any combination of components? This is not totally clear to me, but from that thread of random people on the internet it looks like people are on the hook even if you write a wrapper to call the library.

I am not a lawyer, so all my information comes from the internet and past experiences trying to get permission to use stuff GPL licensed. We use a lot of MIT and Apache licensed stuff to build our project so I also lean towards MIT when possible. Random people contributing will probably not care about the license at all. But companies might even avoid download our stuff because of the license.

We might also have an ideological difference. I don't think people copying our stuff and putting their banner on top of it it's a bad thing. In essence that is what distributions are and the best one wins. I don't want people avoiding us for licenses and also don't want to force anyone to contribute if they don't want to. If someone wants to build a proprietary system on top of our stuff, put resources to make something more compelling than our OpenSource version, then, they have bested us and we should have done better.

jmarrero:
Merging based on my conversation with @benradey in the office.

PR #73, opened 2019-06-18
add pf4-layout to login and module & change highlight color on search

jmarrero:
This PR adds the bullseye patternfly layout to the login and module pages. It also changes the text input on these pages for formgroups, this allows an user to click on the label and have the text input box focused automatically.

The blue highlight on the search was changed for a grayish color.

I also disabled the no-console tslint rule, while we are developing the console log is very helpful, when we have alerts boxes and upgrade this project to GA we should re-enable the rule.


jmarrero:
thanks Ben!

PR #77, opened 2019-06-26
CCS-2931: Add Git Import page

jmarrero:


jmarrero:
work in progress not ready for review.

frontend/src/app/gitImport.tsx
```diff
@@ -0,0 +1,159 @@
...
+            body: JSON.stringify(payload),
+            method: "POST"
+          }).then(response => {
+            if (response.status === 201 || response.status === 200) {
+              console.log(" Works " + response.status)
+              this.setState({ isFormSubmitted: true, isSucess: true, msgType: "success", submitMsg: "The git import has been submitted it might take up to 1 minute to see it in the module library" })
+            } else if (response.status === 500) {
+                console.log(" Failed " + response.status)
+                response.text().then((text) => {
+                  this.setState({ submitMsg: text })
```
benradey: This needs a bit of polish. When I run this locally with no go component, I get an error dialog that fills the entire page and starts with `<html> <head> <title>Error while processing /&lt;!DOCTYPE html&gt;&lt;html...`

frontend/src/app/gitImport.tsx
```diff
@@ -0,0 +1,159 @@
...
+    console.log("My repo is: " + this.state.repository + " and branch is " + this.state.branch + ".")
+    if (this.state.repository === "") {
+      this.setState({ isMissingFields: true })
+    } else {
+      const hdrs = {
+        'Accept': 'application/json',
+        'cache-control': 'no-cache'
+      }
+
+      fetch('/content/repositories/GIT_INTEGRATION_URL')
```
benradey: We should decide on a naming convention here, I'm not wild about the CAPS_CASE.

frontend/src/app/gitImport.tsx
```diff
@@ -0,0 +1,159 @@
...
+        .then((resp) => resp.text()).then((text) => {
+          this.setState({ git2pantheonURL: text })
+        }).then(() => {
+          console.log("The git2pantheon URL is: " + this.state.git2pantheonURL)
+          const payload = {
+            branch: this.state.branch,
+            repo: this.state.repository
+          };
+          fetch(this.state.git2pantheonURL + "/clone", {
```
benradey: Although I'm not in love with this double fetch, I think I understand why you have to do it. There's no way for pantheon to dynamically bake the git url into the javascript, so this is the next best option, right?

tools/git2pantheon/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
benradey: Is this necessary? Don't we already have a license in the project root that would cover this?

tools/git2pantheon/README.md
```diff
@@ -0,0 +1,89 @@
+**Setting up your development environment**
```
benradey: THANK YOU for this nice readme!

tools/git2pantheon/git_uploader.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+func cleanup(directory string) {
+	//Todo
+	log.Printf("Cleanup: deleting directory: %s", directory)
+	os.RemoveAll(directory)
+}
+
+func randomAlphaNumericString() string {
+	const chars = "0123456789abcdefghijklmnopqrstuvwxyzREDHAT"
```
benradey: Nice touch. :)

frontend/src/app/gitImport.tsx
jmarrero: mm that is unexpected. Let me see why my dialog box is not showing up.

frontend/src/app/gitImport.tsx
jmarrero: renamed to pantGitServiceURL as we talked about.

frontend/src/app/gitImport.tsx
jmarrero: yeah the other way would be using build time ENV variables. I think we are stuck with this for now.

tools/git2pantheon/LICENSE
jmarrero: Removing it.

tools/git2pantheon/README.md
jmarrero: my pleasure :dancer:

frontend/src/app/gitImport.tsx
jmarrero: Thanks for catching this, the issue came from the configuration not being defined in sling and react was doing a call to localhost:8080/clone which redirects to /pantheon so the whole html payload was in the response. This should be fixed now.

frontend/src/app/gitImport.tsx
```diff
@@ -97,35 +98,40 @@ class GitImport extends Component {
...
-          };
-          fetch(this.state.git2pantheonURL + "/clone", {
-            body: JSON.stringify(payload),
-            method: "POST"
-          }).then(response => {
-            if (response.status === 201 || response.status === 200) {
-              console.log(" Works " + response.status)
-              this.setState({ isFormSubmitted: true, isSucess: true, msgType: "success", submitMsg: "The git import has been submitted it might take up to 1 minute to see it in the module library" })
-            } else if (response.status === 500) {
+          if (this.state.git2pantheonURL.includes("DOCTYPE html")) {
```
benradey: Hmm... I'm not sure how I feel about this new technique. It assumes that any error message will be html, and anything else is a success message. I don't think that is a reliable assumption, and I think it may only be true because we wrote the redirect handler (i.e. assume anything that doesn't resolve is a router request like /login or /search). Was relying on the status code not a reliable technique?

jmarrero: So if I curl with -v to see the exchange I see that our servlet is always giving us 200 even if the resource does not exist. 

```
curl -v localhost:8080/content/repository/random
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 8080 (#0)
> GET /content/repository/random HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.64.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Date: Mon, 01 Jul 2019 19:44:41 GMT
< X-Content-Type-Options: nosniff
< X-Frame-Options: SAMEORIGIN
< Content-Type: text/html;charset=utf-8
< Transfer-Encoding: chunked
< 
<!DOCTYPE html>
<html lang="en-US">

<head>
  <meta charset="utf-8">
  <title>Pantheon</title>
  <meta id="appName" name="application-name" content="Pantheon">
  <meta name="viewport" content="width=device-width, initial-scale=1.0">
  <link rel="icon" type="image/png" href="https://www.patternfly.org/components/patternfly/dist/img/favicon.ico">
<link href="app.css" rel="stylesheet"></head>

<body>
  <noscript>Enabling JavaScript is required to run this app.</noscript>
  <div id="root"></div>
<script type="text/javascript" src="app.bundle.js"></script></body>

</html>

```

I can change it to do an additional check, if it's not 200 then throw the same Alert. However, for the issue you saw I think checking for HTML is the way to catch it because of what we are doing in the backend.

jmarrero: @benradey added some more error handling.

PR #79, opened 2019-06-28
Uploader

ankkit24:
Changes to the uploader python script to allow it to upload files which are passed as arguments with the script.


jmarrero:
@ankkit24 I understand the while true there and trying to find the current dir. I wonder if that is the safest way of doing that in python. For example I would be happier with some something like while (currentdir != parentdir) instead of declaring an infinite loop and breaking inside the infinite loop.

I will deffer on @xdavidson approval here as she has worked on the uploader before. I have setup a meeting to go over this and the admin panel reviews with Ivan and Lisa.

jmarrero:
This looks good, but we would be introducing a change that has the potential of adding more requirements on the team in the long run. Since this has not been asked for we are just closing this PR and will not implement this at the moment.

PR #81, opened 2019-07-02
Asciidoctor service

carlosmunoz:
extract the html generation logic to an independent service
refactor the code for conciseness and readability
add a test for the new AsciidoctorService


carlosmunoz:
@jmarrero with Ben out I thought I'd move this to you. Let me know what you think.

PR #83, opened 2019-07-05
Check if user is logged on module

jmarrero:
This is a temporary fix. Another similar to this will come after the git stuff is approved and merged for that page. While this adds another call to the backend, It adds security.
I am working in another PR to extract all this login checks into it's own component, but that will take a little more time.


jmarrero:
adding this and the gitImport page check too to the git import PR instead...

PR #84, opened 2019-07-09
Update handlebars dependency

jmarrero:


jmarrero:
Updating jest which depends on handlebars. Handlebars 4.1.0 has a vulnerability that is fixed in 4.1.2 as seen in https://github.com/wycats/handlebars.js/blob/master/release-notes.md

PR #85, opened 2019-07-09
Admin

ankkit24:
Admin Panel created with the below details in it-
* Browser link
* Welcome link
* Console link
* Current Build date
Also, the admin panel is accessible only if the user is logged in as an admin.


jmarrero:
@ankkit24 my comment is still valid please make sure that any code that does not need to change for this PR is not changed.

PR #89, opened 2019-07-12
[CCS-2999] add support for subdir/*

xdavidson:
- add docstrings to methods
- Refactor code
- Per discussion with Carlos, it's better to use glob's out of box behaviour when uploading everything under a subdirecory. User should use "subdir/**/*" instead of "subdir/*"
https://docs.python.org/3/library/glob.html

### DIFF
--- a/uploader/pantheon.py
+++ b/uploader/pantheon.py
@@ -9,6 +9,10 @@ import socket
 import base64
 import sys
 import requests
+import pathlib
+import fnmatch
+import re
+import glob
 from pathlib import PurePath
 
 DEFAULT_SERVER = 'http://localhost:8080'
@@ -134,6 +138,10 @@ if pw == '-':
     pw = getpass.getpass()
 
 config = None
+
+if not os.path.exists(args.directory):
+    raise ValueError("Directory not found {}".format(args.directory))
+
 try:
     config = yaml.safe_load(open(args.directory + '/' + CONFIG_FILE))
 except FileNotFoundError:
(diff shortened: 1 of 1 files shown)


uploader/pantheon.py
```diff
@@ -161,6 +157,146 @@ def remove_trailing_slash(path):
         path = path[:-1]
     return path
 
+def recursive_glob(directory, pattern='*'):
```
carlosmunoz: Hi Lisa, my suggestion with methods like these is to add some documentation strings (or whatever they are called in python) so that it's easier to follow the review.

uploader/pantheon.py
```diff
@@ -161,6 +157,146 @@ def remove_trailing_slash(path):
...
+        hiddenFolder = False
+        while not ppath == PurePath(args.directory):
+            logger.debug('ppath: %s', str(ppath.stem))
+            if ppath.stem[0] == '.':
+                hiddenFolder = True
+                break
+            ppath = ppath.parent
+        if hiddenFolder:
+            logger.info('Skipping %s because it is hidden.', str(path))
+            logger.info('')
```
carlosmunoz: I would make this a debug level statement.... for a quieter console.

uploader/pantheon.py
xdavidson: Carlos,

I updated the new methods with docstrings.

uploader/pantheon.py
xdavidson: updated.

jmarrero:
This has been reviewed by @ivanhorvath and me. This looks good to us.

anarang:
@xdavidson I was just referring to [https://docs.python.org/3/library/glob.html)] and found that the string './**/' was acceptable and would iterate through everything in the current directory, but I tried the same pattern in our config file pantheon2.yml and it doesn't seem to be working. Is there an exception to that? I think we should have a way where we can just say iterate recursively through everything in the current directory whereas currently we have to specify all the modules in the base directory.

anarang:
@xdavidson I see that '- ./**' without the trailing slash is working fine and including all the subdirectories and modules present.So, this is great!  Do we want to add some code to handle the trailing slash and work in the same way with or without slash?

carlosmunoz:
I would stick with whatever the glob library supports, and I am assuming that it supports the unix glob patterns). This way users would not be surprised by custom behaviour with the specified patterns.

carlosmunoz:
...also, code coverage is down :frowning_face:

jmarrero:
@carlosmunoz the travisci script is not explicitly uploading the report for python. It must be just reading each file's line count and getting confused somehow. Even in the report we are not really looking at and even the diff here:
https://codecov.io/gh/redhataccess/pantheon/pull/89/diff
marks the changed file as untracked.

xdavidson:
@anarang, 

Regarding your experiment with pattern './**/, here is the copy and paste of the output from python.org
https://docs.python.org/3/library/glob.html
>>> glob.glob('./**/', recursive=True)
['./', './sub/']

Notice the ouput above. It returns the matched directories. Uploader only processes patterns for files matched. It ignores patterns for directories matched.

Lisa

anarang:
@xdavidson fair enough, that makes sense. thanks for the clarification there :) So, we will not have anything uploaded if there is a trailing slash in the above pattern since that will only list the subdirectories? right?

xdavidson:
@anarang correct

PR #92, opened 2019-07-19
WIP: test uploader macos packaging

jmarrero:
Trying to build the uploader as a mac application. The build will happen on travis.


jmarrero:
It seems like the mac users are just a couple, so it makes no sense maintaining this in the long term for now.

PR #94, opened 2019-07-19
Ccs 2961 products

xdavidson:
- Create Product and ProductVersion nodes
- Create Product and ProductVersion model classes with basic/minimal fields
- Create UnitTests 

Testing notes:
To create a product called "Openshift Dedicated"
$  curl -u admin -F"name=Openshift Dedicated" -F"jcr:primaryType=pant:product"  -F"sling:resourceType=pantheon/product" -F"description=A laaS cloud Management Platform" http://localhost:8080/content/products/openshift_dedicated

To create a productVersion 3 for the above product
$  curl -u admin -F"name=3" -F"sling:resourceType=pantheon/productVersion" -F"jcr:primaryType=pant:productVersion" http://localhost:8080/content/products/openshift_dedicated/versions/3


src/main/java/com/redhat/pantheon/model/Product.java
```diff
@@ -0,0 +1,31 @@
...
+import javax.annotation.Nonnull;
+
+import org.apache.sling.api.resource.Resource;
+import org.apache.sling.models.annotations.Default;
+
+import com.redhat.pantheon.model.api.SlingResource;
+
+public class Product extends SlingResource {
+
+    @Default(values = "pantheon/product")
```
carlosmunoz: Just FYI: This annotation does nothing. It's just a reminder of what the default type should be.

src/main/resources/SLING-INF/content/products.json
```diff
@@ -0,0 +1,4 @@
+{
+  "jcr:primaryType": "sling:Folder",
+  "sling:resourceType": "pantheon/product/all"
```
carlosmunoz: I wouldn't bother with the resource type here. I don't think we're going to use it for these nodes.

src/main/java/com/redhat/pantheon/model/Product.java
```diff
@@ -0,0 +1,31 @@
...
+
+    public final Field<Calendar> created = new Field<>(Calendar.class, "jcr:created");
+
+    public final Field<String> createdBy = new Field<>(String.class, "jcr:createdBy");
+
+    public final Field<String> primaryType = new Field<>(String.class, "jcr:primaryType");
+
+    public final Field<String> name = new Field<>(String.class, "name");
+    
+    public final Field<String> productDescription = new Field<>(String.class, "productDescription");
```
carlosmunoz: I would go with just `description` here. There's a mixin called `jmix:description` which gives us a free `jcr:description` field, but not sure if we would want to use it.

src/main/java/com/redhat/pantheon/model/Product.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.Calendar;
+
+import javax.annotation.Nonnull;
+
+import org.apache.sling.api.resource.Resource;
+import org.apache.sling.models.annotations.Default;
+
+import com.redhat.pantheon.model.api.SlingResource;
+
+public class Product extends SlingResource {
```
carlosmunoz: Can we have some class javadocs please? :smile:

src/main/java/com/redhat/pantheon/model/Product.java
xdavidson: ack

src/main/resources/SLING-INF/content/products.json
xdavidson: line 3 is removed.

src/main/java/com/redhat/pantheon/model/Product.java
xdavidson: i renamed it to 'description'.

src/main/java/com/redhat/pantheon/model/Product.java
xdavidson: done.

PR #95, opened 2019-07-22
Admin page UI changes

ankkit24:
Added links for the below in the admin panel-
* Browser url
* Web console url
* Sling welcome page

- Also, the latest build date is displayed below the links


frontend/src/app/admin.tsx
```diff
@@ -1,6 +1,11 @@
 import React, { Component } from 'react';
-import { Button } from '@patternfly/react-core';
 import { Redirect } from 'react-router-dom';
+import { Button } from '@patternfly/react-core';
+import { Grid, GridItem } from '@patternfly/react-core';
+import browseri from './sample.svg';
```
jmarrero: You are importing these svgs (console1.svg, etc) but they are not being used, instead it looks like the SVG is copied on the admin.tsx as well. I would delete the paths with the svg vectors from this file and use the imported ones.

ankkit24: Hey Joseph!
I have addressed the issue.
The images are appearing now.

jmarrero:
@ankkit24 The code looks good now. However why jpg? and not svg like before? now when I make my window smaller it distorts the image.

jmarrero:
It seems like using svg has a similar problem. I think this is fine for now.

PR #97, opened 2019-07-24
create a single abstract class to encapsultate commonly used query logic

carlosmunoz:
... and in the process remove a lot of duplicate code.

The new `AbstractJsonQueryServlet` should serve as the super class for any servlet which needs to do a query on the server. It offers  out-of-the-box pagination, custom queries and custom results.

This PR also refactors two of the existing query servlets to use this new base class, thus eliminating a bunch of similar code, and hopefully simplifying the system going forward.


frontend/src/app/search.tsx
```diff
@@ -458,22 +458,20 @@ export default class Search extends Component {
 
   private buildSearchUrl() {
     let backend = "/modules.json?search="
-    if (this.state.input != null && this.state.input !== "") {
-      backend += this.state.input
```
jmarrero: This is needed to pass the parameter down to the search. Currently we get:
```
/modules.json?search=&key=&direction=desc&offset=0&limit=10
```
that should include the input after search= for example
search="hello"

carlosmunoz: very good catch :+1:

carlosmunoz: I found another problem as a result of fixing this.

PR #98, opened 2019-07-25
add a metadata extractor

carlosmunoz:
The metadata extractor is an asciidoctor extension which is capable of extracting metadata from the Asciidoctor AST. In this revision only title and abstract are extracted, and with limitations. The metadata extractor is called after each module revision creation.

This change also fixes a memory leak in the Asciidoctor pool.

### DIFF
--- a/src/main/java/com/redhat/pantheon/asciidoctor/extension/MetadataExtractorTreeProcessor.java
+++ b/src/main/java/com/redhat/pantheon/asciidoctor/extension/MetadataExtractorTreeProcessor.java
@@ -1,0 +1,62 @@
+package com.redhat.pantheon.asciidoctor.extension;
+
+import com.redhat.pantheon.model.ModuleRevision;
+import org.asciidoctor.ast.Document;
+import org.asciidoctor.ast.StructuralNode;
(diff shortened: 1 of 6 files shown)


src/main/java/com/redhat/pantheon/asciidoctor/extension/MetadataExtractorTreeProcessor.java
```diff
@@ -0,0 +1,62 @@
...
+    @Override
+    public Document process(Document document) {
+        extractDocTitle(document);
+        extractAbstract(document);
+        return document;
+    }
+
+    private void extractDocTitle(Document document) {
+        String docTitle = document.getDoctitle();
+        if(!isNullOrEmpty(docTitle)) {
```
jmarrero: This is pretty neat for some reason I imagined that this was going to be harder, like writing our own parser.

carlosmunoz: Yeah, just using asciidoctor's parsing capabilities. There are some nuances there still, but this is a good start.

PR #100, opened 2019-08-02
Add Dockerfile and yaml files to build on openshift.

jmarrero:


container/Dockerfile
```diff
@@ -0,0 +1,37 @@
+#Build stage
+FROM fedora:30 AS builder
```
anarang: @jmarrero are we sure that we want to use Fedora 30 base image as Fedora has more frequent release cycles than CentOS or RHEL. In the past, I have seen mostly CentOS being used for containers. I am not so sure about it, but just wanted to bring it up since it came to my mind while going through the code.

carlosmunoz: This is just for the builder image as I understand it, not the actual and final pantheon image.

jmarrero: Yeah, the final image is generated with this base:
```
FROM registry.access.redhat.com/ubi8/ubi
```
which is the RHEL 8 based Universal Basic Image which is described here:
https://www.redhat.com/en/blog/introducing-red-hat-universal-base-image

jmarrero: @anarang also this explains the two stages and what I did there with the two diferent bases:

PR #101, opened 2019-08-09
Refactor the revisioning system

carlosmunoz:
Refactor the model to only hold one draft and one released revision of a module at any given time.

PR #102, opened 2019-08-13
Update preview call in react for drafts

jmarrero:
Depends on @carlosmunoz #101


jmarrero:
@carlosmunoz  you can also just add this like to your PR and merge it there.

carlosmunoz:
@jmarrero the CI build for this PR failed.
I have merged the dependent PR already so this should be good to go. One question: this will still result in 404s for the released revision previews (if there isn't one). Perhaps we should add two links to the results' table: one for draft (if there is one) and one for released (again, if there is one)

jmarrero:
@carlosmunoz since there is no way to mark anything as released via the UI this PR is just trying to maintain status quo until the UI that manages revisions is ready.

PR #106, opened 2019-08-21
Add build date and commit info to header

jmarrero:


frontend/src/app/components/Chrome/Header/BuildInfo.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+    private getBuildInfo() {
+        const backend = "/pantheon/builddate.json"
+        if (this.state.buildDate === '') {
+            fetch(backend)
+                .then(response => response.json())
+                .then(responseJSON => this.setState({ buildDate: "Built Date: " + responseJSON.buildDate, commitHash: responseJSON.commitHash }))
+                .then(() => {
+                    if (!(this.state.commitHash.includes("not set"))) {
+                        this.setState({ commitText: 'commit hash' })
```
benradey: Is this intentional? You are literally showing the words "commit hash" as the link text to the user?

frontend/src/app/components/Chrome/Header/BuildInfo.tsx
```diff
@@ -0,0 +1,44 @@
...
+            </React.Fragment>
+        );
+    }
+
+    private getBuildInfo() {
+        const backend = "/pantheon/builddate.json"
+        if (this.state.buildDate === '') {
+            fetch(backend)
+                .then(response => response.json())
+                .then(responseJSON => this.setState({ buildDate: "Built Date: " + responseJSON.buildDate, commitHash: responseJSON.commitHash }))
```
benradey: Possible typo? "Built" vs "Build"?

src/main/java/com/redhat/pantheon/sling/PantheonRepositoryInitializer.java
```diff
@@ -36,13 +38,36 @@
...
+
+    private JackrabbitSession getSession(SlingRepository slingRepository) throws RepositoryException {
+        return (JackrabbitSession) slingRepository.loginAdministrative(null);
+    }
+
+    private void setGitServiceURL(JackrabbitSession s) throws RepositoryException {
+        if (System.getenv("GIT_SERVICE_URL") != null) {
+            assignPermissionToPrincipal(s,"pantheon","/conf/pantheon", null, Privilege.JCR_READ, Privilege.JCR_WRITE);
+            assignPermissionToPrincipal(s,"pantheon-users","/conf/pantheon", null, Privilege.JCR_READ);
+            s.getNode("/conf/pantheon").setProperty("pant:GitServiceURL",System.getenv("GIT_SERVICE_URL"));
```
benradey: Is this `GIT_SERVICE_URL` property just in case you're hosting your code somewhere other than github, so you can link to it correctly?

Edit: No, that must not be it, since github.com is hardcoded in the link redirect. So what is the purpose of this code?

src/test/java/com/redhat/pantheon/use/PlatformDataTest.java
```diff
@@ -0,0 +1,36 @@
...
+    @Test
+    @DisplayName("Jar buildDate must not be set at build time")
+    public void testJarBuildDate() throws Exception {
+        //Given
+        PlatformData data = new PlatformData();
+
+        //When
+
+        //Then
+        assertEquals(true,data.getJarBuildDate().contains("Unable to determine build date") );
```
benradey: Why did you write this test, and does it provide value? All it's really verifying is that the code is failing in a scenario where it can't possibly succeed anyway, so... I'm not sure I see what this achieves.

src/main/java/com/redhat/pantheon/sling/PantheonRepositoryInitializer.java
jmarrero: This is the git2pantheon URL.

frontend/src/app/components/Chrome/Header/BuildInfo.tsx
jmarrero: yeah typo

frontend/src/app/components/Chrome/Header/BuildInfo.tsx
jmarrero: That will never be shown, it's left behind from previous commit. I will remove that now.

src/test/java/com/redhat/pantheon/use/PlatformDataTest.java
jmarrero: It is testing specifically that Class without going thru the servlet. If we are not writing test for each class then I will remove it.

PR #108, opened 2019-08-23
(CCS-3074) Generate on upload

benradey:
A few changes as a result of not versioning assets:

- Modules now generate immediately on upload
- Released modules can never be regenerated
- Assets must now be uploaded prior to modules
- Also changed module version naming scheme from v0, v1, v2... to 1, 2, 3... under the rationale that the "v" is unnecessary and that users will expect numbering to start at 1

### DIFF
--- a/src/test/java/com/redhat/pantheon/asciidoctor/AsciidoctorServiceTest.java
+++ b/src/test/java/com/redhat/pantheon/asciidoctor/AsciidoctorServiceTest.java
@@ -1,7 +1,9 @@
 package com.redhat.pantheon.asciidoctor;
 
 import com.redhat.pantheon.conf.GlobalConfig;
+import com.redhat.pantheon.model.api.FileResource;
 import com.redhat.pantheon.model.module.Content;
+import com.redhat.pantheon.model.module.Metadata;
 import com.redhat.pantheon.model.module.Module;
 import com.redhat.pantheon.model.module.ModuleRevision;
 import com.redhat.pantheon.sling.ServiceResourceResolverProvider;
@@ -9,12 +11,14 @@ import org.apache.sling.api.resource.Resource;
 import org.apache.sling.testing.mock.sling.junit5.SlingContext;
 import org.apache.sling.testing.mock.sling.junit5.SlingContextExtension;
 import org.asciidoctor.Asciidoctor;
+import org.asciidoctor.extension.JavaExtensionRegistry;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
 
 import java.io.IOException;
+import java.util.Map;
 import java.util.Optional;
 
 import static com.google.common.collect.Maps.newHashMap;
@@ -22,6 +26,9 @@ import static com.redhat.pantheon.util.TestUtils.registerMockAdapter;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.lenient;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
 
 @ExtendWith({SlingContextExtension.class, MockitoExtension.class})
 class AsciidoctorServiceTest {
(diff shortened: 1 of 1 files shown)


src/main/java/com/redhat/pantheon/asciidoctor/AsciidoctorPool.java
```diff
@@ -44,9 +44,14 @@ public AsciidoctorPool(@Reference GlobalConfig globalConfig) {
      */
     public Asciidoctor borrowObject(Resource base) {
         Asciidoctor asciidoctor = super.borrowObject();
-        asciidoctor.javaExtensionRegistry().includeProcessor(
-                new SlingResourceIncludeProcessor(base));
-        return asciidoctor;
+        try {
+            asciidoctor.javaExtensionRegistry().includeProcessor(
+                    new SlingResourceIncludeProcessor(base));
```
carlosmunoz: I've always thought that we should change this. The responsibility of declaring the necessary extensions should be left up to the caller of the pool, and the pool simply clears them up upon return of the borrowed object.

src/main/java/com/redhat/pantheon/asciidoctor/AsciidoctorPool.java
```diff
@@ -44,9 +44,14 @@ public AsciidoctorPool(@Reference GlobalConfig globalConfig) {
...
+            return asciidoctor;
+        } catch (Exception e) {
+            returnObject(asciidoctor);
```
carlosmunoz: Hhmm... this is an interesting scenario. It actually reaffirms my position in my previous comment. The only reason for this overloaded method is so we can introduce the proper asciidoctor extensions. If we leave that responsibility to the caller, this method becomes unnecessary.

src/main/java/com/redhat/pantheon/asciidoctor/AsciidoctorService.java
```diff
@@ -100,6 +103,24 @@ public String getModuleHtml(@Nonnull ModuleRevision moduleRevision,
         return html;
     }
 
+    /**
+     * Builds a context Map that is initially populated from request parameters which are prefixed with "ctx_".
+     * @param request The http request provided by Sling
+     * @return A Map object with all context parameters as keypairs, minus the "ctx_" prefix
+     */
+    public static Map<String, Object> buildContextFromRequest(SlingHttpServletRequest request) {
```
carlosmunoz: :+1: on adding this method. Nice encapsulation.

src/main/java/com/redhat/pantheon/model/module/Module.java
```diff
@@ -139,7 +139,7 @@ public ModuleRevision createNextRevision() {
         }
 
         private String generateNextRevisionName() {
-            return "v" + stream(this.getChildren()).collect(counting());
+            return "" + (stream(this.getChildren()).collect(counting()) + 1);
```
carlosmunoz: I see you've changed the revision naming strategy. No complaints from me. There might be a few header comments (in the Module class I think) where we might want to update the examples.

src/main/java/com/redhat/pantheon/servlet/AsciidocRenderingServlet.java
```diff
@@ -66,7 +67,7 @@ public AsciidocRenderingServlet(
     protected void doGet(SlingHttpServletRequest request,
             SlingHttpServletResponse response) throws ServletException, IOException {
         String locale = paramValue(request, "locale", DEFAULT_MODULE_LOCALE.toString());
-        boolean draft = paramValueAsBoolean(request, "draft");
+        boolean draft = paramValueAsBoolean(request, PARAM_DRAFT);
```
carlosmunoz: :+1: I guess while we are at this, line 69 could also use a static variable extraction.

src/main/java/com/redhat/pantheon/servlet/AsciidocRenderingServlet.java
```diff
@@ -87,16 +88,11 @@ protected void doGet(SlingHttpServletRequest request,
...
-            .collect(toMap(
-                    reqParam -> reqParam.getName().replaceFirst("ctx_", ""),
-                    reqParam -> reqParam.getString())
-            );
+            Map<String, Object> context = asciidoctorService.buildContextFromRequest(request);
 
+            // only allow forced rerendering if this is a draft revision. Released and historical revs are written in stone.
             String html = asciidoctorService.getModuleHtml(
-                    moduleRevision.get(), module, context, paramValueAsBoolean(request, PARAM_RERENDER));
+                    moduleRevision.get(), module, context, draft && paramValueAsBoolean(request, PARAM_RERENDER));
```
carlosmunoz: :+1: good call

src/main/java/com/redhat/pantheon/servlet/ModuleRevisionUpload.java
```diff
@@ -54,11 +56,15 @@
...
 
+    private AsciidoctorService asciidoctorService;
+
     private AsciidoctorPool asciidoctorPool;
 
     @Activate
-    public ModuleRevisionUpload(@Reference AsciidoctorPool asciidoctorPool) {
+    public ModuleRevisionUpload(@Reference AsciidoctorPool asciidoctorPool,
+                                @Reference AsciidoctorService asciidoctorService) {
         this.asciidoctorPool = asciidoctorPool;
```
carlosmunoz: I wonder if we still need the pool object here. Since we are calling the `asciidoctorService` directly now, I don't think we need to get asciidoctor instances directly anymore.

src/main/java/com/redhat/pantheon/servlet/ModuleRevisionUpload.java
```diff
@@ -104,6 +110,12 @@ protected void doRun(SlingHttpServletRequest request, PostResponse response, Lis
                     .content.getOrCreate()
                     .asciidoc.getOrCreate()
                     .jcrContent.getOrCreate();
+            boolean generateHtml = false;
+            String jcrData = jcrContent.jcrData.get();
+
+            if ((jcrData != null && !jcrData.equals(asciidocContent)) || draftRevision.get().content.get().cachedHtml.get() == null) {
```
carlosmunoz: There's a danger here that any of those `.get()` calls returns a null, in which case the whole chain of calls should be considered as a null. This is an upgrade to the framework I have on a branch. Might be a good time to submit it.

src/main/java/com/redhat/pantheon/servlet/ModuleRevisionUpload.java
```diff
@@ -116,6 +128,14 @@ protected void doRun(SlingHttpServletRequest request, PostResponse response, Lis
 
             request.getResourceResolver().commit();
 
+            if (generateHtml) {
+                if (moduleResource == null) {
```
carlosmunoz: Is this conditional necessary? Why not just use the `module` variable everywhere.

src/main/java/com/redhat/pantheon/servlet/ModuleRevisionUpload.java
carlosmunoz: The changes I mentioned above have already been integrated. The null-safe way to execute these set of calls and to validate if there is cached html already is as follows:

```java
draftRevision.map(i -> i.content.get()).map(i -> i.cachedHtml.get()).isPresent()
```
(I know it's not the most readable way, but blame java for its verboseness)
This would guarantee that at any point in the call if there is a missing child in the tree, this would return false, without any NPEs thrown.

src/main/java/com/redhat/pantheon/asciidoctor/AsciidoctorPool.java
benradey: Ok, that's probably a good idea. I suspect that 95% of the time, the intention will be to build a Resource, but I suppose at some point in the future we might be interested in building some in-memory-only asciidoc. We could refactor this then. The immediate concern was that I wrote some code that caused the constructor of the SlingResourceIncludeProcessor to throw an exception, which had the ultimate effect of the asciidcotor instance never being returned to the pool - so after failing to render 10 docs, the system was totally unable to render any further docs. This code update fixes that issue.

carlosmunoz: Yes, I saw that, which made be go down the thought process... but if we remove line 48 and leave it up to the caller to add their own asciidoctor extensions, we remove the chance for there to be an exception and hence the need for this method in the first place.

PR #115, opened 2019-08-30
[CCS-3004] Adds a new API for module rendering

anknite:
Adds a new API /module/render - GET operation to render a Module data in JSON format
(The code is still in-progress)

ToDo - Add test coverage for the API

P.S. \o/ my 1st Pantheon2 PR - feel free to rip it apart :dagger:

### DIFF
--- a/src/main/java/com/redhat/pantheon/model/QuerySingleResult.java
+++ b/src/main/java/com/redhat/pantheon/model/QuerySingleResult.java
@@ -1,0 +1,20 @@
+package com.redhat.pantheon.model;
+
+import java.util.Map;
+
+/**
+ * A simple model to represent a single element result when running queries.
+ */
+public class QuerySingleResult {
+
+    private final Map<String, Object> result;
+
+    public QuerySingleResult(Map<String, Object> result) {
+        this.result = result;
+    }
+
+    public Map<String, Object> getResult() {
+        return result;
+    }
+
+}
--- a/src/main/java/com/redhat/pantheon/servlet/AbstractJsonSingleQueryServlet.java
+++ b/src/main/java/com/redhat/pantheon/servlet/AbstractJsonSingleQueryServlet.java
@@ -1,0 +1,90 @@
+package com.redhat.pantheon.servlet;
+
+import com.redhat.pantheon.jcr.JcrQueryHelper;
+import org.apache.sling.api.SlingHttpServletRequest;
+import org.apache.sling.api.SlingHttpServletResponse;
+import org.apache.sling.api.resource.Resource;
+import org.apache.sling.api.servlets.SlingSafeMethodsServlet;
+import org.jetbrains.annotations.NotNull;
+
+import javax.annotation.Nonnull;
+import javax.jcr.RepositoryException;
+import javax.servlet.ServletException;
+import java.io.IOException;
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Stream;
+
+import static com.google.common.collect.Maps.newHashMap;
+import static com.redhat.pantheon.servlet.ServletUtils.writeAsJson;
+import static javax.servlet.http.HttpServletResponse.SC_NOT_FOUND;
+
+
+/**
+ * Servlet that helps create simple Json result from custom queries.
+ * The servlet returns the first object from the query response irrespective
+ * of the number of elements returned in the response.
+ * This servlet should be extended in the case where you need only the 1st/single element
+ * from the result in json format. See methods to be overriden for details on how to
+ * customize sub classes.
+ *
+ * @author Ankit Gadgil
+ */
+public abstract class AbstractJsonSingleQueryServlet extends SlingSafeMethodsServlet {
+
+    /**
+     * Returns the query to execute. The query may be modified depending on the provided
+     * parameters in the request.
+     * @param request The sling servlet request
+     * @return A string with the query to execute. The results of this query will be used
+     * to render the list of results.
+     */
+    protected abstract String getQuery(SlingHttpServletRequest request);
+
+    /**
+     * Provides a way to modify the returned objects based on the found resources.
+     * The default implementation just returns the corresponding value map.
+     * @param resource The Resource obtained as a result of the query.
+     * @return A map with the actual value to be returned to the servlet's caller.
+     */
+    protected Map<String, Object> resourceToMap(@NotNull Resource resource) {
+        return newHashMap(resource.getValueMap());
+    }
+
+    /**
+     * A post-processor that provides a way to add filter logic for returned objects
+     * The default implementation returns true and can be overriden to check for conditions
+     * @param resource The Resource obtained as a result of the query.
+     * @return Boolean true.
+    */
+    protected boolean isValidResource(@Nonnull Resource resource) {
+        return true;
+    }
+
+    @Override
+    protected final void doGet(@Nonnull SlingHttpServletRequest request, @Nonnull SlingHttpServletResponse response)
+            throws ServletException, IOException {
+
+        JcrQueryHelper queryHelper = new JcrQueryHelper(request.getResourceResolver());
+        try {
+            Stream<Resource> resultStream = queryHelper.query(getQuery(request));
+
+            Optional<Resource> firstResource = resultStream.findFirst();
+            if(firstResource.isPresent()) {
+                if(isValidResource((firstResource.get()))) {
+                    // Convert the resource to JSON
+                    writeAsJson(response, resourceToMap(firstResource.get()));
+                }
+                else {
+                    response.sendError(SC_NOT_FOUND, " Requested resource was invalid.");
+                }
+            } else {
+                response.sendError(SC_NOT_FOUND, " Requested content not found.");
+            }
+
+        } catch (RepositoryException e) {
+            throw new ServletException(e);
+        }
+    }
+
+}
--- a/src/test/java/com/redhat/pantheon/servlet/ModuleJsonServletTest.java
+++ b/src/test/java/com/redhat/pantheon/servlet/ModuleJsonServletTest.java
@@ -1,0 +1,114 @@
+package com.redhat.pantheon.servlet;
+
+import com.redhat.pantheon.model.module.Module;
+import org.apache.sling.api.resource.ModifiableValueMap;
+import org.apache.sling.testing.mock.sling.ResourceResolverType;
+import org.apache.sling.testing.mock.sling.junit5.SlingContext;
+import org.apache.sling.testing.mock.sling.junit5.SlingContextExtension;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+import javax.jcr.query.Query;
+import java.util.HashMap;
+import java.util.Map;
+
+import static com.google.common.collect.Maps.newHashMap;
+import static com.redhat.pantheon.conf.GlobalConfig.DEFAULT_MODULE_LOCALE;
+import static com.redhat.pantheon.util.TestUtils.registerMockAdapter;
+import static javax.servlet.http.HttpServletResponse.SC_OK;
+import static org.junit.jupiter.api.Assertions.*;
+
+@ExtendWith({SlingContextExtension.class})
+class ModuleJsonServletTest {
+
+    SlingContext slingContext = new SlingContext(ResourceResolverType.JCR_OAK);
+    String testHTML = "<!DOCTYPE html> <html lang=\"en\"> <head><title>test title</title></head> <body " +
+            "class=\"article\"><h1>test title</h1> </header> </body> </html>";
+
+    @Test
+    void getQueryNoParams() {
+        // Given
+        ModuleJsonServlet servlet = new ModuleJsonServlet();
+
+        // When
+        String query = servlet.getQuery(slingContext.request());
+
+        // Then
+        // make sure queries don't throw exceptions when executed against the JCR repository
+        assertDoesNotThrow(() -> slingContext.resourceResolver().queryResources(query, Query.JCR_SQL2));
+    }
+
+    @Test
+    void getQuery() {
+        // Given
+        ModuleJsonServlet servlet = new ModuleJsonServlet();
+        Map<String, Object> map = newHashMap();
+        map.put("locale", DEFAULT_MODULE_LOCALE.toString());
+        map.put("module_id", "jcr:uuid");
+        slingContext.request().setParameterMap(map);
+
+        // When
+        String query = servlet.getQuery(slingContext.request());
+
+        // Then
+        // make sure queries don't throw exceptions when executed against the JCR repository
+        assertDoesNotThrow(() -> slingContext.resourceResolver().queryResources(query, Query.JCR_SQL2));
+
+        // make sure query response is not null
+        assertNotNull(slingContext.response());
+    }
+
+
+    @Test
+    void resourceToMap() {
+        // Given
+        slingContext.create()
+                .resource("/content/repositories/repo/module/en_US/1",
+                        "jcr:primaryType", "pant:moduleRevision");
+        slingContext.create()
+                .resource("/content/repositories/repo/module/en_US/1/metadata",
+                        "jcr:title", "A title",
+                        "jcr:description", "A description");
+        slingContext.create()
+                .resource("/content/repositories/repo/module/en_US/1/content/asciidoc",
+                        "jcr:primaryType", "nt:file");
+        slingContext.create()
+                .resource("/content/repositories/repo/module/en_US/1/content/cachedHtml",
+                        "jcr:data", testHTML,
+                        "pant:hash", "2a0e2c43");
+        slingContext.resourceResolver().getResource("/content/repositories/repo/module/en_US").adaptTo(ModifiableValueMap.class)
+                .put("released", slingContext.resourceResolver().getResource("/content/repositories/repo/module/en_US/1").getValueMap()
+                        .get("jcr:uuid"));
+
+        registerMockAdapter(Module.class, slingContext);
+        ModuleJsonServlet servlet = new ModuleJsonServlet();
+        slingContext.request().setResource( slingContext.resourceResolver().getResource("/module") );
+
+        // When
+        Map<String, Object> map = servlet.resourceToMap(
+                slingContext.resourceResolver().getResource("/content/repositories/repo/module"));
+        Map<String, Object> moduleMap = (Map<String, Object>)map.get("module");
+
+        // Then
+        assertTrue(map.containsKey("status"));
+        assertTrue(map.containsKey("message"));
+        assertTrue(map.containsKey("module"));
+        assertTrue(moduleMap.containsKey("module_uuid"));
+        assertTrue(moduleMap.containsKey("product_version"));
+        assertTrue(moduleMap.containsKey("description"));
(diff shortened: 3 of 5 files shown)


src/main/java/com/redhat/pantheon/servlet/AbstractJsonSingleQueryServlet.java
```diff
@@ -0,0 +1,69 @@
...
+    @Override
+    protected final void doGet(@Nonnull SlingHttpServletRequest request, @Nonnull SlingHttpServletResponse response)
+            throws ServletException, IOException {
+
+        JcrQueryHelper queryHelper = new JcrQueryHelper(request.getResourceResolver());
+        try {
+            Stream<Resource> resultStream = queryHelper.query(getQuery(request));
+            List<Map<String, Object>> resultList = resultStream.map(this::resourceToMap)
+                    .collect(Collectors.toList());
+            QuerySingleResult queryResult = new QuerySingleResult(resultList.get(0));
```
carlosmunoz: Probably need to check there is at least one result, otherwise this will throw an exception. 
And if there are no results found, this should probably return a 404 to the caller.

src/main/java/com/redhat/pantheon/servlet/ModuleRenderServlet.java
```diff
@@ -0,0 +1,128 @@
...
+        String localeParam = ServletUtils.paramValue(request, "locale", DEFAULT_MODULE_LOCALE.toString());
+        String uuidParam = paramValue(request, "module_id", "");
+        String nodeType = paramValue(request, "nodeType","pant:module");
+
+        StringBuilder query = new StringBuilder("select * from ")
+                .append("[").append(nodeType).append("]")
+                .append(" as module WHERE ")
+//                .append("where (isdescendantnode(module, '/content/repositories') ")
+//                .append("or isdescendantnode(module, '/content/modules') ")
+//                .append("or isdescendantnode(module, '/content/sandbox')) ")
```
carlosmunoz: Lets get rid of commented-out code, unless strictly necessary.

src/main/java/com/redhat/pantheon/servlet/ModuleRenderServlet.java
```diff
@@ -0,0 +1,128 @@
...
+ * The said url is accessible outside of the system without any authentication.
+ *
+ * @author Ankit Gadgil
+ */
+@Component(
+        service = Servlet.class,
+        property = {
+                "sling.servlet.resourceTypes=sling/servlet/default",
+                "sling.servlet.selectors=query",
+                "sling.servlet.extensions=json",
```
carlosmunoz: These mappings are not right. They will actually conflict with the default query servlet.
I think your annotations should look something line this:
```java
@Component(
        service = Servlet.class,
        property = {
                Constants.SERVICE_DESCRIPTION + "=...",
                Constants.SERVICE_VENDOR + "=...."
        })
// Use whatever path you are exposing here
@SlingServletPaths(value = "/api/module/render")
```

src/main/java/com/redhat/pantheon/servlet/ModuleRenderServlet.java
```diff
@@ -0,0 +1,128 @@
...
+@SlingServletPaths(value = "/api/module/render")
+public class ModuleRenderServlet extends AbstractJsonSingleQueryServlet {
+    private final Logger log = LoggerFactory.getLogger(ModuleRenderServlet.class);
+
+    @Override
+    protected String getQuery(SlingHttpServletRequest request) {
+        // Get the query parameter(s)
+        String localeParam = ServletUtils.paramValue(request, "locale", DEFAULT_MODULE_LOCALE.toString());
+        String uuidParam = paramValue(request, "module_id", "");
+        String nodeType = paramValue(request, "nodeType","pant:module");
```
carlosmunoz: I don't think this parameter is necessary. The api is hardcoded to look for modules.

src/main/java/com/redhat/pantheon/servlet/ModuleRenderServlet.java
```diff
@@ -0,0 +1,128 @@
...
+        Optional<ModuleRevision> releasedRevision = module.getReleasedRevision(DEFAULT_MODULE_LOCALE);
+
+        Map<String, Object> moduleMap = super.resourceToMap(resource);
+        Map<String, Object> moduleDetails = new HashMap<>();
+
+        moduleDetails.put("status", "SUCCESS");
+        moduleDetails.put("message", "Module Found");
+
+        String resourcePath = resource.getPath();
+        moduleMap.put("locale", module.getModuleLocale(DEFAULT_MODULE_LOCALE).getName());
```
carlosmunoz: I mentioned this already, but I'll do it again in the review. I think a better idea here, instead of using a `toMap` function, is to simply have a function like
```java
protected Object transform(Resource resource) {
    return resource.toValueMap();
}
```
This way the default transformation is just to a value map, but overrides may transform into anything that can be turned into json.

In our specific case, we create a `Module` DTO of sorts which has all these fields baked in as hava properties. e.g.
```java
// when converted to json this will generate a field of name "title"
String getTitle() {
   return title
}
```
Anyway, this can be a later improvement since our query method already has a `resourceToMap` method.

src/main/java/com/redhat/pantheon/model/QuerySingleResult.java
```diff
@@ -0,0 +1,20 @@
+package com.redhat.pantheon.model;
+
+import java.util.Map;
+
+/**
+ * A simple model to represent a single element result when running queries.
+ */
+public class QuerySingleResult {
```
carlosmunoz: I'm not sure we need this. If the only thing this is keeping is the result itself (and no additional information), why not just return the result directly. We are only adding one additional level to the json response. i.e:
```json
{
   "result": 
       "module" {
       }
}
```
vs
```json
"module" {
}
```
that intermediate `result` section is adding no value.

src/main/java/com/redhat/pantheon/servlet/AbstractJsonSingleQueryServlet.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.stream.Stream;
+
+import static com.google.common.collect.Maps.newHashMap;
+import static com.redhat.pantheon.servlet.ServletUtils.writeAsJson;
+
+/**
+ * A base servlet that helps to create simple Json result from custom queries.
+ * This servlet should be extended to create components that render a single element of
+ * results in json format. See methods to be overriden for details on how to
+ * customize sub classes.
```
carlosmunoz: Comments probably need to be updated as they seem like a copy-paste from the multi-result query servlet. Since you are executing any custom query, I would include something about how the servlet behaves when it finds more than one result. i.e. it renders the first one.

src/main/java/com/redhat/pantheon/servlet/ModuleRenderServlet.java
anknite: ahh, makes sense.

src/main/java/com/redhat/pantheon/servlet/ModuleRenderServlet.java
anknite: sure let's remove it, I left it there because we wanted to keep the code for future reference.

src/main/java/com/redhat/pantheon/servlet/ModuleRenderServlet.java
anknite: Makes sense, I was just trying to keep it generic lol

src/main/java/com/redhat/pantheon/servlet/AbstractJsonSingleQueryServlet.java
anknite: Good point.

src/main/java/com/redhat/pantheon/model/QuerySingleResult.java
anknite: Well, I wanted to maintain consistency between APIs we are building, we can remove it and just return the module object WDUT? @carlosmunoz

carlosmunoz: I don't think we need API's to be as consistent for the single-return ones. I mean, if the `result` object is going to potentially have other properties I say we leave it, but if we are just keeping it to maintain the same nomenclature at the first level, then there's no real reason to keep it. 

Based on this I say we remove it, and if we need to change it later (because we add more data to the response for example), then we handle it.

src/main/java/com/redhat/pantheon/servlet/AbstractJsonSingleQueryServlet.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    @Override
+    protected final void doGet(@Nonnull SlingHttpServletRequest request, @Nonnull SlingHttpServletResponse response)
+            throws ServletException, IOException {
+
+        JcrQueryHelper queryHelper = new JcrQueryHelper(request.getResourceResolver());
+        try {
+            Stream<Resource> resultStream = queryHelper.query(getQuery(request));
+            List<Map<String, Object>> resultList = resultStream.map(this::resourceToMap)
```
carlosmunoz: you only need to transform the first resource, not all of them. I would move this call to after you have extracted the first result.

carlosmunoz: I think, the idea is that yes it should return a single result (we have another abstract servlet for multiple results).

As to throwing an exception if there is more than one result... I’m not sure. It depends how much responsibilities we want to give the abstract class. We could leave said responsibilities to the sub-classes, or hard-code them here.

src/main/java/com/redhat/pantheon/servlet/ModuleRenderServlet.java
```diff
@@ -0,0 +1,121 @@
...
+ * @author Ankit Gadgil
+ */
+@Component(
+        service = Servlet.class,
+        property = {
+                Constants.SERVICE_DESCRIPTION + "=Servlet to facilitate GET operation which accepts locale and module uuid to output module data",
+                Constants.SERVICE_VENDOR + "=Red Hat Content Tooling team"
+        })
+@SlingServletPaths(value = "/api/module/render")
+public class ModuleRenderServlet extends AbstractJsonSingleQueryServlet {
```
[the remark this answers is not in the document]
carlosmunoz: This is a naming standard based on the initial decision. I agree with Ben that it might be an inaccurate name. Now is the time to change the naming if we are going to do so.

src/main/java/com/redhat/pantheon/servlet/AbstractJsonSingleQueryServlet.java
benradey: Perhaps we have a `protected Resource handleMultipleQueryResults(Stream<Resource> results)` method in the abstract class which by default just throws an exception in the face of multiple results. Then subclasses can override if/when it makes sense. That would be my suggestion.

src/main/java/com/redhat/pantheon/servlet/ModuleRenderServlet.java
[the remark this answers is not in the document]
anknite: Renaming the servlet makes sense and I am all for it.

src/main/java/com/redhat/pantheon/servlet/ModuleRenderServlet.java
```diff
@@ -0,0 +1,121 @@
...
+                .append("module.[jcr:uuid] = '" + uuidParam + "'");
+        return query.toString();
+    }
+
+    @Override
+    protected Map<String, Object> resourceToMap(@NotNull Resource resource) {
+        Module module = resource.adaptTo(Module.class);
+        // The DEFAULT_MODULE_LOCALE should later be replaced with 'localeParam' variable
+        // this needs to be done while handling localizations
+        Optional<Metadata> draftMetadata = module.getDraftMetadata(DEFAULT_MODULE_LOCALE);
```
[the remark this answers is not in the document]
anknite: Sure, there was no specific way for me to choose DEFAULT_MODULE_LOCALE
I was considering not transforming the DEFAULT_MODULE_LOCALE into a locale param till we get to localizations. However, this can make us ready for those changes.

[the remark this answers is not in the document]
anknite: The only major reason I did not do this is that the localeParam is local to the getQuery method and we would have to export those to the be used in the Map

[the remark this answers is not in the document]
benradey: Ah I see, I didn't notice that `localeParam` was a local variable for the `getQuery` method only. Here's one suggestion - perhaps in the `AbstractJsonSingleQueryServlet`, you store the locale in a private field in the `doGet` method, and expose it via a `protected final String getLocale()` method or something like that. Then the `resourceToMap` method can call it whenever necessary.

Anyway, just a thought, feel free to take it or leave it.

You are right that we don't support localizations yet, but in the cases where it's easy to add future support (like here), we should go ahead and do it.

src/main/java/com/redhat/pantheon/servlet/ModuleRenderServlet.java
anknite: That's a good improvement to look into.

src/main/java/com/redhat/pantheon/servlet/ModuleRenderServlet.java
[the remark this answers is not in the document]
anknite: While creating the API fro modules I came across this issue of using locale=en_us as a query param. The code checks for existing locale param from the API url and if not present uses the DEFAULT_MODULE_LOCALE

While converting the the locale string to a Locale object:

localeParam = new Locale(ServletUtils.paramValue(request, "locale", DEFAULT_MODULE_LOCALE.toString()));

While adding as a Locale object the locale variable (string) is converted to a lowercase string by Java:  https://docs.oracle.com/javase/7/docs/api/java/util/Locale.html
This results in a mismatch on the locale for released and draft revisions for a module. Sling is not able to find the releasedRevision for the module and the API throws a 404

Proposed Solution: We need to change the way we store locale as a node.
The typography on the locale node will have to be a lower cased locale, example en_usand not en_US

src/main/java/com/redhat/pantheon/servlet/ModuleRenderServlet.java
```diff
@@ -0,0 +1,121 @@
...
+                Constants.SERVICE_VENDOR + "=Red Hat Content Tooling team"
+        })
+@SlingServletPaths(value = "/api/module/render")
+public class ModuleRenderServlet extends AbstractJsonSingleQueryServlet {
+    private final Logger log = LoggerFactory.getLogger(ModuleRenderServlet.class);
+
+    @Override
+    protected String getQuery(SlingHttpServletRequest request) {
+        // Get the query parameter(s)
+        String localeParam = ServletUtils.paramValue(request, "locale", DEFAULT_MODULE_LOCALE.toString());
```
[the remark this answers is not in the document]
anknite: removed.

src/test/java/com/redhat/pantheon/servlet/ModuleRenderServletTest.java
```diff
@@ -0,0 +1,82 @@
...
+//        // When
+//        Map<String, Object> map = servlet.resourceToMap(
+//                slingContext.resourceResolver().getResource("/content/repositories/repo/module/en_US/1"));
+//
+//        // Then
+//        assertTrue(map.containsKey("module"));
+//        assertTrue(map.containsKey("status"));
+//        assertTrue(map.containsKey("message"));
+////        assertTrue(map.containsKey("pant:transientSource"));
+//    }
```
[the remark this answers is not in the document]
anknite: it was in-progress :p

src/main/java/com/redhat/pantheon/servlet/ModuleRenderServlet.java
anknite: changed.

src/main/java/com/redhat/pantheon/servlet/AbstractJsonSingleQueryServlet.java
anknite: changed

src/test/java/com/redhat/pantheon/servlet/ModuleRenderServletTest.java
```diff
@@ -0,0 +1,82 @@
...
+        map.put("locale", DEFAULT_MODULE_LOCALE.toString());
+        map.put("module_id", "jcr:uuid");
+        slingContext.request().setParameterMap(map);
+
+        // When
+        String query = servlet.getQuery(slingContext.request());
+
+        // Then
+        // make sure queries don't throw exceptions when executed against the JCR repository
+        assertDoesNotThrow(() -> slingContext.resourceResolver().queryResources(query, Query.JCR_SQL2));
```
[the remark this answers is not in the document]
anknite: Added a test to check slingContext.response() is not null

src/main/java/com/redhat/pantheon/servlet/ModuleRenderServlet.java
carlosmunoz: Um... this seems to be the same servlet as `ModuleJsonServlet` is it not?
I think this one is a duplicate.

src/test/java/com/redhat/pantheon/servlet/ModuleJsonServletTest.java
```diff
@@ -0,0 +1,102 @@
...
+        Map<String, Object> map = servlet.resourceToMap(
+                slingContext.resourceResolver().getResource("/content/repositories/repo/module"));
+
+        // Then
+        assertTrue(map.containsKey("module"));
+        assertTrue(map.containsKey("status"));
+        assertTrue(map.containsKey("message"));
+        assertEquals((map.get("message")), "Module Found");
+        assertEquals((map.get("status")), "SUCCESS");
+        assertEquals((map.get("module").toString()), expectedModuleJson);
```
carlosmunoz: These tests tend to be very brittle (i.e. the slightest format change will trigger a test failure, even if the json format itself is still valid and complete). I won't stop the PR for this one, but in the future lets change/stay away from these types of tests.

anknite: I do agree with your point of not using this kind of hardcoded testing mocks.
However, IMO we should either move towards testing API endpoints against mock data or test if the responses contain required keys or not. This was an attempt to check against mock of real data to make sure the response is as expected.
We might need to take a call on which type tests we want to go with for API endpoints oing ahead.

src/main/java/com/redhat/pantheon/servlet/ModuleRenderServlet.java
anknite: Ha, I was under the impression the IDE must have handled this when I renamed the ModuleRenderServlet to ModuleJasoServlet
Looks like I missed, it did not.

anknite: I've removed the file.

anknite:
Is this fine to be merged to master then @carlosmunoz

carlosmunoz:
I think our **current** process is still to merge to master and test afterwards.

PR #116, opened 2019-09-03
Set server only if PANTHEON_SERVER ENVVAR is not set.

jmarrero:
Introduce an Environment variable that will supersede user configuration. This is intended to be used on Openshift and automated deployments where git2pantheon calls the uploader. This will ensure that git2pantheon and the uploader talk to the intended environment and will disregard the pantheon2.yml server configuration that an user might be using to test locally.

Proof that the code works:
Var not set:
```
$ python3 uploader/pantheon.py push -d ~/Development/adocs/ 
Could not find a valid config file(pantheon2.yml) in this directory; all files will be treated as resource uploads.
server http://localhost:8080 is not reachable
```
I set the variable:
```
$ export PANTHEON2_SERVER=coolbeans.redhat.com
```

Result with the variable set:
```
$ python3 uploader/pantheon.py push -d ~/Development/adocs/ 
Could not find a valid config file(pantheon2.yml) in this directory; all files will be treated as resource uploads.
server coolbeans.redhat.com is not reachable
```


jmarrero:
@ivanhorvath changed the ENV to PANTHEON_SERVER and added a workaround for getting the username inside the container. This should hold us over until the mayor refactor.

PR #122, opened 2019-09-10
CCS 2986 revisions

ankkit24:
Pull request for module display and module revisions.

* Edit Meta data will be in a separate PR.
* Unpublish functionality will be addressed when code for the same is merged.
* Product details are not implemented in this PR.


benradey:
Noticed an issue: if you go to a module detail page and click the 'Modules' breadcrumb (which takes you back to search results), then click into a module detail page again and click the 'Modules' breadcrumb again, it does nothing the 2nd time around. I think it has something to do with the URL being different between the 1st module detail page and the 2nd module detail page.

benradey:
Another note: If you go to a module detail page and press refresh, you are taken back to module search.

benradey:
What's up with the files added in /node_modules/yarn/? Are those new/required?

PR #124, opened 2019-09-17
fix uploaded date module sorting

carlosmunoz:
create new module revision metadata field for upload date
fix non-working sorting by other fields
changed response code to 201 when creating a module vs updating
revert module jcr query to search accross all nodes


anknite:
This looks good to me.

PR #129, opened 2019-09-20
Renaming all instances of 'module revision' to 'module version'

benradey:


jmarrero:
@benradey I think Lisa and Ankit N have made some changes to some of these places. Maybe double check that there are not more instances of revision now before merging.

PR #134, opened 2019-09-27
Stop trying to route to /logout when that route does not exist

jmarrero:


xdavidson:
according to this chart there is no change in coverageL  
https://codecov.io/gh/redhataccess/pantheon/compare/fix-logout...master

carlosmunoz:
I don't think you meant to close the PR, did you? :laughing:

xdavidson:
> I don't think you meant to close the PR, did you?

somebody is too quick!

PR #144, opened 2019-10-10
Changes on module api

anknite:
This is a simple API property value improvement PR for:

- correcting the value sent on title
- correcting the module_url_fragment

has no effect on test coverage


anknite:
Thanks Ben

PR #146, opened 2019-10-11
Repeated modules

ankkit24:
This PR relates to the removal of duplicate values in the search page whenever a new draft is uploaded.

Expected output: 
* Each module should have just one record in the search page.
* The draft or released versions of that module can be viewed on its corresponding module display page.


src/main/java/com/redhat/pantheon/servlet/ModuleListingServlet.java
```diff
@@ -52,14 +52,11 @@ protected String getQuery(SlingHttpServletRequest request) {
...
         StringBuilder queryBuilder = new StringBuilder()
                 .append("select m.* from [nt:base] as m ")
-                    .append("INNER JOIN [pant:moduleVersion] as rev ON ISDESCENDANTNODE(rev, m) ")
+                    .append("INNER JOIN [pant:moduleLocale] as rev ON ISDESCENDANTNODE(rev, m) ")
                 .append("where (isdescendantnode(m, '/content/repositories') ")
                     .append("or isdescendantnode(m, '/content/modules') ")
                     .append("or isdescendantnode(m, '/content/sandbox')) ")
-                .append("AND m.[jcr:primaryType] = 'pant:module' ")
-                // look in ALL versions (all locales)
-                .append("AND (rev.[metadata/jcr:title] like '%" + searchParam + "%' ")
```
carlosmunoz: This removes filtering by title and description.

src/main/java/com/redhat/pantheon/servlet/ModuleListingServlet.java
```diff
@@ -52,14 +52,11 @@ protected String getQuery(SlingHttpServletRequest request) {
         // is not returning any results
         StringBuilder queryBuilder = new StringBuilder()
                 .append("select m.* from [nt:base] as m ")
```
carlosmunoz: You might want to try a `select distinct` instead if you want to remove duplicate entries

frontend/src/app/versions.tsx
```diff
@@ -372,7 +372,7 @@ class Versions extends Component<IProps, any> {
                         const draftTag = responseJSON.draft;
                         const versionCount = responseJSON.__children__.length
 
-                        for (let i = versionCount - 1; i > versionCount - 3 && i >= 0; i--) {
+                        for (let i = versionCount - 1; i >= 0; i--) {
```
benradey: What is the rationale for removing the condition here? Especially because this PR is for search results, and versions.tsx is unrelated to that.

src/main/java/com/redhat/pantheon/servlet/ModuleListingServlet.java
benradey: Agreed, if that's an option, it seems ideal.

src/main/java/com/redhat/pantheon/servlet/ModuleListingServlet.java
```diff
@@ -52,14 +52,11 @@ protected String getQuery(SlingHttpServletRequest request) {
...
-                    .append("INNER JOIN [pant:moduleVersion] as rev ON ISDESCENDANTNODE(rev, m) ")
+                    .append("INNER JOIN [pant:moduleLocale] as rev ON ISDESCENDANTNODE(rev, m) ")
```
benradey: I think this might not ultimately solve the problem. Today we only have en_US locale nodes in the system, what about in the future where we have other locales? Would we get duplicate results at that point with this query?

src/main/java/com/redhat/pantheon/servlet/ModuleListingServlet.java
benradey: To draw clearer attention to it, it also makes `searchParam` a completely unused variable.

This would effectively remove the capability to search from our search page, turning it instead into a simple module listing page.

benradey:
HMMMMM.

Unfortunately, this does not seem to fix the problem - which sucks, because we spent so much time putting our heads together last week and I was sure we had something working at that point.

Regardless, I am still seeing that when I upload a module and create 3 versions, I get 3 duplicate search results. Maybe we need to re-evaluate our query.

benradey:
Here is the query that I used which I extracted straight from the servlet to simulate an empty (default) search query:

SELECT m.* from [nt:base] AS m  LEFT OUTER JOIN [nt:base] AS loc ON  ISDESCENDANTNODE(loc, m)  LEFT OUTER JOIN [nt:base] AS draft ON  draft.[jcr:uuid] = loc.[draft]  LEFT OUTER JOIN [nt:base] AS release ON  release.[jcr:uuid] = loc.[released]  LEFT OUTER JOIN [nt:base] AS ver ON  ISDESCENDANTNODE(ver, loc)  WHERE m.[jcr:primaryType] = 'pant:module'  AND loc.[jcr:primaryType] = 'pant:moduleLocale'  AND ver.[jcr:primaryType] = 'pant:moduleVersion'  AND (draft.[jcr:primaryType] = 'pant:moduleVersion' OR draft.[jcr:primaryType] IS NULL)  AND (release.[jcr:primaryType] = 'pant:moduleVersion' OR release.[jcr:primaryType] IS NULL)  AND (draft.[metadata/jcr:title] LIKE '%%'  OR draft.[metadata/jcr:description] LIKE '%%'  OR release.[metadata/jcr:title] LIKE '%%'  OR release.[metadata/jcr:description] LIKE '%%')

Attaching the result as a screenshot.

Untitled

benradey:
Note that when I run THIS query - with both lines that reference `ver` removed - this works:

SELECT m.* from [nt:base] AS m 
LEFT OUTER JOIN [nt:base] AS loc ON  ISDESCENDANTNODE(loc, m) 
LEFT OUTER JOIN [nt:base] AS draft ON  draft.[jcr:uuid] = loc.[draft] 
LEFT OUTER JOIN [nt:base] AS release ON  release.[jcr:uuid] = loc.[released] 
WHERE m.[jcr:primaryType] = 'pant:module' 
AND loc.[jcr:primaryType] = 'pant:moduleLocale' 
AND (draft.[jcr:primaryType] = 'pant:moduleVersion' OR draft.[jcr:primaryType] IS NULL) 
AND (release.[jcr:primaryType] = 'pant:moduleVersion' OR release.[jcr:primaryType] IS NULL) 
AND (draft.[metadata/jcr:title] LIKE '%%' 
OR draft.[metadata/jcr:description] LIKE '%%' 
OR release.[metadata/jcr:title] LIKE '%%' 
OR release.[metadata/jcr:description] LIKE '%%')

So you're close, you just need to fix the `ver` part (which I think is used for sorting and nothing else, right?).

benradey:
Commenting out the `ver` lines in the servlet (and as a consequence, the `if` block which does the sorting as well) also remedies the search page, FYI.

ankkit24:
@benradey You are right, this does show duplicate results now.
Probably need a work around for the ORDER BY condition, the rest of the query looks good.

src/main/java/com/redhat/pantheon/servlet/ModuleListingServlet.java
```diff
@@ -51,19 +53,23 @@ protected String getQuery(SlingHttpServletRequest request) {
...
+                .append("AND (draft.[metadata/jcr:title] LIKE '%" + searchParam + "%' ")
+                    .append("OR draft.[metadata/jcr:description] LIKE '%" + searchParam + "%' ")
+                    .append("OR release.[metadata/jcr:title] LIKE '%" + searchParam + "%' ")
+                    .append("OR release.[metadata/jcr:description] LIKE '%" + searchParam + "%')");
 
         if(!isNullOrEmpty(keyParam) && !isNullOrEmpty(directionParam)) {
-            queryBuilder.append(" order by rev.[metadata/")
-                    .append(keyParam).append("] ")
+            queryBuilder.append(" ORDER BY coalesce(release.[metadata/")
+                    .append(keyParam).append("],draft.[metadata/")
```
benradey: Hey Ankit, please reverse the order between `release` and `draft` here. We need to have `draft` as the first parameter for `coalesce`, followed by `release`. If you have `release` first, you get weird behavior when sorting by Upload Time - where the thing that you just uploaded a draft version for doesn't necessarily get sorted to the top of the list. I mode the code update on my local system and it appears to work properly. Let me know if you disagree with the change. Otherwise this looks great. Thanks!

ankkit24: Hey Ben, 
Thanks for pointing this out.
I have made the changes and tested and it looks good to me now.
Let me know if there are any more issues.

benradey:
Looks good to me, Ankit. I am going to dismiss Carlos' stale review, I believe he would approve of these changes as well.

PR #151, opened 2019-10-16
Podman instructions

carlosmunoz:
Add documentation on how to run with podman using pods and an included Kubernetes pod definition.


jmarrero:
:shipit:

PR #155, opened 2019-10-21
Update haml templates

carlosmunoz:
Update the asciidoctor backend templates to use their latest version.
Move the raw asciidoc content servlet to the 'module' package.


carlosmunoz:
It does. I myself do not have a good grasp on the changes, they simply reflect the latest work from Mark on the asciidoctor backend so I assume they are correct. At one point the team will need to get familiar with haml and these templates so we can move them forward.

PR #157, opened 2019-10-22
Ccs 2962 Add validation before publish a module

xdavidson:
This PR includes:

- Ensure Metadata is populated before publishing a module
- fix various tslint complaints


frontend/src/app/versions.tsx
```diff
@@ -116,14 +119,23 @@ class Versions extends Component<IProps, any> {
...
+                {this.state.successAlertVisible && <Alert
                     variant="success"
                     title="Edit Metadata"
                     action={<AlertActionCloseButton onClose={this.hideSuccessAlert} />}
                 >
                     Update Successful!
           </Alert>
                 }
+
+                {this.state.publishAlertVisible && this.state.canChangePublishState === false && <Alert
```
benradey: Why is `canChangePublishState` also in this conditional? Why isn't `publishAlertVisible` sufficient by itself?

frontend/src/app/versions.tsx
```diff
@@ -419,29 +432,36 @@ class Versions extends Component<IProps, any> {
     }
 
     private changePublishState = (buttonText) => {
-        const formData = new FormData();
-        if (buttonText === "Publish") {
-            formData.append(":operation", "pant:release");
-            // console.log('Published file path:', this.props.modulePath)
-            this.draft[0].version = "";
+        // Validate productValue before Publishs
+        if (this.props.productInfo !== undefined && this.props.productInfo.trim() === "" && buttonText === "Publish") {
```
benradey: If `productInfo` IS undefined, wouldn't we want to show the dialog in that scenario as well?

frontend/src/app/versions.tsx
```diff
@@ -419,29 +432,36 @@ class Versions extends Component<IProps, any> {
...
+                fetch("/content" + this.props.modulePath, {
+                    body: formData,
+                    method: 'post'
+                }).then(response => {
+                    if (response.status === 201 || response.status === 200) {
+                        console.log(buttonText + " works: " + response.status)
+                        this.setState({ initialLoad: true, changePublishState: true, publishAlertVisble: false, canChangePublishState: true })
+                    } else {
+                        console.log(buttonText + " failed " + response.status)
+                        this.setState({ initialLoad: true, changePublishState: true, publishAlertVisble: true })
```
benradey: These references to `publishAlertVisible` and `changePublishState` don't appear to be necessary since neither field is used anywhere else in the file.

frontend/src/app/versions.tsx
benradey: Actually, now that I'm thinking about it, I would really like this to be a server-side check. Although 99% of the time the users will be publishing from our web UI - we can't guarantee that that will always be the case. I'd like to see the server providing a computed property with the metadata - something like `canPublish` - that the UI keys off of, rather than making the decision at the UI layer like you have here. Can you implement that?

frontend/src/app/versions.tsx
xdavidson: canChangePublishState is introduced in this feature branch as a flag that checks if metadata is set before a version can be published

frontend/src/app/versions.tsx
xdavidson: Ben, 

I agree that we should have a server side validation. Let's log a jira to capture that effort so that we don't expand the scope of this jira.

frontend/src/app/versions.tsx
benradey: Ok, I understand the function of the property, I'm just not sure why it appears in this specific place in the code. Would we ever run into a scenario where `publishAlertVisible` was true and `canChangePublishState` was also true? If not then why not simplify the code and remove `canChangePublishState` from the conditional?

frontend/src/app/versions.tsx
benradey: I agree that the server-side check is extra effort, but I don't like the idea of committing this UI-side check with the intention of fixing it up properly later (but maybe never). I checked the story in Jira and it doesn't specify one way or the other, so I'd really prefer that we implement this the right way the first time. If we need to repoint the story as a result of the increased effort, then we can do so.

frontend/src/app/versions.tsx
xdavidson: publishAlertVisible is used to show the alert message when medata data is not updated before publishing. 

I removed "changePublishState"

frontend/src/app/versions.tsx
xdavidson: CCS 2962 is updated to reflect that UI work. 
I created a new jira for the backend effort: https://projects.engineering.redhat.com/browse/CCS-3237

benradey: Well, I still think we could have resized the original story, but that's a small detail. Just to be sure, is your intention to work that new story before moving forward with this PR?

frontend/src/app/versions.tsx
benradey: So are you saying that the behavior of the application DOES change if we remove THIS particular check for `canChangePublishState`?

frontend/src/app/versions.tsx
benradey: Ah... it appears you have a typo here then. The `publishAlertVisble` that you have here is missing an 'i' in the field name. Elsewhere in the code, there is no typo, so this is technically a different field. Subsequently, this code probably does not do what you think it does.

frontend/src/app/versions.tsx
xdavidson: Yes, it stops working.

frontend/src/app/versions.tsx
xdavidson: My goal is to get this one moved forward then address other stories.

frontend/src/app/versions.tsx
xdavidson: the typo has been corrected.

frontend/src/app/versions.tsx
xdavidson: Ben, per our offline conversation, I pushed up the validation for ReleaseDraftVersion servlet

frontend/src/app/versions.tsx
benradey: So I looked into this myself and, lo and behold, removing this check DOES change the behavior of the application!

...but, it's because of a bug elsewhere in the code (a small one). :)

Rather than try and describe it here, I just fixed it in my own branch. I'll open a PR for it.

src/main/java/com/redhat/pantheon/servlet/module/ReleaseDraftVersion.java
```diff
@@ -51,6 +51,16 @@ protected void doRun(SlingHttpServletRequest request, PostResponse response, Lis
...
+        } else if (versionToRelease.get().metadata.getOrCreate().productVersion.get() == null
+        		||  versionToRelease.get().metadata.getOrCreate().productVersion.get().isEmpty()) {
+        	// Check if productVersion is set
+        	response.setStatus(HttpServletResponse.SC_PRECONDITION_FAILED,
+                    "The version to be released doesn't have productVersion metadata");
+        } else if (versionToRelease.get().metadata.getOrCreate().urlFragment.get() == null
+        		||  versionToRelease.get().metadata.getOrCreate().urlFragment.get().isEmpty()) {
+        	// Check if urlFragment is set
+        	response.setStatus(HttpServletResponse.SC_PRECONDITION_FAILED,
+                    "The version to be released doesn't have urlFragment metadata");
```
benradey: Tested and it works!

Because I'm a pain-in-the-ass perfectionist, I have to point out that technically, this code doesn't match what the UI checks for. The UI only checks for productVersion, but this checks for both productVersion and urlFragment.

That kind of works because the UI is operating on an all-or-nothing basis right now, but that might not always be the case. The cleanest implementation of this would check for everything that's relevant (product version, url fragment, and search keywords) in both places, but since we've gone back and forth on this PR enough already, let's save those details for another day.

carlosmunoz:
@xdavidson just wanted to point out that this one went in while lowering our code coverage.

PR #165, opened 2019-11-06
CCS-3208: add a static image delivery mechanism

carlosmunoz:
Create a static image delivery mechanism for proxies and edge platforms. The mechanism consists of two parts:

1. A re-encoding of all generated `` tags in the generated html so that they refer to images in an absolute way. The resulting html will have image tags of the form:
`` with `{unique image id}` being a url friendly string by which the image can be uniquely identified in the system.

2. A system endpoint capable of handling the requested image asset urls, decoding the unique identifier and serving the image properly.

This mechanism should allow the edge to detect calls to the unique prefix in the system and route them appropriately to pantheon for subsequent caching.

### DIFF
--- a/pantheon-bundle/src/main/java/com/redhat/pantheon/asciidoctor/extension/ImageSrcTransformer.java
+++ b/pantheon-bundle/src/main/java/com/redhat/pantheon/asciidoctor/extension/ImageSrcTransformer.java
@@ -1,0 +1,52 @@
+package com.redhat.pantheon.asciidoctor.extension;
+
+import org.apache.sling.api.resource.Resource;
+import org.apache.sling.api.resource.ResourceResolver;
+import org.asciidoctor.ast.Document;
+import org.asciidoctor.extension.Postprocessor;
+import org.jsoup.Jsoup;
+
+import java.util.Base64;
+
+import static com.redhat.pantheon.conf.GlobalConfig.IMAGE_PATH_PREFIX;
+
+/**
+ * An asciidoctor post processor which transforms all image locations in the resulting html to an encoded and absolutely
+ * resolving path comprised of a prefix and a unique identifier for the image. This unique identifier consists of the
+ * Base64 encoding of the actual absolute path of the image resource in the JCR repository.
+ *
+ * @see com.redhat.pantheon.servlet.assets.ImageServletFilter for the servlet serving these paths
+ * @author Carlos Munoz
+ */
+public class ImageSrcTransformer extends Postprocessor {
+
+    private final Resource module;
+
+    public ImageSrcTransformer(Resource module) {
+        this.module = module;
+    }
+
+    @Override
+    public String process(Document document, String output) {
+        org.jsoup.nodes.Document doc = Jsoup.parse(output, "UTF-8");
+
+        ResourceResolver resourceResolver = module.getResourceResolver();
+
+        // add a prefix to all images
+        doc.select("img")
+                .forEach(imageElement -> {
+                    String imgSrc = imageElement.attr("src");
+                    Resource resolvedImage = resourceResolver.getResource(module.getParent(), imgSrc);
+                    if(resolvedImage != null) {
+                        imageElement.attr("src", encodeImgSrc(resolvedImage.getPath()));
+                    }
+                });
+        output = doc.html();
+
+        return output;
+    }
+
+    String encodeImgSrc(String imageResourcePath) {
+        return IMAGE_PATH_PREFIX + "/" + Base64.getUrlEncoder().encodeToString(imageResourcePath.getBytes());
+    }
+}
--- a/pantheon-bundle/src/main/java/com/redhat/pantheon/asciidoctor/extension/ImageSrcTransformer.java
+++ b/pantheon-bundle/src/main/java/com/redhat/pantheon/asciidoctor/extension/ImageSrcTransformer.java
@@ -1,5 +1,6 @@
 package com.redhat.pantheon.asciidoctor.extension;
 
+import org.apache.jackrabbit.oak.commons.PathUtils;
 import org.apache.sling.api.resource.Resource;
 import org.apache.sling.api.resource.ResourceResolver;
 import org.asciidoctor.ast.Document;
@@ -36,10 +37,8 @@ public class ImageSrcTransformer extends Postprocessor {
         doc.select("img")
                 .forEach(imageElement -> {
                     String imgSrc = imageElement.attr("src");
-                    Resource resolvedImage = resourceResolver.getResource(module.getParent(), imgSrc);
-                    if(resolvedImage != null) {
-                        imageElement.attr("src", encodeImgSrc(resolvedImage.getPath()));
-                    }
+                    String imagePath = PathUtils.concat(module.getParent().getPath(), imgSrc);
+                    imageElement.attr("src", encodeImgSrc(imagePath));
                 });
         output = doc.html();
 
--- a/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/assets/ImageServletFilter.java
+++ b/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/assets/ImageServletFilter.java
@@ -1,0 +1,57 @@
+package com.redhat.pantheon.servlet.assets;
+
+import org.apache.sling.servlets.annotations.SlingServletFilter;
+import org.osgi.service.component.annotations.Component;
+
+import javax.servlet.Filter;
+import javax.servlet.FilterChain;
+import javax.servlet.FilterConfig;
+import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import javax.servlet.http.HttpServletRequest;
+import java.io.IOException;
+import java.util.Base64;
+
+import static com.redhat.pantheon.conf.GlobalConfig.IMAGE_PATH_PREFIX;
+
+/**
+ * A servlet filter which serves image assets only. Images must be referened using an absolute path comprised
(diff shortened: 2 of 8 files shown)


pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/assets/ImageServletFilter.java
```diff
@@ -0,0 +1,57 @@
...
+ * in the system.
+ *
+ * @author Carlos Munoz
+ */
+@Component(
+        service = Filter.class
+)
+@SlingServletFilter(
+        methods = "GET",
+        pattern = "/imageassets/.*")
```
benradey: I'm still not wild about using a Filter *as* a servlet, but this does appear to be the only way to do it. I tested a regular servlet with `@SlingServletPaths(value = "/pantheon/ben/.*")` and it did not behave the way I hoped. I had to literally make a request to `/pantheon/ben/.*` to execute the servlet - requesting `/pantheon/ben/anything` produced a 404.

One other possibility - we could write our own `DefaultErrorHandlerServlet` that is able to check for cases where the request starts with `/imageassets/` and do something clever there. We could also extend the servlet to check for things like `/modules/by-uuid/` and do something different. Not sure that that's any better really... but might be worth debating the pros and cons.

pantheon-bundle/src/test/java/com/redhat/pantheon/servlet/assets/ImageServletFilterTest.java
```diff
@@ -0,0 +1,57 @@
...
+        when(requestDispatcherFactory.getRequestDispatcher(anyString(), any())).thenReturn(requestDispatcher);
+        ImageServletFilter filter = new ImageServletFilter();
+
+        // When
+        filter.init(mock(FilterConfig.class));
+        filter.doFilter(sc.request(), sc.response(), filterChain);
+
+        // Then
+        verify(requestDispatcherFactory).getRequestDispatcher(eq(imagePath), any());
+        verify(requestDispatcher).forward(any(), any());
```
benradey: Looks good, I think this is just verifying that your filter forwards the request when appropriate, right? Might be worth it to include a negative test - ensure that your filter doesn't interfere in a request that doesn't start with `IMAGE_PATH_PREFIX`.

pantheon-bundle/src/main/java/com/redhat/pantheon/asciidoctor/extension/ImageSrcTransformer.java
```diff
@@ -0,0 +1,52 @@
...
+        org.jsoup.nodes.Document doc = Jsoup.parse(output, "UTF-8");
+
+        ResourceResolver resourceResolver = module.getResourceResolver();
+
+        // add a prefix to all images
+        doc.select("img")
+                .forEach(imageElement -> {
+                    String imgSrc = imageElement.attr("src");
+                    Resource resolvedImage = resourceResolver.getResource(module.getParent(), imgSrc);
+                    if(resolvedImage != null) {
```
benradey: Are you sure we want to be making this check? I just tested a file with a broken image link (pointing at ./images/tiger.png instead of ./tiger.png where the file was actually located) and was confused for some time why the code appeared not to be doing the substitution. I don't think we gain anything by skipping the logic in this scenario. It's broken either way - let's at least make the behavior uniform, unless you can think of a string argument why we should not do so.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/assets/ImageServletFilter.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Override
+    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain) throws IOException, ServletException {
+        // Get the id, everything after the prefix
+        String contextPath = ((HttpServletRequest)request).getPathInfo();
+        contextPath = contextPath.substring((IMAGE_PATH_PREFIX + "/").length());
+
+        String imagePath = new String(Base64.getUrlDecoder().decode(contextPath));
+        request.getRequestDispatcher(imagePath)
+            .forward(request, response);
```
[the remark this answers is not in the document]
benradey: Proof of concept:
```
$ echo -n '/conf/pantheon.json' | base64
L2NvbmYvcGFudGhlb24uanNvbg==
```

Making the request to `http://localhost:8080/imageassets/L2NvbmYvcGFudGhlb24uanNvbg==` produces identical output to `http://localhost:8080/conf/pantheon.json`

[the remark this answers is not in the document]
benradey: image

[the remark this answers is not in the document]
carlosmunoz: In the current revision, we are not checking that it just works for images. So as you tested, any resource can be accessed by this mechanism (which is actually interesting, this might just be a unique id servlet for all resources). 

Users wouldn't be able to have access to anything they don't already have access to. When the filter forwards the request, it will check all necessary permissions. So, your proof of concept allows you to access the resource because in your browser session you already have admin access to all resources, but try from a non-authenticated context like this:

```
curl http://localhost:8080/imageassets/L2NvbmYvcGFudGhlb24uanNvbg==
```
and you should get a 404

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/assets/ImageServletFilter.java
carlosmunoz: This is a restriction in Sling's implementation of the servlet spec. In other implementations like Tomcat or Jetty you are allowed to map servlets to wildcard patterns.

Your suggested approach seems similar to the Filter approach, just using a different mechanism (an error handler). In which case I'd argue a filter is still a better approach. But I admit I still don't quite understand your concern about using a filter. I actually see this as a perfect example of how it should be used. The server is capturing a request, modifying it and then sending it the right way.

pantheon-bundle/src/test/java/com/redhat/pantheon/servlet/assets/ImageServletFilterTest.java
carlosmunoz: :+1: I can add that

pantheon-bundle/src/main/java/com/redhat/pantheon/asciidoctor/extension/ImageSrcTransformer.java
carlosmunoz: makes sense... I'll remove the check.

pantheon-bundle/src/test/java/com/redhat/pantheon/servlet/assets/ImageServletFilterTest.java
carlosmunoz: I think I read this wrong the first time.
I can't exactly test for that (at least not in a unit test) because that falls outside the responsibility of the filter. The filter assumes that it is being invoked. I would need to test the servlet framework class parsing the url and constructing the filter chain to make sure this filter is only invoked when the path matches.

pantheon-bundle/src/main/java/com/redhat/pantheon/asciidoctor/extension/ImageSrcTransformer.java
carlosmunoz: Ok, I spoke to soon on this one too.

I still need to make the check, because I need to encode the absolute path of the resource, and for that I need the resource not to be null.
What I could do in case of the resource not being found is to encode whatever path is in the image. This would ensure all images are in the same format, but it would add an extra step to debugging a non-existing image, because a developer would have to decode the path to get to the actual path being resolved. Not a big deal, so I'm still good if you want to encode all paths.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/assets/ImageServletFilter.java
benradey: The "sending it the right way" part is what I've got a problem with. In my mind, a request ought to be handled successfully (albeit perhaps not perfectly) even if you remove all filters from the system. So filters are "nice to have" not necessarily "critical to have." In this case, removing the filter would cause the request to fail outright.

I agree that this is the best we can do in Sling, and that the Error Handler approach isn't "better," just "different." But I do think that logically it makes more sense. Sling is throwing up its hands and saying "I don't know what to do with this, deal with it somehow" and then we deal with it however we like.

I don't feel strongly enough about it to ask for a change, but it's something we can keep in mind if circumstances change in the future, such as if we need to implement another filter for some other feature and they wind up conflicting with each other.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/assets/ImageServletFilter.java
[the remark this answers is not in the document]
benradey: Fair enough.

When I make the `imageassets` request as...
admin - it works
demo - it works
unauthenticated - it 404's
...which matches the behavior I get from the `/conf/pantheon.json` request as well.

So, I'm happy.

pantheon-bundle/src/test/java/com/redhat/pantheon/servlet/assets/ImageServletFilterTest.java
benradey: Ok, I'll buy that argument.

pantheon-bundle/src/main/java/com/redhat/pantheon/asciidoctor/extension/ImageSrcTransformer.java
benradey: Hmm. Let me think on this for a bit.

PR #167, opened 2019-11-11
Force the React files to be overwritten when deploying a new version

jmarrero:
This changes are based on this doc:
 https://sling.apache.org/documentation/bundles/content-loading-jcr-contentloader.html

I confirmed that when I re-installed the bundle the static file's dates were updated.

### DIFF
--- a/pantheon-bundle/pom.xml
+++ b/pantheon-bundle/pom.xml
@@ -44,7 +44,7 @@
             </resource>
             <resource>
                 <directory>frontend/dist</directory>
-                <targetPath>SLING-INF/content/pantheon/</targetPath>
+                <targetPath>SLING-INF/SPA</targetPath>
                 <filtering>false</filtering>
             </resource>
         </resources>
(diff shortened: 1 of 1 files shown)


xdavidson:
I tested out this branch. After I new deployment in the jcr tree
    

1.  content/pantheon/app.bundle.js(sling:folder) has the old timestamp while it's child node  "jcr:content" has the new jcr:lastModified timestamp, which means the content is updated!
2.   content/pantheon/app.bundle.js.map(sling:folder) has the old timestamp while its child node  
"jcr:content" has the new jcr:lastModified timestamp, which means the content is updated!
 3.  content/pantheon/app.css(sling:folder) has the old timestamp while its child node "jcr:content" has the new jcr:lastModified timestamp,  which means the content is updated!
 4.  content/pantheon/fonts(sling:folder) has the new timestamp and its children nodes have the new timestamp as well!
 5.  content/pantheon/images(sling:folder)  has the new timestamp and its children nodes have the new timestamp as well!
 6.  content/index.html(nt:file) has the old timestamp while its child node "jcr:content" has the new timestamp. Does that mean index.html is updated?

jmarrero:
I am not sure if that means that /content/ rules for index are being ignored. 

I think we might need to deploy into the stage environment to double check that the problem gets fixed. I building an image now. I was just looking at the /content/pantheon folder. It might be the case that in our local environments we don't see the issue at all and these rules are being ignored? We always are able to get the new react code locally.

jmarrero:
last modified get's updated on stage and I don't see /content/index.html on stage it should be /content/pantheon/index.html I don't see anything bad happening so we could iterate over this if we see the issue again. But If you guys were not seeing the lastModified being updated it does now.

PR #176, opened 2019-12-05
map pantheon environemnt with portal

xdavidson:
This PR introduces a mapping between pantheon 2 environment and customer portal environment. So that after a module is published, we have the correct url to customer portal for the corresponding environment. 

An new environment variable is introduced during run time. To start pantheon app, we need pass `-e PORTAL_URL='https://hostname'`. 
e.g. `-e PORTAL_URL='https://access.devgssci.devlab.phx1.redhat.com'` for portal dev environment


pantheon-bundle/frontend/src/app/moduleDisplay.tsx
```diff
@@ -226,6 +230,29 @@ class ModuleDisplay extends Component<any, any, any> {
...
+                portal = 'https://access.devgssci.devlab.phx1.redhat.com';
+                break;
+            case 'pantheon2-qa':
+                portal = 'https://access.qa.redhat.com';
+                break;
+            case 'pantheon2-stage':
+                portal = 'https://access.stage.redhat.com';
+                break;
+            default:
+                portal = 'https://access.redhat.com';
```
carlosmunoz: I feel this should come from a configuration value and not hard coded into the front-end.

xdavidson: Yes, ideally we should have it configured somewhere. Let me try to do that.

jmarrero: You can try using something like what we do for git2pantheon.

1. set the env variables to the container (internal infra stuff)
2. have the backend read the ENV and save it to /conf/pantheon in the JCR
3. Have the frontend read it from the backend if available.

Code is in: PantheonRepositoryInitializer.java & gitImport.tsx

Look for SYNC_SERVICE_URL in the backend and /conf/pantheon/pant:syncServiceUrl in the frontend code.

PR #180, opened 2019-12-12
add view_uri to modules in portal

xdavidson:
This PR introduces two changes
* replace "solr_command: index" with "event:publish" in  messages to Hydra broker
* add "view_uri" in modules api for search integration.


pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/ModuleJsonServlet.java
```diff
@@ -137,6 +137,14 @@ protected boolean isValidResource(@Nonnull Resource resource) {
             moduleMap.put("vanity_url_fragment", urlFragment);
         }
 
+        // Process view_uri
+        if (System.getenv("PORTAL_URL") != null) {
+            //@TODO: map Pantheon locale to view_uri locale
+            String view_uri_locale = "en-us";
```
carlosmunoz: If you look at PR 181, there's already a 'locale' field in the map, which seems to be a duplicate of this.

PR #182, opened 2019-12-12
bug fix: anonymous module rendering is not working

carlosmunoz:
This change fixes a bug where anonymous users may not generate a module's preview. Since adding automatic metadata extraction from the asciidoc content, the process which writes the metadata to the JCR tree was not given the right permissions. This change ensures that the whole asciidoc generation process uses a service level resource resolver and thus ensure the appropriate write access.

### DIFF
--- a/pantheon-bundle/src/main/java/com/redhat/pantheon/sling/ServiceResourceResolverProvider.java
+++ b/pantheon-bundle/src/main/java/com/redhat/pantheon/sling/ServiceResourceResolverProvider.java
@@ -1,5 +1,6 @@
 package com.redhat.pantheon.sling;
 
+import com.redhat.pantheon.model.api.SlingModel;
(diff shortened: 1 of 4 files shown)


pantheon-bundle/src/main/java/com/redhat/pantheon/sling/ServiceResourceResolverProvider.java
```diff
@@ -1,5 +1,6 @@
 package com.redhat.pantheon.sling;
 
+import com.redhat.pantheon.model.api.SlingModel;
```
jmarrero: Why is this new import needed? the only change in this class other than the import are comments.

carlosmunoz: You are correct sir :smile:

PR #184, opened 2019-12-16
Symlinks support

benradey:
Pantheon 2 and the accompanying uploader now both understand what symlinks are.

So if you have asciidoc with a line like this:
include::recursiveSymlink/recursiveSymlink/recursiveSymlink/include.adoc

...and a directory structure like this:
$ ls -l
recursiveSymlink -> .
include.adoc
master.adoc

...then you can upload via a simple "pantheon push" and both the upload and the preview will complete successfully.
(Previously, the uploader would have churned forever, uploading endless duplicate copies of files, and Pantheon 2 would have ultimately run out of disk space. This no longer happens.)

There is a change in uploader behavior as a result of this PR.
1. You no longer specify 'resources' in your repository. Everything that isn't a module is now a resource and is uploaded as such, to ensure that all symlink-based includes resolve correctly.
2. Removed the option to specify 'titles' because it was a meaningless option that caused confusion.

Note: don't forget to install the new version of pantheon.py before testing this PR.

### DIFF
--- a/uploader/pantheon.py
+++ b/uploader/pantheon.py
@@ -2,9 +2,9 @@
 import argparse
 import base64
 import getpass
-import glob
 import logging
 import os
+import re
 import sys
 from pathlib import PurePath
 
@@ -12,7 +12,7 @@ import requests
 import yaml
 
 DEFAULT_SERVER = 'http://localhost:8080'
-if "PANTHEON_SERVER" in os.environ:
+if 'PANTHEON_SERVER' in os.environ:
     DEFAULT_REPOSITORY = 'gitImport'
 else:
     DEFAULT_REPOSITORY = getpass.getuser()
@@ -30,14 +30,14 @@ def _generate_data(jcr_primary_type, base_name, path_name, asccidoc_type):
     """
     data = {}
     if jcr_primary_type:
-        data["jcr:primaryType"] = jcr_primary_type
+        data['jcr:primaryType'] = jcr_primary_type
     if base_name:
-        data["jcr:title"] = base_name
-        data["jcr:description"] = base_name
+        data['jcr:title'] = base_name
+        data['jcr:description'] = base_name
     if path_name:
-        data["pant:originalName"] = path_name
+        data['pant:originalName'] = path_name
     if asccidoc_type:
-        data["asciidoc@TypeHint"] = asccidoc_type
+        data['asciidoc@TypeHint'] = asccidoc_type
 
     return data
 
@@ -47,7 +47,7 @@ def _info(message, colored=True):
     Print an info message on the console. Warning messages are cyan
     """
     if colored:
-        print("\033[96m{}\033[00m" .format(message))
+        print('\033[96m{}\033[00m' .format(message))
     else:
         print(message)
 
@@ -57,7 +57,7 @@ def _warn(message, colored=True):
     Print a warning message on the console. Warning messages are yellow
     """
     if colored:
-        print("\033[93m{}\033[00m" .format(message))
+        print('\033[93m{}\033[00m' .format(message))
     else:
         print(message)
 
@@ -67,7 +67,7 @@ def _error(message, colored=True):
     Print an error message on the console. Warning messages are red
     """
     if colored:
-        print("\033[91m{}\033[00m" .format(message))
+        print('\033[91m{}\033[00m' .format(message))
     else:
         print(message)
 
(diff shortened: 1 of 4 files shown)


carlosmunoz:
So my only question at this point is if it’s necessary to include ALL non-module files as resources. While this might not be a problem because documentation files are usually kept isolated from others (e.g source code), I do wonder if we could gain efficiency by only uploading the strictly necessary files. 

If this is not possible, could we consider the inverse? mainly specifying an “exclude” section where all specified files or file patterns are ignored when uploading.

benradey:
@carlosmunoz I do think it is necessary to upload everything as resources by default - here's why.

Let's say you have a project structure like this:
```
$ tree
.
├── modules
│   └── master.adoc
├── other_not_uploaded
│   └── symlink -> ../resources
└── resources
    └── attributes.adoc

4 directories, 2 files
```

And let's say your master.adoc includes the attributes.adoc file, but not via its real relative path of `../resources/attributes.adoc`, but rather via it's symlinked path of `../other_not_uploaded/symlink/attributs.adoc`.

It would be very easy to misconfigure the pantheon2.yml file to only upload the `modules` and `resources` folders - after all, those folders contain all the files you care about, right? But in that scenario, the 'include' statement would fail to resolve.

We _could_ also ask the user to specify symlinks in the yaml, but that is a clunky solution that nobody would be happy with. So, the solution is to upload everything.

I would certainly entertain the possibility of a blacklist. I suppose we could also discuss uploading all _directories_ in a project (including symlinks), but not all _files_ in a project (only files specified in the config would go up). That would be an odd structure with lots of empty nodes in the JCR, but it would also solve the problem.

We could also specify whitelisted file extensions in the yaml, defaulting to .adoc, .asciidoc, .jpg, .png, and .gif. If the authors have needs for others, they can add them, but I would imagine that those defaults would cover most of our needs.

Lots of ways to approach this. Thoughts? Should we take it to CCS?

carlosmunoz:
Ok, I think I understand: what we want to prevent here is the scenario where a resource is being symlinked from a module and either the symlink or the target file itself are not being pushed to pantheon. In this case a user must make sure that both files are being pushed or else the module won’t generate properly. 

If this is correct, I feel like it’s still a bit of a sledge hammer, and I’d advocate for adding the opt-out feature (i.e. the blacklist approach). I’m personally more in favor of high specificity so I’d argue that asking users to include ALL the necessary files is not such a high burden, but I’m happy to go with a different solution.

uploader/pantheon.py
```diff
@@ -206,102 +201,114 @@ def process_file(path, filetype):
...
-            processed_files.append(path)
-            logger.debug('')
-        elif isResource:
+            if target[0] == '/':
+                _error('Absolute symlink paths are unsupported: ' + str(path) + ' -> ' + target)
+            elif not args.dry:
+                symlinkData = {}
+                symlinkData['jcr:primaryType'] = 'nt:unstructured'
+                symlinkData['sling:resourceType'] = 'pant:symlink'
+                symlinkData['pant:target'] = target
```
carlosmunoz: Perhaps we can use sling’s `sling:redirect` Property here. It might yield benefits in the long term.
See: https://sling.apache.org/documentation/bundles/rendering-content-default-get-servlets.html#redirectservlet for more details.

benradey: I already played with `sling:redirect` and didn't like it. Here's a rough overview of the scenarios I tried:

Access /realNode/redirectNode via browser: works, redirects you to /realNode/realNode2
Access /realNode/redirectNode/realNode2 via browser: doesn't work
Get /realNode/redirectNode via resource resolver: "works" in that you get the redirectNode itself, but if you want the target node, you have to get it manually
Get /realNode/redirectNode/realNode2 via resource resolver: doesn't work
(In the above examples, realNode2 is a child of the target node)

So given the fact that `sling:redirect` offered almost none of the behavior that actually mattered for us, I felt it was best to just ignore it and write our own solution from scratch rather than to wedge in something that isn't a good fit to start with and might bring some unintended side effects with it.

Additionally, if you have a "link" node and a "target" node that are in the *same directory* and you want to use a `sling:redirect` to point from the 1st to the 2nd, you don't specify "target" as the target, you specify "../target" which differs from the posix standard. So even if we wanted to use `sling:redirect`, we'd have to start modifying the target path, and I'd rather not do that unless there's a clear advantage that makes it worth the effort.

benradey:
@carlosmunoz I have changed the uploader logic so that resource files must now still be specified in the config.

However, the uploader will still scan the entire project for symlinks, and upload all of them.

I feel that this is the best tradeoff between the ability to reliably build documents and not accidentally exploding the repository with irrelevant data (source code and whatnot).

carlosmunoz:
Ok, sounds good, and I see Ivan already approved. 👍

PR #186, opened 2019-12-20
Metadata now copies from release to draft

benradey:


pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/ModuleVersionUpload.java
```diff
@@ -57,6 +61,16 @@
     private static final Logger log = LoggerFactory.getLogger(ModuleVersionUpload.class);
 
     private AsciidoctorService asciidoctorService;
+    private Set<String> excludes = Collections.unmodifiableSet(
```
carlosmunoz: Sorry, very late to the party 😞 ... but when/if you get a chance, you should make this static and final.

PR #194, opened 2020-01-07
Fcceud 22

emarcusRH:
redoing the commit as requested by Ben R


benradey:
Thanks Eli, I was actually able to work around the issue we had with your original PR and your changes should be in master now. Please check and let me know if that is NOT the case. In the meantime, I'll close this PR which we don't need. Thank you though!

PR #196, opened 2020-01-08
Trivial extraction of constants

benradey:
Spent awhile trying to diagnose a phantom issue. Turns out it was just a corrupt `sling/` folder. Regardless, this tweak I made along the way makes the project 0.1% better, so why not merge it?


anknite:
As this is just extraction LGT be approved.

PR #200, opened 2020-01-13
Split products

benradey:
Getting around our "200 maximum json result" limit for the product & productVersion dropdowns in module metadata.


benradey:
Need to update test classes, but functionality works.

PR #201, opened 2020-01-14
Update git2pantheon to read user & pass from environment

jmarrero:
Had to move 	flag.Parse() because of:
https://golang.org/doc/go1.13#testing


jmarrero:
> go test -cover 
Shows 79.5% coverage for the go code. For some reason codecov is not reading the test coverage for go... notice the "?" for missing data in the report.

PR #209, opened 2020-01-16
Ccs-3262

ankkit24:
Changes in the search page as per the updated wireframe.

### DIFF
--- a/pantheon-bundle/frontend/src/app/search.test.tsx
+++ b/pantheon-bundle/frontend/src/app/search.test.tsx
@@ -21,185 +21,185 @@ describe('Search tests', () => {
     expect(button.exists()).toBe(true)
   })
 
-  // it('should render a DataList', () => {
-  //   const wrapper = mount(<Router><Search {...mockStateUser} /></Router>)
-  //   const dataList = wrapper.find(DataList)
-  //   expect(dataList.exists()).toBe(true)
-  // })
+  it('should render a DataList', () => {
+    const wrapper = mount(<Router><Search {...mockStateUser} /></Router>)
+    const dataList = wrapper.find(DataList)
+    expect(dataList.exists()).toBe(true)
+  })
 
-  // it('should render a TextInput', () => {
-  //   const wrapper = mount(<Router><Search {...mockStateUser} /></Router>)
-  //   const textInput = wrapper.find(TextInput)
-  //   expect(textInput.exists()).toBe(true)
-  // })
+  it('should render a TextInput', () => {
+    const wrapper = mount(<Router><Search {...mockStateUser} /></Router>)
+    const textInput = wrapper.find(TextInput)
+    expect(textInput.exists()).toBe(true)
+  })
 
-  // it('should render a BuildInfo component', () => {
-  //   const wrapper = mount(<Router><Search {...mockStateUser} /></Router>)
-  //   const buildInfo = wrapper.find(BuildInfo)
-  //   expect(buildInfo.exists()).toBe(true)
-  // })
+  it('should render a BuildInfo component', () => {
+    const wrapper = mount(<Router><Search {...mockStateUser} /></Router>)
+    const buildInfo = wrapper.find(BuildInfo)
+    expect(buildInfo.exists()).toBe(true)
+  })
 
-  // it('should render a level component', () => {
-  //   const wrapper = mount(<Router><Search {...mockStateUser} /></Router>)
-  //   const level = wrapper.find(Level)
-  //   expect(level.exists()).toBe(true)
-  // })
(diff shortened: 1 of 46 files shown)


pantheon-bundle/frontend/src/app/search.tsx
```diff
@@ -158,6 +156,9 @@ class Search extends Component<IAppState, ISearchState> {
                   </Level>
 
                 )}
+                
+                {console.log("res ui: ",this.state.results)}
```
benradey: Unless this is still necessary for some reason, can you delete this line?

pantheon-bundle/frontend/src/app/search.tsx
```diff
@@ -170,27 +171,23 @@ class Search extends Component<IAppState, ISearchState> {
...
+                      />}                                   
                     <DataListItemCells key={"cells_" + key}
                       dataListCells={[
                         <DataListCell key={"title_" + key} width={2}>
                           {this.props.userAuthenticated &&
                             <Link to={data['pant:transientPath']} key={"link_" + key}>{data["jcr:title"]}</Link>}
                           {!this.props.userAuthenticated &&
                             <a href={"/" + data['pant:transientPath'] + ".preview"} target="_blank">{data["jcr:title"]}</a>}
+                        </DataListCell>,      
+                        <DataListCell key={"transient-source_" + key}>
```
benradey: This isn't a BIG deal, but since I asked for other changes anyway, can you make sure that the `key` property is not only uniquely named, but also has something to do with its purpose? So in this case, change it from "transient-source_" to "published-date_" or something like that. Same thing for the 'date uploaded' field below. Please update attributes, especially `key` and `id` when appropriate as a general rule.

pantheon-bundle/src/main/java/com/redhat/pantheon/model/module/Metadata.java
```diff
@@ -8,6 +8,7 @@
 
 import javax.inject.Named;
 import java.util.Calendar;
+import java.util.Date;
```
benradey: Again not a big deal here, but can you just revert this file since there are no meaningful changes?

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/PublishDraftVersion.java
```diff
@@ -91,7 +91,7 @@ protected void doRun(SlingHttpServletRequest request, PostResponse response, Lis
             versionToRelease.get()
                     .metadata().getOrCreate()
                     .datePublished().set(Calendar.getInstance());
-            changes.add(Modification.onModified(module.getPath()));
+            changes.add(Modification.onModified(module.getPath()));
```
benradey: Ah, same thing here, if there are no meaningful changes in this file then please revert it.

pantheon-bundle/frontend/src/app/search.tsx
ankkit24: Yes Ben, I will take care of this.
Even I feel these changes are necessary.

pantheon-bundle/src/main/java/com/redhat/pantheon/model/module/Metadata.java
ankkit24: Done!

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/PublishDraftVersion.java
ankkit24: Did the changes here as well.
I don't see any changes apart from an extra space, which I have taken care of.

pantheon-bundle/frontend/src/app/search.tsx
ankkit24: This does not seem necessary now, I have removed it.

pantheon-bundle/frontend/src/app/search.test.tsx
```diff
@@ -21,185 +21,185 @@ describe('Search tests', () => {
...
+  //   const wrapper = renderer.create(<Router><Search {...mockStateUser} /></Router>)
+  //   const inst = wrapper.getInstance()
+  //   expect(inst.changePerPageLimit).toMatchSnapshot()
+  // })
+
+  // it('test buildTransientPathArray function', () => {
+  //   const wrapper = renderer.create(<Router><Search {...mockStateUser} /></Router>)
+  //   const inst = wrapper.getInstance()
+  //   expect(inst.buildTransientPathArray).toMatchSnapshot()
+  // })
```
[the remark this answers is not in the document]
ankkit24: Done.

PR #214, opened 2020-01-23
Module query modifications

jmarrero:
nt:base is not needed anymore after removing the modules.json index. The index was having unintended consequences making the query not return results when we use pant:modules.


jmarrero:
This code is Carlos' I just helped debugging.

benradey:
@carlosmunoz I hesitate to approve these changes because the `nt:base` vs `pant:module` is a very old issue and AFAIK it predates us having any index rules in place. Please see commit f260579dce9884aaccf15c2e0f3baa128af31e6b which IIRC was our very first attempt to troubleshoot this problem. Did that commit have any index rules in place?

benradey:
Well I just answered my own question and by golly it seems like we did.

https://github.com/redhataccess/pantheon/blob/f260579dce9884aaccf15c2e0f3baa128af31e6b/src/main/resources/SLING-INF/oak:index/modules.json

carlosmunoz:
I was going to say that Ben... and I'll take the blame for that one. We need to get indexing right, and that file was just the wrong attempt. It was interfering with regular query execution.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/ModuleListingServlet.java
```diff
@@ -148,10 +163,9 @@ protected String getQuery(SlingHttpServletRequest request) {
         productCondition = "AND (" + StringUtils.join(conditions, " OR ") + ") ";
 
         StringBuilder query = new StringBuilder()
-                .append("SELECT pv.* from [nt:base] AS pv ")
-                .append("INNER JOIN [nt:base] AS product ON ISDESCENDANTNODE(pv, product) ")
-                .append("WHERE pv.[jcr:primaryType] = 'pant:productVersion' ")
-                .append("AND product.[jcr:primaryType] = 'pant:product' ")
+                .append("SELECT pv.* from [pant:productVersion] AS pv ")
+                .append("INNER JOIN [pant:product] AS product ON ISDESCENDANTNODE(pv, product) ")
```
benradey: We still have this inner join - is it worth eliminating? We could probably easily use deep attribute inspection here.

carlosmunoz: This one is trickier to turn into XPATH, as we are actually filtering on a parent node's properties from the child. A single join doesn't seem to be as costly also. In addition, this one is only being executed when the product filters are used, so any additional cost is limited to those scenarios.

benradey: Ooof. Good point. Alright, we can save this one for later/never.

PR #217, opened 2020-01-28
Change draft/released references to named nodes

carlosmunoz:


pantheon-bundle/src/main/java/com/redhat/pantheon/model/module/ModuleLocale.java
```diff
@@ -25,6 +28,61 @@ default ModuleVersion getOrCreateVersion(String name) {
...
+    default ModuleVersion releaseDraftVersion() {
+        if(draft().get() == null) {
+            // nothing to do
+            return null;
+        }
+
+        // Rename 'draft' to 'released'
+        try {
+            // TODO externalize this magic string
+            return SlingModels.rename(draft().get(), "released", ModuleVersion.class);
```
benradey: I'm surprised you don't call `archiveReleasedVersion` defensively prior to executing the rename.

pantheon-bundle/src/main/java/com/redhat/pantheon/model/module/ModuleLocale.java
```diff
@@ -25,6 +28,61 @@ default ModuleVersion getOrCreateVersion(String name) {
...
+        if(released().get() == null) {
+            // nothing to do
+            return null;
+        }
+
+        try {
+            // If there is no 'draft' version already, rename 'released' to 'draft'
+            if(draft().get() == null) {
+                // TODO externalize this magic string
+                return SlingModels.rename(released().get(), "draft", ModuleVersion.class);
```
benradey: This would need to get a little bit more robust once we have assemblies - what if an assembly is referring to this module version, we can't simply turn it into a draft that might get overwritten with another upload... but, can you even unpublish a module that an assembly is currently depending on? Oof - makes my head hurt. This is fine for now.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/PublishDraftVersion.java
```diff
@@ -85,10 +86,10 @@ protected void doRun(SlingHttpServletRequest request, PostResponse response, Lis
         } else {
             // Draft becomes the new released version
             ModuleLocale moduleLocale = module.getModuleLocale(locale);
-            moduleLocale.released().set( moduleLocale.draft().get() );
-            moduleLocale.draft().set( null );
+            moduleLocale.archiveReleasedVersion();
+            ModuleVersion releasedVersion = moduleLocale.releaseDraftVersion();
```
benradey: Yeah, perfect followup from my previous comment - it doesn't make sense for the developer to need to call both 'archive' and 'release' methods - the release process should imply any necessary archiving.

pantheon-bundle/src/test/java/com/redhat/pantheon/servlet/ModuleJsonServletTest.java
```diff
@@ -68,22 +68,19 @@ void getQuery() {
     void resourceToMap() throws Exception {
         // Given
         slingContext.create()
-                .resource("/content/repositories/repo/module/en_US/1",
+                .resource("/content/repositories/repo/module/en_US/released",
```
benradey: Not sure these test updates are necessary - but since you're making them, it might be worth it to use string constants to avoid confusion between "release" vs "released". I noticed in the previous class that you didn't do it and I neglected to mention anything since it was pretty local, but now that I see this, it gives me pause.

pantheon-bundle/src/main/java/com/redhat/pantheon/model/module/ModuleLocale.java
carlosmunoz: You made me think here... :smiley: 
I guess the approach was to have really atomic operations which could be put together in different ways. The defensive call you mention is being done in the servlet itself.

pantheon-bundle/src/main/java/com/redhat/pantheon/model/module/ModuleLocale.java
carlosmunoz: You are right, we will need to revisit this. All I was trying to do here was replicate whatever functionality the system offers before the change.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/PublishDraftVersion.java
carlosmunoz: We still need the `archiveReleasedVersion` to be separate as we use it in the unpublish operation.
But I could create more "business level" methods which just `publishDraft` and `recallReleased` 
Thoughts on that approach?

pantheon-bundle/src/test/java/com/redhat/pantheon/servlet/ModuleJsonServletTest.java
carlosmunoz: Yeah, these magic strings are all over the place. I have TODOs to externalize the strings. I can do it as part of this PR.

carlosmunoz: Not sure what you mean by not necessary though... every test method which is fetching the draft or released version of a module specifically will need to be updated.

On second thought about the magic strings: I will externalize the ones in the model objects, but not the tests as I think it reduces their readability, specially since they are using these strings as hard-coded prerequisite data.

carlosmunoz:
@xdavidson I merged your branch but the build failed

PR #218, opened 2020-01-28
[CCS-2965] update unpublish event message

xdavidson:
This PR updates the message we send to Hydra for unpublished events.

### DIFF
--- a/pantheon-bundle/src/main/java/com/redhat/pantheon/extension/HydraIntegration.java
+++ b/pantheon-bundle/src/main/java/com/redhat/pantheon/extension/HydraIntegration.java
@@ -117,22 +117,27 @@ public class HydraIntegration implements EventProcessingExtension {
                 + moduleUUID : "";
         String uriValue = "";
         String msg = "";
-        if (System.getenv(PORTAL_URL) != null) {
-            uriValue = System.getenv(PORTAL_URL) + "/topics/" + ServletUtils.toLanguageTag(DEFAULT_MODULE_LOCALE) + "/" + moduleUUID;
-        }
+
         if (ModuleVersionPublishedEvent.class.equals(event.getClass())) {
             msg = "{\""
                     + ID_KEY + "\":" + "\"" + idValue +"\","
                     + "\"" + EVENT_KEY + "\":" + "\"" + eventValue + "\"}";
         } else {
-            msg = "{\""
+            if (System.getenv(PORTAL_URL) != null) {
+                uriValue = System.getenv(PORTAL_URL) + "/topics/" + ServletUtils.toLanguageTag(DEFAULT_MODULE_LOCALE) + "/" + moduleUUID;
+
+                msg = "{\""
                     + ID_KEY + "\":" + "\"" + idValue +"\","
                     + "\"" + EVENT_KEY + "\":" + "\"" + eventValue + "\","
                     + "\"" + URI_KEY + "\":" + "\"" + uriValue + "\"}";
+            }
         }
-
-        producer.send(session.createTextMessage(msg));
-        log.info("[" + HydraIntegration.class.getSimpleName() + "] message sent: " + session.createTextMessage(msg) );
+        if (!msg.isEmpty()) {
+            producer.send(session.createTextMessage(msg));
+            log.info("[" + HydraIntegration.class.getSimpleName() + "] message sent: " + session.createTextMessage(msg) );
+        } else {
+            log.info("[" + HydraIntegration.class.getSimpleName() + "] empty message!");
+        }
 
         connection.close();
     }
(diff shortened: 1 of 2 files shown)


pantheon-bundle/src/main/java/com/redhat/pantheon/extension/HydraIntegration.java
```diff
@@ -104,11 +113,26 @@ public void processEvent(Event event) throws Exception {
...
         String moduleUUID = module.getValueMap().get(UUID_FIELD, String.class);
         String eventValue = ModuleVersionPublishedEvent.class.equals(event.getClass()) ? EVENT_PUBLISH_VALUE : EVENT_UNPUBLISH_VALUE;
-        String msg = "{\""
-                + ID_KEY + "\":" + "\"" + this.getPantheonHost() + PANTHEON_MODULE_API_PATH + moduleUUID +"\","
-                + "\"" + EVENT_KEY + "\":" + "\"" + eventValue + "\"}";
+        String idValue = ModuleVersionPublishedEvent.class.equals(event.getClass()) ? this.getPantheonHost() + PANTHEON_MODULE_API_PATH
+                + moduleUUID : "";
+        String uriValue = "";
+        String msg = "";
+        if (System.getenv(PORTAL_URL) != null) {
```
carlosmunoz: what  happens if the portal URL is not defined? seems like it would still seek to publish an empty id.
Should we abort the message in that case?

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/ModuleJsonServlet.java
```diff
@@ -158,12 +158,6 @@ protected boolean isValidResource(@Nonnull SlingHttpServletRequest request, @Non
             moduleMap.put(VIEW_URI, view_uri);
         }
 
-        // Process view_uri
-        if (System.getenv("PORTAL_URL") != null) {
```
carlosmunoz: So, this will NOT be appearing in the API anymore?
Just confirming...

pantheon-bundle/src/main/java/com/redhat/pantheon/extension/HydraIntegration.java
```diff
@@ -23,19 +22,27 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.redhat.pantheon.extension.events.ModuleVersionPublishStateEvent;
 import com.redhat.pantheon.extension.events.ModuleVersionPublishedEvent;
```
carlosmunoz: I'm having a bit of trouble following these event types. Could you explain what the difference is between `ModuleVersionPublishStateEvent` and `ModuleVerisonPublishedEvent`? they seem to be the same thing

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/ModuleJsonServlet.java
xdavidson: line 161 is a duplicated code block.

pantheon-bundle/src/main/java/com/redhat/pantheon/extension/HydraIntegration.java
xdavidson: line 25 is a result of Eclipse reorganizing its import statements. It used to be on line 15.
I think ModuleVersionPublishStateEvent can be a ModuleVersionPublishedEvent or ModuleVersionUnpublishedEvent

pantheon-bundle/src/main/java/com/redhat/pantheon/extension/HydraIntegration.java
```diff
@@ -104,11 +113,26 @@ public void processEvent(Event event) throws Exception {
...
+        String uriValue = "";
+        String msg = "";
+        if (System.getenv(PORTAL_URL) != null) {
+            uriValue = System.getenv(PORTAL_URL) + "/topics/" + ServletUtils.toLanguageTag(DEFAULT_MODULE_LOCALE) + "/" + moduleUUID;
+        }
+        if (ModuleVersionPublishedEvent.class.equals(event.getClass())) {
+            msg = "{\""
+                    + ID_KEY + "\":" + "\"" + idValue +"\","
+                    + "\"" + EVENT_KEY + "\":" + "\"" + eventValue + "\"}";
+        } else {
```
carlosmunoz: Ok, I think I understand now.... in this case I would turn this into an `else if` for the specific implementation class, and have the `else` statement log some sort of warning. It will help us identify when we have an event type that is not being handled by the code.

pantheon-bundle/src/main/java/com/redhat/pantheon/extension/HydraIntegration.java
xdavidson: good catch. i've refactored the code.

PR #222, opened 2020-01-29
fix pager issue

xdavidson:
This PR fixes the pagination issue on search page


pantheon-bundle/frontend/src/app/search.tsx
```diff
@@ -421,7 +421,10 @@ class Search extends Component<IAppState, ISearchState> {
   private buildSearchUrl() {
     let backend = "/modules.json?"
     backend += this.state.filterQuery
-    backend += "&offset=" + ((this.state.page - 1) * this.state.pageLimit) + "&limit=" + this.state.pageLimit
+    if (this.state.filterQuery.trim() !== "") {
```
carlosmunoz: I prefer `this.state.filterQuery.trim().length > 0` but I won't hold the PR for this.

PR #230, opened 2020-02-12
FCCEUD-20 Describe the pantheon2.yml file

stoobie:
Description of this file also impacts the topic on the uploader tool, so I made edits to that too. And since that topic is a procedure, not a reference, I renamed the file to reflect this.


pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/con_pantheon_yaml_file.adoc
```diff
@@ -0,0 +1,45 @@
...
+----
+# Config file for Pantheon
+## server: Pantheon server URL
+## repository: a unique name, which is visible in the user facing URL
+
+## Note: Due to yaml syntax, any filepaths that start with a wildcard must be surrounded in quotes like so:
+# modules:
+#  - '*.adoc'
+
+server: http://pantheon2.int.us-west.aws.prod.paas.redhat.com[]
```
adahms: The current URL for Pantheon is internal only, so we should probably change this and all other references to a replaceable such as '<server_url>' for the time being.

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/proc_uploader_tool.adoc
```diff
@@ -0,0 +1,66 @@
...
+[options="nowrap" subs="normal"]
+----
+usage: pantheon [-h] [--server SERVER] [--repository REPOSITORY] [--user USER]
+[--password PASSWORD] [--directory DIRECTORY] [--verbose]
+[--dry] [--sandbox] [--sample]
+push [push ...]
+pantheon: error: the following arguments are required: push
+----
+
+.To configure the repository as a content source for Pantheon:
```
adahms: Should this and the procedure below be separate procedure modules?

Happy to change this into an assembly with three procedures if that's easier.

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/con_pantheon_yaml_file.adoc
```diff
@@ -0,0 +1,45 @@
+[id='con-pantheon-yaml-file_{context}']
+
+= The Pantheon configuration file
+
```
theashiot: Perhaps an introductory statement about what is a pantheon configuration file here?

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/con_pantheon_yaml_file.adoc
```diff
@@ -0,0 +1,45 @@
...
+The configuration file enables the following:
+
+* Importing content from a Git repository into Pantheon
+* Uploading modules into Pantheon using the uploader tool
+
+The configuration file defines the following:
+
+`server`:: The URL of the Pantheon server.
+`repository`:: A unique ID to reference the URL of the repository. You specify the actual URL of the repository when you synchronize the repository the first time.
```
theashiot: I did not understand the difference between a unique ID and the actual URL.

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/con_pantheon_yaml_file.adoc
```diff
@@ -0,0 +1,45 @@
...
+`modules`:: Patterns for modules that will show up in the search results on Pantheon.
+`resources`:: Image files, legal information or other content that is not.
```
theashiot: Incomplete sentence!

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/con_pantheon_yaml_file.adoc
theashiot: @adahms I think this is a good candidate for an attribute. We can define it as :ProductUrl: (or some such) in the attributes file.

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/proc_uploader_tool.adoc
```diff
@@ -0,0 +1,66 @@
+[id="uploader_tool"]
+
+= Uploading content with the Uploader tool
+
+Many of the operations available in the Pantheon user interface can also be performed using commands in the command line interface.
```
theashiot: I suggest not referring to the Pantheon user interface in this module. I think "The pantheon uploader tool provides many Pantheon operations in the CLI" (something on these lines) should suffice.

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/proc_uploader_tool.adoc
theashiot: Agree with @adahms that these should be different procedures:

- Installing and configuring the uploader 
- Uploading content using the uploader

However, I don't think an assembly is required at this point.

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/con_pantheon_yaml_file.adoc
```diff
@@ -0,0 +1,45 @@
+[id='con-pantheon-yaml-file_{context}']
```
theashiot: s / con-pantheon-yaml-file_{context} / pantheon-yaml-file_{context}

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/proc_uploader_tool.adoc
```diff
@@ -0,0 +1,66 @@
+[id="uploader_tool"]
```
theashiot: s / uploader_tool / uploading-content-with-the-uploader-tool_{context}

theashiot:
Hi @stoobie,

I have added a few suggestions inline. 

A global comment I have is to follow the "<module-type>_module-heading.adoc" file naming convention. Similarly, the module id should be id='module-heading_{context}'.

Hope the comments make sense!

best,
ashwin

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/con_pantheon_yaml_file.adoc
stoobie: Done.

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/con_pantheon_yaml_file.adoc
stoobie: I thought that's what the first sentence is. What do you suggest?

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/con_pantheon_yaml_file.adoc
stoobie: Done.

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/con_pantheon_yaml_file.adoc
stoobie: Yeah, I think an example would help. @benradey Can you please provide an example of a unique ID vs a URL for a repo, and how the unique ID and repo are associated with each other in a command-line?

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/con_pantheon_yaml_file.adoc
stoobie: Good catch. I guess you don't like to play guessing games, huh?

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/proc_uploader_tool.adoc
stoobie: Done

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/proc_uploader_tool.adoc
stoobie: Fixed.

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/proc_uploader_tool.adoc
stoobie: @theashiot @adahms I agree, that's what I suggested at first for this task. Configuration here is not just about the uploader tool, it's about configuring your repo to import content into Pantheon. Now that I've written a draft, including the example file, I suggest having the following modules:

1. `con_pantheon_yaml_file.adoc`: The Pantheon configuration file
1. `proc_config_repo.adoc`: Configuring a repo as a content source for Pantheon
1. `proc_installing_uploader.adoc`: Installing the Uploader tool
1. `proc_uploader_tool.adoc`:  Uploading content using the Uploader tool
1. `proc_importing_content.adoc`: Importing content from a Git repository

stoobie: I've added:
* proc_config_repo.adoc
* proc_installing_uploader.adoc

I've edited con_pantheon_yaml_file.adoc and proc_uploader_tool.adoc.

I don't have the info for importing content from a Git repo.

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/con_pantheon_yaml_file.adoc
benradey: Sure.

The `repository` field is supposed to be a unique human-readable name that currently serves only one purpose - to ensure that the mapping between 'git repository' and 'pantheon repository' is persistent. (Otherwise, Pantheon currently has no memory of a repository that has been previously synced, vs. a brand-new repository being synced for the first time.) In the future, the repository name might also be used to achieve cross-repository module includes for assemblies, but details here haven't been ironed out yet.

The `server` field is used mostly for debugging right now - it's only used when invoking the pantheon uploader from the command line (keep in mind that command line usage is unsupported for MS2) so that the uploader knows where to attempt to sync the files to. What invoking a git sync from the Pantheon UI, the `server` field is ignored because the location of the Pantheon instance is obviously known.

Does that help?

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/con_pantheon_yaml_file.adoc
theashiot: Something on the lines of "A Pantheon configuration file is a yaml file containing configuration information required by Pantheon for publishing modules".

The reason for this suggestion is that when a user read the title "The configuration file", the reader would expect the following text to say what is the configuration file. The first sentence, however, is about using a Git repository to import content into pantheon. Reversing the clauses, such as "You need to create a configuration file, named `pantheon2.yml` in the root directory .." should suffice though. Although, I'm not a big fan of having instructions in a conceptual topic - but that's just my thing!

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/proc_config_repo.adoc
```diff
@@ -0,0 +1,29 @@
+[id="config_repo_{context}"]
```
theashiot: s " "config_repo_{context}" "/" "configuring-a-repository-as-a-content-source-for-pantheon_{context}"

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/con_pantheon_yaml_file.adoc
stoobie: I made the change you suggested. Please take a look at my changes:
> You need to create a configuration file, named `pantheon2.yml`, in the root directory of your repository, to enable a Git repository as a source that Pantheon can use to import content. This yaml file provides necessary information that Pantheon uses to interact with files in the repository.

I understand what you're saying about not including instructions in a conceptual file. I don't really see these as instructions, per se. It's more a statement that this is something that is required, and it requires you to take action to do it. It doesn't tell you how to do it.

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/con_pantheon_yaml_file.adoc
stoobie: > keep in mind that command line usage is unsupported for MS2

Will it be supported in the future? In other words, should I document it or not?

benradey: It will be supported in the future for uploading to personal sandboxes, which is a feature that does not exist today.

It might be helpful to document the usage of the tool for troubleshooting purposes. Users can't run the tool against our production instance of Pantheon, but they could run it against our lower environments. Perhaps somebody is trying to troubleshoot their module/resource wildcard patterns, and is running the tool against our stage environment to make sure that the correct set of files gets uploaded.

So bottom line, I believe that having it documented would be helpful.

stoobie: I think that for it to be useful in the way you describe, we should have a separate topic on Troubleshooting, where we say specifically how to do what you're referring to. Otherwise, I don't see it as helpful. Also, who would be doing the troubleshooting you're talking about? CCS? DPX? @adahms I'd like your input here.

adahms: Thanks for the ping, @stoobie, and my apologies for the delay in getting back to you.

I think it's useful to have the content and procedures today, even if we don't necessarily include them in the official help section in the software right away. We can always expand on or refine them in the future as the need arises, and since it's a piece of functionality that does already exist, the content you've created can help general users understand what it's all about.

As for a troubleshooting procedure, I'm a little worried that could expand the scope of this particular issue beyond its original boundaries, so perhaps that's something we can raise a ticket for in the future when it's needed.

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/con_pantheon_yaml_file.adoc
adahms: @stoobie - we won't have a separate downstream and upstream version right away, but I agree that using an attribute as suggested could be a good approach for now.

Would you be able to add the attribute in the attributes file for this, and make the value a replaceable for now?

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/con_pantheon_yaml_file.adoc
```diff
@@ -0,0 +1,47 @@
...
+
+modules:
+ - master.adoc
+ - modules/*.adoc
+
+resources:
+ - shared/legal.adoc
+ - shared/foreword.adoc
+ - resources/*
+''')
```
benradey: I think including the `''')` here is a mistake.

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/proc_config_repo.adoc
```diff
@@ -0,0 +1,29 @@
+[id="configuring-a-repository-as-a-content-source-for-pantheon_{context}"]
+
+= Configuring a repository as a content source for Pantheon
+
+.Procedure
+
+. In a text editor, paste the content of the sample configuration file in xref:pantheon-yaml-file_{context}[] into a new text file. Name the file [filename]`pantheon2.yml`. This file is the Pantheon configuration file for your repository.
+
+. Replace the default value for ‘server’ with the URL you receive from a Pantheon administrator. and ‘repository’ with U:
```
benradey: `and ‘repository’ with U` seems like this is a text artifact that should be removed?

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/proc_uploader_tool.adoc
```diff
@@ -0,0 +1,33 @@
+[id="uploading-content-with-the-uploader-tool_{context}"]
```
benradey: I think this discussion came up in another EUD PR, but ultimately, users will not have much use for the uploader tool until we implement sandboxes in Pantheon. At that point, users can follow this procedure to upload their docs to their personal sandbox.

I call this out because currently, this text as-written implies that that the uploader will be used to sync repository content, which is not the case.

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/con_pantheon_yaml_file.adoc
stoobie: @benradey I copied this directly from the sample file. So if this is a bug, then I guess you have to fix it in the sample file too.

stoobie: Is the request here to delete line 46?

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/proc_config_repo.adoc
stoobie: Yep. I missed that.

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/proc_uploader_tool.adoc
stoobie: OK. I'll take a look. Maybe we need to just remove this topic.

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/con_pantheon_yaml_file.adoc
benradey: Yes, it looks like this was copy/pasted from the python source, and line 46 with the triple-quote is the signal to python that that's where the content ends.

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/proc_config_repo.adoc
```diff
@@ -0,0 +1,29 @@
+[id="configuring-a-repository-as-a-content-source-for-pantheon_{context}"]
+
+= Configuring a repository as a content source for Pantheon
+
+.Procedure
+
+. In a text editor, paste the content of the sample configuration file in xref:pantheon-yaml-file_{context}[] into a new text file. Name the file [filename]`pantheon2.yml`. This file is the Pantheon configuration file for your repository.
```
theashiot: I suggest: 
s / 'xref:pantheon-yaml-file_{context}[]' / 'xref:pantheon-yaml-file_{context}[pantheon-yaml-file]' or some such, so that the rendered text looks better.

Also, would it not be better to use 'link' instead of 'xref'?

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/proc_config_repo.adoc
```diff
@@ -0,0 +1,29 @@
...
+server: http://pantheon.example.com[]
+repository: <UNIQUE_ID>
+----
+. Replace the default value for ‘repository’ with a unique ID, not a URL:
++
+[options="nowrap" subs="normal"]
+----
+repository: <UNIQUE_ID>
+----
+[NOTE]
```
theashiot: Require + above [NOTE]

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/proc_installing_uploader_tool.adoc
```diff
@@ -0,0 +1,34 @@
...
+
+= Installing the Uploader tool
+
+.Procedure
+
+. Download the Uploader script.
++
+[options="nowrap" subs="normal"]
+----
+$ curl -S https://raw.githubusercontent.com/redhataccess/pantheon/master/uploader/setup.sh
```
theashiot: Wouldn't wget be better than curl? curl only displays the file content and there is then an additional step to store the displayed contents in a file.

If using curl, I suggest modifying the step to:

curl -S https://raw.githubusercontent.com/redhataccess/pantheon/master/uploader/setup.sh > setup.sh

so that we don't need another step to store the script.

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/con_pantheon_yaml_file.adoc
theashiot: Agreed.

theashiot:
Hi @stoobie,

I've added a few suggestions. A global suggestion is to add an introduction in each module procedure module. See: https://redhat-documentation.github.io/modular-docs/#creating-procedure-modules 

best,
ashwin

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/con_pantheon_yaml_file.adoc
stoobie: Done:

:ProductUrl: http://pantheon2.int.us-west.aws.prod.paas.redhat.com[]

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/con_pantheon_yaml_file.adoc
stoobie: I removed  `''')`

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/proc_config_repo.adoc
stoobie: Fixed.

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/proc_uploader_tool.adoc
stoobie: As per @adahms in another comment, we'll keep the module since it's already authored, but we don't have to include it when we publish the docs.

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/proc_config_repo.adoc
stoobie: @theashiot Why is it better to use link instead of xref?

And why would the rendered text look better the way you suggest?

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/proc_config_repo.adoc
stoobie: Good catch. Fixed.

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/proc_installing_uploader_tool.adoc
stoobie: @benradey I refer @theashiot's question about curl vs wget to you. For now, I'm taking the suggestion and changing curl to wget.

benradey: The intention is to pipe the script to bash, not redirect to a file. But it works either way. I will admit that piping to bash is generally bad practice, but I would consider us a trustworthy source, IMHO.

PR #233, opened 2020-02-13
CCS-3270: File name not displayed instead of the title in search results on uploading the module twice

ankkit24:
On uploading the adoc files multiple times, the module title remains the same now.

### DIFF
--- a/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/ModuleVersionUpload.java
+++ b/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/ModuleVersionUpload.java
@@ -161,7 +161,10 @@ public class ModuleVersionUpload extends AbstractPostOperation {
 
             Metadata metadata = draftVersion.get()
                     .metadata().getOrCreate();
-            metadata.title().set(moduleName);
+            
+            if(metadata.title().get()==null){
+                metadata.title().set(moduleName);
+            }                    
             metadata.description().set(description);
             Calendar now = Calendar.getInstance();
             metadata.dateModified().set(now);


pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/ModuleVersionUpload.java
```diff
@@ -161,7 +161,10 @@ protected void doRun(SlingHttpServletRequest request, PostResponse response, Lis
 
             Metadata metadata = draftVersion.get()
                     .metadata().getOrCreate();
-            metadata.title().set(moduleName);
+            
+            if(metadata.title().get()==null){
+                metadata.title().set(moduleName);
+            }
```
jmarrero: The problem with this approach is that I don't think it takes in consideration when the title name changes.
So if module version2 has a different title I don't think this will update the title metadata right?

carlosmunoz: Joseph is correct here, the title attribute comes from the content itself and gets populated in the html generation phase (line 174). I think this section should be moved to line 185 so that the default title only gets populated IFF there is no module title. See line 180-184 for a similar approach with a different property.

jmarrero: Ankit tested this, and showed it to me, It looks like my assumption is wrong this works because when we upload the second version it is creating a second metadata object if the file is new. If not then it does a get. This is thanks to the getorcreate magic.

PR #235, opened 2020-02-19
Ccs 3384 backend fixes2

xdavidson:
This PR injects metadata values to the header section of the html

### DIFF
--- a/pantheon-bundle/src/main/java/com/redhat/pantheon/asciidoctor/AsciidoctorService.java
+++ b/pantheon-bundle/src/main/java/com/redhat/pantheon/asciidoctor/AsciidoctorService.java
@@ -166,27 +166,18 @@ public class AsciidoctorService {
             if (productVersion != null) {
                 productName = productVersion.getProduct().getValueMap().get("name").toString();
             }
-            log.info("[" + AsciidoctorService.class.getSimpleName() + "] productName: " + productName);
-            log.info("[" + AsciidoctorService.class.getSimpleName() + "] versionName: " + productVersion.getValueMap().get("name"));
             Calendar updatedDate = moduleVersion.metadata()
                     .map(Metadata::dateUploaded)
                     .map(Supplier::get)
                     .get();
 
             Calendar publishedDate = null;
-//            log.info("[" + AsciidoctorService.class.getSimpleName() + "] datePublished: " + moduleVersion.metadata().map(Metadata::datePublished));
-//            if (moduleVersion.metadata()
-//                    .map(Metadata::datePublished)
-//                    .map(Supplier::get)
-//                    .get() != null) {
-//                publishedDate = moduleVersion.metadata()
-//                          .map(Metadata::datePublished)
-//                          .map(Supplier::get)
-//                          .get();
-//            }
+            publishedDate = moduleVersion.metadata()
+                    .map(Metadata::datePublished)
+                    .map(Supplier::get)
+                    .get();
             
             SimpleDateFormat dateFormat = new SimpleDateFormat("dd MMMMM yyyy");
-            log.info("[" + AsciidoctorService.class.getSimpleName() + "] updatedDate: " + dateFormat.format(updatedDate.getTime()));
             // build the attributes (default + those coming from http parameters)
             AttributesBuilder atts = AttributesBuilder.attributes()
                     // show the title on the generated html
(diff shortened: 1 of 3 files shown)


pantheon-bundle/src/main/java/com/redhat/pantheon/asciidoctor/AsciidoctorService.java
```diff
@@ -27,10 +25,21 @@
...
+import com.redhat.pantheon.asciidoctor.extension.MetadataExtractorTreeProcessor;
+import com.redhat.pantheon.asciidoctor.extension.SlingResourceIncludeProcessor;
+import com.redhat.pantheon.conf.GlobalConfig;
+import com.redhat.pantheon.extension.HydraIntegration;
+import com.redhat.pantheon.model.ProductVersion;
+import com.redhat.pantheon.model.module.Content;
+import com.redhat.pantheon.model.module.Metadata;
+import com.redhat.pantheon.model.module.Module;
+import com.redhat.pantheon.model.module.ModuleVersion;
+import com.redhat.pantheon.sling.ServiceResourceResolverProvider;
```
carlosmunoz: massive import order change :disappointed: 
We should align all of our editors... somehow, as these changes obfuscate things

pantheon-bundle/src/main/java/com/redhat/pantheon/asciidoctor/AsciidoctorService.java
```diff
@@ -142,25 +151,55 @@ private String buildModule(Module base, ModuleVersion moduleVersion, Map<String,
...
+                        try {
+                            return t.getReference();
+                        } catch (RepositoryException e) {
+                            return null;
+                        }
+                    })
+                    .get();
+            String productName = null;
+            if (productVersion != null) {
+                productName = productVersion.getProduct().getValueMap().get("name").toString();
```
carlosmunoz: you should use the new method here
```suggestion
                productName = productVersion.getProduct().name().get();
```

pantheon-bundle/src/main/java/com/redhat/pantheon/asciidoctor/AsciidoctorService.java
xdavidson: yes! I need start using IntelliJ.

pantheon-bundle/src/main/java/com/redhat/pantheon/asciidoctor/AsciidoctorService.java
```diff
@@ -153,30 +151,38 @@ private String buildModule(Module base, ModuleVersion moduleVersion, Map<String,
...
+                        .get();
+            }
+
             String productName = null;
             if (productVersion != null) {
-                productName = productVersion.getProduct().getValueMap().get("name").toString();
+                productName = productVersion.getProduct().name().get();
+            }
+            Calendar updatedDate = null;
+            if (moduleVersion.metadata().get().getValueMap().containsKey("dateUploaded")) {
```
carlosmunoz: I think this can be simplified. The `map` calls are supposed to provide null-safety, but from your changes I suspect they aren’t at the end. The last `get()` call will throw a `NoSuchElementException` if the updated date is not present, so I suggest the following:

```suggestion
            Optional<Calendar> updatedDate = moduleVersion.metadata()
                    .map(Metadata::dataUploaded)
                    .map(Supplier::get);
```
And then in line 179 you should be able to do:
```java
            if(updatedDate.isPresent()) {
                arts.attribute(“pantheonupdateddate”,  dateFormat.format(updatedDate.get().getTime()));
            }
```

pantheon-bundle/src/main/java/com/redhat/pantheon/asciidoctor/AsciidoctorService.java
xdavidson: done

pantheon-bundle/src/test/java/com/redhat/pantheon/asciidoctor/AsciidoctorServiceTest.java
```diff
@@ -75,16 +76,16 @@ public void testGetModuleHtmlWhenCached() throws Exception {
 
         // Given
         slingContext.build()
-                .resource("/module/locales/en_US/released/metadata")
-                .resource("/module/locales/en_US/released/content/asciidoc/jcr:content",
+                .resource("/module/en_US/released/metadata")
+                .resource("/module/en_US/released/content/asciidoc/jcr:content",
```
carlosmunoz: :+1: I guess this was never changed, but I wonder why it's presenting problems only now :thinking:

xdavidson: Line 60 failed on NoSuchElementException. Line 79-84 didn't fail. I changed it so that they are consistent.

pantheon-bundle/src/main/java/com/redhat/pantheon/asciidoctor/AsciidoctorService.java
xdavidson: updated!

PR #236, opened 2020-02-20
feat: Adding Sass files and front-end build tools to pantheon

wesruv:
Added gulp build process with browser-sync and a sample page
Removed cruft from old rh-styles repo
Made small optimizations for front-end performance


carlosmunoz:
Thanks for this Wes! 
My only feedback right now is the place where you put these files... it means they will end up in the pantheon distribution and hence accessible from a browser. This is not necessarily a problem, but it's not really needed. Correct me if I'm wrong but I believe in the end we only need the resulting css file.

I would look at moving this to either the `frontend` module or it's own separate module folder. Thoughts?

wesruv:
That sounds much better! I figured it'd probably need to be moved, but didn't know where.

Does `pantheon/pantheon-bundle/frontend/src/web-assets` make sense?

I want to avoid being too 'css' or 'Sass' specific in the naming, that front end build tools I'm adding could easily grow to add things like javascript, svgs, or image optimization.

wesruv:
I would make it so the CSS still ends up where it is currently, but move everything else.

carlosmunoz:
That sounds ok to me. @benradey any thoughts?

benradey:
That sounds like a logical starting point, I'm on board with it!

wesruv:
Got everything moved and verified that it compiles, might be good for someone else to try and run it to test it?

To test I deleted the current rhdocs file:
`rm src/main/resources/SLING-INF/content/static/rhdocs.css`

Confirm it's gone with `git status`

Navigate to frontend build folder:
`cd pantheon-bundle/frontend/src/web-assets`

Install deps:
`yarn install`

Build CSS:
`yarn run build`

Run `git status` again and the file should be there (it's minifed so it might still show changes?)

wesruv:
Updated the docs per my updates to the build process, so this is ready to merge!

wesruv:
Err... now it is :D

PR #239, opened 2020-02-26
doc is now rebuilt immediately before sending publish event

benradey:
Steps to test:

1. Upload a doc with content similar to this:
```
Pantheon attributes:

showtitle: {showtitle}

pantheonproduct: {pantheonproduct}

pantheonversion: {pantheonversion}

pantheonupdateddate: {pantheonupdateddate}

pantheonpublisheddate: {pantheonpublisheddate}
```
2. Don't assign metadata
3. Preview the doc
4. Assign all necessary metadata
5. Publish the doc
6. View the rendered doc
Old behavior: variables are missing
New behavior: variables are available

### DIFF
--- a/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/PublishDraftVersion.java
+++ b/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/PublishDraftVersion.java
@@ -1,5 +1,6 @@
 package com.redhat.pantheon.servlet.module;
 
+import com.redhat.pantheon.asciidoctor.AsciidoctorService;
 import com.redhat.pantheon.conf.GlobalConfig;
 import com.redhat.pantheon.extension.Events;
 import com.redhat.pantheon.extension.events.ModuleVersionPublishedEvent;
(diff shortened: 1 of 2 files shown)


pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/PublishDraftVersion.java
```diff
@@ -58,6 +63,14 @@ public void run(SlingHttpServletRequest request, PostResponse response, SlingPos
             Locale locale = getLocale(request);
             Module module = getModule(request);
             ModuleLocale moduleLocale = module.getModuleLocale(locale);
+
+
+            //FIXME - this is a hack that needs to be removed when we have the attribute placeholder logic implemented
+            Optional<ModuleVersion> versionToRelease = getModule(request).getReleasedVersion(getLocale(request));
```
carlosmunoz: seems like we already have fetched some of those values:
```suggestion
            Optional<ModuleVersion> versionToRelease = module.getReleasedVersion(locale);
```

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/PublishDraftVersion.java
```diff
@@ -58,6 +63,14 @@ public void run(SlingHttpServletRequest request, PostResponse response, SlingPos
...
+            Map<String, Object> context = asciidoctorService.buildContextFromRequest(request);
```
carlosmunoz: not sure we need this... an empty context will do in this scenario

PR #247, opened 2020-03-06
Update rules for /content/* persistence

jmarrero:


jmarrero:
@carlosmunoz These new changes are a working solution. I seems like we just need the redirect on /content not in the jcr:root (/).

PR #249, opened 2020-03-07
Bump pantheon's version

jmarrero:
I think we should approve/merge this before doing the next minor tag.

### DIFF
--- a/container/containerInstall.sh
+++ b/container/containerInstall.sh
@@ -4,7 +4,7 @@ do
   STATUS=$(curl --max-time 4 -s -o /dev/null -w '%{http_code}' http://localhost:8080/is_alive)
   if [ $STATUS -eq 404 ]; then
     echo "Got 404! Sling should be ready!"
-    cp /pantheon-1.0-SNAPSHOT.jar /install/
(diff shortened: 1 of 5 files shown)


benradey:
This is straightforward enough, but is it necessary? It seems like this doesn't really add any value, and just introduces overhead to our release process, especially early on when we'll be releasing frequently. If we're going to be implementing a process like this, we should wait until later when Pantheon is more mature and stable.

jmarrero:
This is your @benradey and @carlosmunoz call. I think I remember Carlos asking about this when he was in town. It's up to you guys to decide I can close the PR if you guys don't need it.

benradey:
@carlosmunoz, do you have a strong opinion here? I would vote to close it.

carlosmunoz:
I think we need to solve this. I'm tagging @aprajshekhar as he might be familiar with the maven plugins available for automatically updating versions when we release.

Re: this specifically, either way. I guess I would merge just to have the latest version show up in the right place and create a different bundle while we're at it.

jmarrero:
We just released 0.1.1 without this change. I am going to close this PR and let this discussion happen in another forum.

PR #250, opened 2020-03-09
 vanityURL should be optional

xdavidson:
This PR addresses a business require change. vanityURL was initially introduced as a required field on medata data. During the MS3 discovery, our business stake holders stated that it should be optional.


benradey:
@xdavidson Is this PR still valid? It's been sitting for a month and it looks like no one ever reviewed it. What should we do with it?

xdavidson:
> @xdavidson Is this PR still valid? It's been sitting for a month and it looks like no one ever reviewed it. What should we do with it?

@benradey , i am going to close this PR. According to the new wireframes, there UI is going to be different.

PR #253, opened 2020-03-17
Ccs 3469 uploader

xdavidson:
This PR makes the following changes:

- adds a custom node type for repository: pant:bucket

- update uploader so that it adds standalongAttributeFile property for pantheon/bucket resource

### DIFF
--- a/pantheon-bundle/src/main/resources/SLING-INF/nodetypes/nodetypes.cnd
+++ b/pantheon-bundle/src/main/resources/SLING-INF/nodetypes/nodetypes.cnd
@@ -57,6 +57,7 @@
 [pant:symlink] > nt:unstructured
     - sling:resourceType (string) = 'pantheon/symlink' mandatory autocreated
 
+// Defines a repository level node. The node type pant:bucket may contain a property for standaloneAttributeFile.
 [pant:bucket] > nt:unstructured, mix:created, mix:lastModified
     - sling:resourceType (string) = 'pantheon/bucket' mandatory autocreated
 // Base type for titles
--- a/pantheon-bundle/src/main/resources/SLING-INF/nodetypes/nodetypes.cnd
+++ b/pantheon-bundle/src/main/resources/SLING-INF/nodetypes/nodetypes.cnd
@@ -56,6 +56,9 @@
 
 [pant:symlink] > nt:unstructured
     - sling:resourceType (string) = 'pantheon/symlink' mandatory autocreated
+// Defines a repository level node. The node type pant:workspace may contain a property for pant:attributeFile.
+[pant:workspace] > nt:unstructured, mix:created, mix:lastModified
+    - sling:resourceType (string) = 'pantheon/workspace' mandatory autocreated
 
 // Base type for titles
 //[pant:title] > nt:unstructured, sling:Resource, mix:referenceable, mix:created, mix:lastModified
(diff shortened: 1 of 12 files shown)


uploader/pantheon.py
```diff
@@ -108,6 +123,7 @@ def _print_response(filetype, path, response_code, reason):
 
 server: http://localhost:8080
 repository: pantheonSampleRepo
+attributeFile: path/to/attribute.adoc
```
carlosmunoz: Is camel-casing the right standard for yaml files?
Also, are you envisioning this as an attributes' file for the whole repository? or just modules? If the latter, I would consider naming this something like `moduleAttributesFile`

pantheon-bundle/src/main/resources/SLING-INF/nodetypes/nodetypes.cnd
```diff
@@ -57,6 +57,8 @@
 [pant:symlink] > nt:unstructured
     - sling:resourceType (string) = 'pantheon/symlink' mandatory autocreated
 
+[pant:bucket] > nt:unstructured, mix:created, mix:lastModified
+    - sling:resourceType (string) = 'pantheon/bucket' mandatory autocreated
```
benradey: I asked earlier in our QA call but I'll ask here so it's recorded - why is creating this new type necessary?

uploader/pantheon.py
benradey: camelCase is the right standard for everything. :)

uploader/pantheon.py
```diff
@@ -180,6 +197,9 @@ def process_file(path, filetype):
 
     Returns:
     list: It returns a list with value of the API call status_code and reason
+    :param path:
+    :param filetype:
+    :return:
```
benradey: Is this part of some doc syntax that I'm not familiar with? Otherwise, it looks like these lines are unnecessary, since you describe the method above.

uploader/pantheon.py
xdavidson: kubernetes use camelCase:
https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/

ansible uses snake_case:
https://github.com/ansible/ansible-examples/blob/master/lamp_simple/site.yml

pantheon-bundle/src/main/resources/SLING-INF/nodetypes/nodetypes.cnd
xdavidson: The node defines with this new type has a special purpose. for instance, it's a repository level node. It may contain a standaloneAttributeFile for the repository.

uploader/pantheon.py
xdavidson: those were generated by my editor. i will move those

uploader/pantheon.py
xdavidson: @carlosmunoz i'm envisioning it as an attribute fiel for the whole repository.

uploader/pantheon.py
```diff
@@ -325,27 +354,58 @@ def processRegexMatches(files, globs, filetype):
...
+
+        if attributeFile and not os.path.isfile(attributeFile.strip()):
+            sys.exit('attributes: ' + attributeFile + ' does not exist.')
+
+        _info('Using ' + mode + ': ' + repository)
+        _info('Using attributes: ' + attributeFile)
+        print('--------------')
+
+        if attributeFile:
+            process_workspace(repository)
```
benradey: Since `process_workspace` is the method that creates the `pant:workspace` jcr node, this should run regardless of the presence of any `attributeFile`.

PR #255, opened 2020-03-18
Unicode file content is now handled properly

benradey:
Upload a module with some asian characters to Pantheon and preview it. Prior to this, the encoding was messed up.

### DIFF
--- a/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/ModuleVersionUpload.java
+++ b/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/ModuleVersionUpload.java
@@ -29,6 +29,7 @@ import org.slf4j.LoggerFactory;
 
 import javax.jcr.RepositoryException;
 import javax.servlet.http.HttpServletResponse;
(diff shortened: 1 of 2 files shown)


pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/ModuleVersionUpload.java
```diff
@@ -90,6 +90,7 @@ protected void doRun(SlingHttpServletRequest request, PostResponse response, Lis
         try {
             String locale = ServletUtils.paramValue(request, "locale", GlobalConfig.DEFAULT_MODULE_LOCALE.toString());
             String asciidocContent = ServletUtils.paramValue(request, "asciidoc");
+            asciidocContent = new String(asciidocContent.getBytes(request.getCharacterEncoding()), "utf-8");
```
carlosmunoz: Would you prefer to use something like `StandardCharsets.UTF_8`?

benradey: Fixed

PR #257, opened 2020-03-20
introduce scripts to run Pantheon locally for development

carlosmunoz:
This change introduces the following two shell scripts for running  Pantheon in a development machine using Podman.

`recreateSlingPod.sh` - (Re)Creates a full running Pantheon pod, database and all.

This script creates a pod with 3 mongo database containers forming a replica set, and a single container for sling running a build of the existing Pantheon source bundle.

`recreateSlingContainer.sh` - Recreates just the running container for Pantheon.

This script is just a shortcut to just re-create the running pantheon container leaving the database intact (to not lose any data, for example).

This is by no means perfect but it would help for all developers running on a similar environment. Suggestions welcome.

### DIFF
--- a/recreateSlingContainer.sh
+++ b/recreateSlingContainer.sh
@@ -1,0 +1,26 @@
+#
+# Recreates the Sling container ONLY. It will recompile the bundle and the slingstart
+# module, and will drop the existing container and recreate it.
+# Useful when the recreateSlingPod.sh script has been run beforehand.
+#
+# This script assumes that a database is already running. See the SLING_DB variable below
+# for details.
+#
+echo "Building Pantheon bundle and slingstart..."
+./mvnw clean package -DskipTests
+
+echo "Building Pantheon image..."
+buildah bud --layers -f container/Dockerfile -t pantheon-app .
+
+echo "Stopping existing Pantheon container..."
+podman stop pantheon-app
+
+echo "Removing previosuly stopped Pantheon container..."
+podman rm -f pantheon-app
+
+echo "Running new Pantheon container..."
+# Use this db property to connect to a running replica set pod (see recreateReplicaSetpod.sh)
+SLING_DB='mongodb://localhost:30001,localhost:30002,localhost:30003'
+# Use this db property to connect to a simple DB
+#SLING_DB='mongodb://localhost:27017'
+podman run --pod pantheon -d -e SLING_OPTS="-Dsling.run.modes=oak_mongo -Doak.mongo.uri=$SLING_DB" --name pantheon-app pantheon-app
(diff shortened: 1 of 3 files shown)


recreateSlingContainer.sh
```diff
@@ -0,0 +1,26 @@
...
+# for details.
+#
+echo "Building Pantheon bundle and slingstart..."
+./mvnw clean package -DskipTests
+
+echo "Building Pantheon image..."
+buildah bud --layers -f container/Dockerfile -t pantheon-app .
+
+echo "Stopping existing Pantheon container..."
+podman stop pantheon-app
```
benradey: Carlos, on a fresh system, the script stops here with this message:

```
Stopping existing Pantheon container...
Error: no container with name or ID pantheon-app found: no such container
Removing previosuly stopped Pantheon container...
Error: Failed to evict container: "": Failed to find container "pantheon-app" in state: no container with name or ID pantheon-app found: no such container
Running new Pantheon container...
Error: no pod with name or ID pantheon found: no such pod
```

recreateSlingPod.sh
```diff
@@ -0,0 +1,59 @@
...
+mongo --port 30001 --eval  "$MONGO_COMMANDS"
+
+echo "Building Pantheon bundle and slingstart..."
+./mvnw clean package -DskipTests
+
+echo "Building Pantheon container image..."
+buildah bud --layers -f container/Dockerfile -t pantheon-app .
+
+echo "Running new Pantheon container..."
+podman run --pod pantheon -d -e SLING_OPTS="-Dsling.run.modes=oak_mongo -Doak.mongo.uri=$SLING_DB" --name pantheon-app pantheon-app
```
benradey: Carlos, this script ran successfully and the container started, but Sling never brought Pantheon up. I navigated to localhost:8080 and I got the "Sling is starting..." page for 15 minutes before I gave up.

benradey:
@carlosmunoz Hey Carlos, it looks like this PR has been idle for almost a month with no activity. What are your plans here?

carlosmunoz:
Well, since it hasn't been reviewed I might just close it. I don't think there's been a lot of appetite from the rest of the team to run in containers and I can see why. Creating a containerized mongo replica set is still a valid addition, so maybe I will submit a new pull request with that.

benradey:
@carlosmunoz I had reviewed it, and based on my old comment above it looks like Pantheon didn't want to start inside the container on my machine. I think we had chatted about it but I don't recall what the ultimate outcome was from that. Do you remember?

carlosmunoz:
We never attempted anything after that. I say we close it and re-open it if a need arises.

benradey:
Sounds good, thanks Carlos. I'll go ahead and close it.

PR #260, opened 2020-03-24
  CCS-3473: Create a validation framework for Pantheon 2.

aprajshekhar:
This PR implements foundational framework for validation. The main components are:

1. Validator - base class for all validations
2. ModuleIncludeValidator - A sample validator
3. ValidatorService - A service that provides validation services. Calls the validator based on the type of validator. Returns list of zero or more validation error messages
4. ValidationMessageHelper - Provides message corresponding to error key

The decision whether or not to persist is deferred to the caller as caller would have the flexibility to persist or not.

### DIFF
--- a/pantheon-bundle/src/main/java/com/redhat/pantheon/validation/engine/ValidatorService.java
+++ b/pantheon-bundle/src/main/java/com/redhat/pantheon/validation/engine/ValidatorService.java
@@ -1,0 +1,68 @@
+package com.redhat.pantheon.validation.engine;
+
+import com.redhat.pantheon.model.api.SlingModel;
+import com.redhat.pantheon.model.module.Module;
+import com.redhat.pantheon.validation.validators.ModuleIncludeValidator;
+import com.redhat.pantheon.validation.validators.ValidationMessageHelper;
+import com.redhat.pantheon.validation.validators.Validator;
+import org.apache.sling.api.resource.Resource;
+import org.osgi.framework.BundleContext;
+import org.osgi.framework.Constants;
+import org.osgi.framework.FrameworkUtil;
+import org.osgi.service.component.annotations.Activate;
+import org.osgi.service.component.annotations.Component;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+
+/**
+ * Business service class that provides validation services.
+ */
+@Component(
+        service = ValidatorService.class,
+        property = {
+                Constants.SERVICE_DESCRIPTION + "=Provides validation services",
+                Constants.SERVICE_VENDOR + "=Red Hat Content Tooling team"
+        })
+
+public class ValidatorService {
+
+    private BundleContext bundleContext;
+    private   HashMap<Class, Validator> map;
+    /**
+     * Calls the validator based on the model and returns zero or more  error messages as a list
+     * @param model sling model to be validated
+     */
+    public List<String> validate(SlingModel model){
+        //get interfaces returns array of interfaces implemented by the model
+        //[0] will contain very first interface. For example, dynamic proxy implements
+        // Module that in turn implements SlingModel. Hence, element at [0] will be Module
+        return map.get(model.getClass().getInterfaces()[0]).validate(model);
+    }
+
+    /**
+     * The context of the current bundle. This is required to initialize
+     * the validators
+     * @return
+     */
+    public BundleContext getBundleContext() {
+        if(bundleContext == null) {
+            return FrameworkUtil.getBundle(
+                    this.getClass()).getBundleContext();
+        }
+        return bundleContext;
+    }
+    public void setBundleContext(BundleContext bundleContext){
+        this.bundleContext = bundleContext;
+    }
+
+    @Activate
+    public void postConstruct(){
+        map = new HashMap<Class, Validator>(){
+            {
+                put(Module.class, new ModuleIncludeValidator(new ValidationMessageHelper(getBundleContext())));
+            }
+        };
+    }
+}
--- a/pantheon-bundle/src/main/java/com/redhat/pantheon/validation/validators/ModuleIncludeValidator.java
+++ b/pantheon-bundle/src/main/java/com/redhat/pantheon/validation/validators/ModuleIncludeValidator.java
@@ -1,0 +1,30 @@
+package com.redhat.pantheon.validation.validators;
+
+import com.redhat.pantheon.model.module.Module;
+import org.apache.sling.api.resource.Resource;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class ModuleIncludeValidator extends Validator<Module> {
+
+
+    public ModuleIncludeValidator(ValidationMessageHelper validationMessageHelper){
+        super(validationMessageHelper);
+    }
+
+    @Override
+    public List<String> validate(Module module) {
+        return checkIfModuleIncludesModule(module, new ArrayList<>());
+    }
+
+    private List<String> checkIfModuleIncludesModule(Module module, List<String> validationErrors) {
+        for (Resource resource:  module.getChildren()) {
+            if(resource.getResourceType().equalsIgnoreCase("pant:module")){
+                validationErrors.add(getValidationMessageHelper().getMessage("module.included"));
+                break;
+            }
+        }
+        return validationErrors;
+    }
+}
--- a/pantheon-bundle/src/main/java/com/redhat/pantheon/validation/validators/ValidationMessageHelper.java
+++ b/pantheon-bundle/src/main/java/com/redhat/pantheon/validation/validators/ValidationMessageHelper.java
@@ -1,0 +1,41 @@
+package com.redhat.pantheon.validation.validators;
+
+import org.osgi.framework.BundleContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.Properties;
+
+public class ValidationMessageHelper {
+    public static final String PROPERTIES_File = "validation-messages.properties";
+    private Properties properties;
+    private BundleContext bundleContext;
+    final Logger log = LoggerFactory.getLogger(ValidationMessageHelper.class);
+    public ValidationMessageHelper(BundleContext bundleContext) {
+        properties = new Properties();
+        this.bundleContext = bundleContext;
+        loadProperties();
+    }
+
+    private void loadProperties() {
+        try{
+            properties.load(bundleContext.getBundle().getResource(PROPERTIES_File).openStream());
+        } catch (IOException e) {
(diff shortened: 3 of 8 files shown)


pantheon-bundle/src/main/java/com/redhat/pantheon/validation/validators/ModuleIncludeValidator.java
```diff
@@ -0,0 +1,30 @@
+package com.redhat.pantheon.validation.validators;
+
+import com.redhat.pantheon.model.module.Module;
+import org.apache.sling.api.resource.Resource;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class ModuleIncludeValidator extends Validator<Module> {
```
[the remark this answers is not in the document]
benradey: That way, we could query for all classes that have those annotations - similar to our existing class `EventJobConsumer.java:84`

pantheon-bundle/src/main/java/com/redhat/pantheon/validation/engine/ValidatorService.java
```diff
@@ -0,0 +1,68 @@
...
+    private   HashMap<Class, Validator> map;
+    /**
+     * Calls the validator based on the model and returns zero or more  error messages as a list
+     * @param model sling model to be validated
+     */
+    public List<String> validate(SlingModel model){
+        //get interfaces returns array of interfaces implemented by the model
+        //[0] will contain very first interface. For example, dynamic proxy implements
+        // Module that in turn implements SlingModel. Hence, element at [0] will be Module
+        return map.get(model.getClass().getInterfaces()[0]).validate(model);
```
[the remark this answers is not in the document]
benradey: I also like that this makes the process of validation totally transparent to the caller. No need to know WHAT validation is being run, just "validate this thing please, whatever that means" is all that's necessary.

pantheon-bundle/src/main/java/com/redhat/pantheon/validation/validators/ModuleIncludeValidator.java
```diff
@@ -0,0 +1,30 @@
...
+        return checkIfModuleIncludesModule(module, new ArrayList<>());
+    }
+
+    private List<String> checkIfModuleIncludesModule(Module module, List<String> validationErrors) {
+        for (Resource resource:  module.getChildren()) {
+            if(resource.getResourceType().equalsIgnoreCase("pant:module")){
+                validationErrors.add(getValidationMessageHelper().getMessage("module.included"));
+                break;
+            }
+        }
```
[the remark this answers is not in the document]
benradey: It's also worth mentioning that the dependencies are actually resolved in the `SlingIncludeProcessor` class - I'm not sure how we could plug a Validator in there (as opposed to writing the logic directly in a Validator like we have here)... I'd have to think about it.

pantheon-bundle/src/main/java/com/redhat/pantheon/validation/validators/ValidationMessageHelper.java
```diff
@@ -0,0 +1,41 @@
+package com.redhat.pantheon.validation.validators;
+
+import org.osgi.framework.BundleContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.Properties;
+
+public class ValidationMessageHelper {
```
benradey: This file is interesting - what is the intention here?

Is it so that we can change the wording of messaging without needing to rebuild and redeploy?

If so, isn't the properties file also part of the bundle, which would mean we'd need to redeploy anyways in order to update it?

pantheon-bundle/src/main/java/com/redhat/pantheon/validation/validators/Validator.java
```diff
@@ -0,0 +1,37 @@
...
+    public  Validator(ValidationMessageHelper validationMessageHelper){
+        setValidationMessageHelper(validationMessageHelper);
+    }
+
+    /**
+     * Validates the object and retuns the error messages as a {@link java.util.List}
+     * @param objectToValidate
+     * @return list of zero or more validation error messages
+     */
+    public abstract List<String> validate(T objectToValidate);
```
benradey: I know I mentioned it earlier but I'll say it again here since this is the declaration, I think we should swap String for something a bit more robust!

pantheon-bundle/src/test/resources/validation-messages.properties
```diff
@@ -0,0 +1 @@
+module.icluded="Module cannot include another module"
```
benradey: Ah... I see that this file is duplicated for testing purposes. I'd like to avoid that if possible. We might be able to use a symlink?

pantheon-bundle/src/test/java/com/redhat/pantheon/validation/engine/ValidatorServiceTest.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    @Test
+    public void testValidationServiceWhenModuleIsIncludedInModule(){
+        module.createChild("childModule", Module.class);
+        ValidatorService validatorService = new ValidatorService();
+        validatorService.setBundleContext(bundleContext);
+        validatorService.postConstruct();
+        validatorService.validate(module);
+        Assertions.assertFalse(validatorService.validate(module).isEmpty());
```
benradey: So this test and the previous test took me a second to understand, but I get it now. Makes sense! Just a reminder though, these tests would need to change once we have the real module inclusion logic ironed out.

aprajshekhar:
@benradey thanks for the review. Based on the review comments, I, now have an idea of what is being envisioned for validation. Based on that these are my inferences (a bit long, please bear with me :) ) :
**Validation**
As you might already know that there are three major variants for implementing validation frameworks:

1. Validators with single entry point based on annotations/XML:
Hibernate validator is example of such a framework. The annotations/XML sets up validation for POJOs. And the validation is triggered via a entry point class.

2. Validators as package of classes where each package validates a specific object
Apache validator comes close to this. The validators are classes and each validator validates a specific type of object eg. Numeric, Currency, Percent etc. In this case, validation framework is a set of helper methods that each package or group provides to either plugin the whole set of validators to another framework (e.g. Apache Validator with Struts) or run the whole set without calling them individually.

3. Validators as rules
In this case, the rule engine is used to execute rules that validate the objects. Drools can be used for such validations.

While implementing, I was leaning towards 1 for initial implementation and 3 for a more encompassing implementation. However, your comments have made me realise that 2 is the best approach for us. In order to go that way, I have following steps in mind:

a. Create packages for each main type of model to be validated. The classes with in the packages would be OSGi components that could be used individually via injection.

b  Implement helper classes for each package that would provide way to run multiple validators on a model e.g. validators for Module. 

In the future, this can be made more OSGi like by:

a. Make the models a separate OSGi bundle so that a validator bundle can use models without creating a circular dependency of Pantheon bundle on validator and validator on pantheon bundle. 
b. Implement Validators as a separate bundle so that new rules can be implemented without pantheon need to be touched unless  pantheon needs to use it (multiple versions).

**Messages**
Validation messages need to be externalized as thy may change from time to time. If messages are hard coded, then change in the message would require redeployment of the bundle. So, if pantheon can pull data from filesystem, we can mitigate this risk.

What are your thoughts?

benradey:
@aprajshekhar I agree that your approach #2 is closest to the approach that I have had in mind. And if my understanding of what you describe is correct, then I think you are exactly on-target with how to implement it while taking advantage of OSGi.

I wonder if we should extend your model to not only describe the object being validated (module, product, etc) but also the event that is firing the validation (upload, metadata update, etc). I think that would help us with the problem of needing to fire the include-validation logic at a very specific point in the build process - the SlingIncludeProcessor could just call the framework and say "please run `Module` validators that handle the `IncludeResolution` event" (or something along those lines) - and then the framework would still be responsible for finding those validators via OSGi magic. The framework itself and all of the events would be part of the main Pantheon bundle (and therefore available for any class to "request" at will), but the validators themselves would be plugged in via supporting minor bundles.

What do you think? Would you suggest another approach?

As far as messages - if you can write something that easily reads from the filesystem - that would be a bonus! But definitely not a requirement. In order to truly avoid the need to redeploy the application for minor language changes, we'll need to figure out how to write something that reads from Openshift - either from a persistent volume store, or environment variables, or something along those lines. We'd also have to modify our build pipeline to create that data automatically. That could be a lot of effort, but if you want to tackle it, you are welcome to give it a shot!

aprajshekhar:
@benradey Using events to plugin validations and provide output of a validation is something I have been thinking about. However, before doing a POC for the approach 2 + events, I would like more info on the following:

1. What exactly are the steps for validating an include resource? I am not asking in terms of implementation but in terms of functionality such as what part of module would be impacted, which attributed of Module (if any) would have to checked etc.? Without having a concrete idea of steps and actors involved, creating a POC is like shooting in the dark and I would really like to avoid that :).
2. If message hardcoding is not an issue for the present, we can table it for later. Anyhow sling's fsresource can be used to read external files.
3. Apart from Module, which other validation can I implement for POC so that I have a good no. of use cases with valid edge cases for the design.

In addition to this, will it be possible to break this story/task into POC, Design, and implementation? 

- The POC will help us in nailing down major use cases and edge cases and will also help us in discussing each part in detail without worrying much about unit tests as unit tests would take away from main aspect of POC. 
- The outcome of design would be a design document that can be used be the base for implementation. 
- The implementation would cover everything including unit tests and how QA would fit into the picture.
cc: @anknite 
Would love to know your thoughts on this.

benradey:
1. The actual check is easy. We just have to make sure that IF the type of document being validated is a module, THEN any document it includes is *not* a module. There are 2 attributes you could use for this, either `jcr:primaryType` or `sling:resourceType` would work. In java, you can use the `getResourceType()` method to inspect these properties, so it will be a very simple check to make.
2. Ok, sounds good!
3. The easiest validation you could use would be to ensure that the module has the required metadata populated. This would be the Product+Version association as well as the Document Use Case. (Currently the Vanity URL is also required, but that will become an optional field soon).

Yes, I think breaking down the story into POC/design/implementation is very reasonable. Please feel free to take this approach.

Regarding bundle dependency, I *think* there is a way in OSGi to indicate that bundle A requires functionality provided by bundle B. (A = a validator bundle, B = pantheon main bundle). I don't know the specifics, but I believe that as long as you make the correct declarations in the bundle jar's manifest then OSGi just handles things appropriately, and I hope that would alleviate the dependency concern. If it turns out that this is something difficult, then let's not worry about splitting things out into separate bundles for now.

Let me know if I missed anything in my reply. Thanks AP!

aprajshekhar:
As discussed, I have created CCS-3525 for the POC. Hence, closing this PR.

PR #264, opened 2020-04-01
FCCEUD-29: REVIEW - amehende

theashiot:
Issue ID: FCCEUD-29
Issue Link: https://projects.engineering.redhat.com/browse/FCCEUD-29

In con_comparison-between-modules-and-DITA-topics.adoc:

* Remove -include::../attributes.adoc[]
* Remove 'topic' at the end of the first sentence
* Capitalize one instance of 'xml'

In con_what-is-a-module.adoc:

* Remove - include::../attributes.adoc[]

### DIFF
--- a/pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/con_comparison-between-modules-and-DITA-topics.adoc
+++ b/pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/con_comparison-between-modules-and-DITA-topics.adoc
@@ -1,14 +1,13 @@
-include::../attributes.adoc[]
 [id='comparison-between-modules-and-DITA-topics_{context}']
 = Comparison between modules and DITA topics
 
-A module is similar to a topic in Darwin Information Typing Architecture (DITA) topic. 
+A module is similar to a topic in Darwin Information Typing Architecture (DITA). 
 
 A module differs from a DITA-topic in the following aspects:
 
 * Source language
 + 
-Modules are written AsciiDoc, whereas DITA topics are written in xml.
+Modules are written AsciiDoc, whereas DITA topics are written in XML.
 
 * Structure
 +
--- a/pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/con_what-is-a-module.adoc
+++ b/pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/con_what-is-a-module.adoc
@@ -1,4 +1,3 @@
-include::../attributes.adoc[]
 [id='what-is-a-module_{context}']
 = What is a module
 


theashiot:
@redhataccess/pantheon-developers Please review for technical completeness and accuracy.

benradey:
LGTM

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/con_comparison-between-modules-and-DITA-topics.adoc
```diff
@@ -1,14 +1,13 @@
...
 
 A module differs from a DITA-topic in the following aspects:
 
 * Source language
 + 
-Modules are written AsciiDoc, whereas DITA topics are written in xml.
+Modules are written AsciiDoc, whereas DITA topics are written in XML.
 
 * Structure
 +
```
stoobie, line 13: For some reason, I can't add comments below this line... @benradey, any idea why that might be?

Anyway, my comments:

"Following is the comparison..." You can just change this to a title for the table. Alternatively, "the following table compares the structure of a DITA topic and a module."

I think that this table is kind of misleading, though...I'm not real sure what you're trying to show here. A module, ideally, should be structured like a DITA topic, even though it doesn't use XML. There's no meaningful difference between `<title>Topic title</title>` and `= Topic title`. And you should ideally have a short description, body, and related links.



Also, technically, you can have nested topics within a topic. You're just not supposed to according to the Mod Docs guidelines.

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/con_comparison-between-modules-and-DITA-topics.adoc
```diff
@@ -1,14 +1,13 @@
...
 
-A module is similar to a topic in Darwin Information Typing Architecture (DITA) topic. 
+A module is similar to a topic in Darwin Information Typing Architecture (DITA). 
 
 A module differs from a DITA-topic in the following aspects:
 
 * Source language
 + 
-Modules are written AsciiDoc, whereas DITA topics are written in xml.
+Modules are written AsciiDoc, whereas DITA topics are written in XML.
```
stoobie, line 10: DITA is a form of XML. DITA-based topics are written in DITA XML, or just DITA.

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/con_what-is-a-module.adoc
```diff
@@ -1,4 +1,3 @@
-include::../attributes.adoc[]
 [id='what-is-a-module_{context}']
 = What is a module
```
stoobie, line 3: Again, I can only add comment up to this line. Strange.

"A module, in modular documentation, is defined as topic-specific structured informational content written in the specified syntax, stored in a file, that can be published on its own and as a part of a larger document in {ProductShortName}."

 => 

"A module is a piece of structured content that can stand on its own or as part of a larger document. Generally, a module is stored in its own file. Asciidoc is the source language for modules in {ProductShortName}."

stoobie: "The following are the properties of a module:" => "A module has the following properties:"

- Self-contained and standalone look redundant to me.
- In Self-contained, you switch from talking about the module to talking about the user. Also uses passive. Feels strange. I suggest you change to: "A module contains all the necessary information that a user needs to achieve a goal. A module explains a concept, provides a procedure to perform a task, or presents reference information."

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/con_what-is-a-module.adoc
```diff
@@ -1,4 +1,3 @@
-include::../attributes.adoc[]
```
stoobie, line 1: You have an attribute in the module: {ProductShortName}. If you're removing the attributes file, you need to hard code this. Why are you removing the attributes file reference?

adahms: Hey @stoobie - modules must not contain includes, even to attributes files. In this case, the attributes shall be supplied at the assembly level, and there is also some ongoing discussion around a solution for standalone modules that we can talk about during the sprint review.

For the time being, this change is designed to bring the modules in line with what we have today, and when it comes time to implement the standalone module change, we'll do that as part of a broader update.

stoobie:
@redhataccess/pantheon-developers Please merge.

PR #266, opened 2020-04-03
Drupal module apis

benradey:
Fixing up the existing module info servlet, and adding a dummy related module servlet, in order to unblock Drupal development. Since the related module servlet is a placeholder, no unit tests have been written. We can handle that once we implement the real logic.

### DIFF
--- a/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/IncludedInAssembliesServlet.java
+++ b/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/IncludedInAssembliesServlet.java
@@ -32,13 +32,13 @@ import static com.redhat.pantheon.servlet.ServletUtils.writeAsJson;
 import static javax.servlet.http.HttpServletResponse.SC_NOT_FOUND;
 
 /**
- * Get operation to render a Released Module's related module list in JSON format.
+ * Get operation to render a Released Module's related assembly list in JSON format.
  * Only two parameters are expected in the Get request:
  * 1. locale - Optional; indicates the locale that the module content is in, defaulted to en-US
  * 2. module_id - indicates the uuid string which uniquely identifies a module
  *
- * The url to GET a request from the server is /api/module/related
- * Example: <server_url>/api/module/related?locale=en-us&module_id=xyz
+ * The url to GET a request from the server is /api/module/assemblies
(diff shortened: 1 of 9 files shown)


pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/RelatedModuleServlet.java
```diff
@@ -0,0 +1,95 @@
...
+            }
+
+            resultStream = queryHelper.query("select * from [pant:module] as module", 3, 0);
+
+            List<Map> related = new ArrayList<>();
+            resultStream.map(r -> r.adaptTo(Module.class))
+                    .forEach(module -> {
+                            Map<String, String> m = new HashMap<>();
+                            m.put("title", module.getModuleLocale(locale).getVersion("1").metadata().get().title().get());
+                            m.put("url", "https://www.redhat.com");
```
xdavidson: I am wondering why do we set the url value to https://www.redhat.com?

benradey: I wanted it to be something that was easy to recognize as wrong. If we made it something realistic, someone might accidentally think that the related modules list was legitimate.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/IncludedInAssembliesServlet.java
```diff
@@ -0,0 +1,95 @@
...
+ *
+ * @author Ben Radey
+ */
+@Component(
+        service = Servlet.class,
+        property = {
+                Constants.SERVICE_DESCRIPTION + "=Servlet to facilitate GET operation which accepts locale and module uuid to output module assemblies",
+                Constants.SERVICE_VENDOR + "=Red Hat Content Tooling team"
+        })
+@SlingServletPaths(value = "/api/module/assemblies")
```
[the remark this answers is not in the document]
benradey: Great question! I don't think so - because this api is "what assemblies use this module" - so it is still module-centric. We will have an "/api/assembly" api in the future, which will likely include "what modules are used by this assembly" - and that will be assembly-centric, justifying the url structure.

PR #268, opened 2020-04-09
extract an abstract json processing servlet for POST and PUT operations

carlosmunoz:
Extract a servlet capable of parsing the request's body as a Json object. This should encapsulate the json parsing logic in a single place, allowing extending classes to focus (a bit more) on business logic.

### DIFF
--- a/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/AbstractJsonPostOrPutServlet.java
+++ b/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/AbstractJsonPostOrPutServlet.java
@@ -1,0 +1,85 @@
+package com.redhat.pantheon.servlet;
+
+import com.redhat.pantheon.helper.TransformToPojo;
+import org.apache.sling.api.SlingHttpServletRequest;
+import org.apache.sling.api.SlingHttpServletResponse;
+import org.apache.sling.api.servlets.SlingAllMethodsServlet;
+
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+/**
+ * Abstract servlet class to provide common behaviour that accepts a json payload in the body of a POST or PUT
+ * web request, parses said payload into a java class, and then operates on the object. All the json
+ * deserialization logic is encapsulated in this class.
+ *
+ * Implementations should override the processPost of processPut methods (both are allowed) to implement
+ * specific logic for either case.
+ *
+ * @author Carlos Munoz
+ */
+public class AbstractJsonPostOrPutServlet<T> extends SlingAllMethodsServlet {
+
+    private final Class<T> jsonType;
+
+    public AbstractJsonPostOrPutServlet(Class<T> jsonType) {
+        this.jsonType = jsonType;
+    }
+
+    @Override
+    protected final void doPost(SlingHttpServletRequest request, SlingHttpServletResponse response)
+            throws ServletException, IOException {
+        TransformToPojo transformToPojo = new TransformToPojo();
+        T jsonPayload = transformToPojo.fromJson(jsonType, request.getReader());
+
+        try {
+            processPost(request, response, jsonPayload);
+        } catch (Exception e) {
+            throw new ServletException(e);
+        }
+    }
+
+    @Override
+    protected final void doPut(SlingHttpServletRequest request, SlingHttpServletResponse response)
+            throws ServletException, IOException {
+        TransformToPojo transformToPojo = new TransformToPojo();
+        T jsonPayload = transformToPojo.fromJson(jsonType, request.getReader());
+
+        try {
+            processPut(request, response, jsonPayload);
+        } catch (Exception e) {
+            throw new ServletException(e);
+        }
+    }
+
+    /**
+     * Implements the logic in the event of a POST invocation. By default it simply sends a 405 response.
+     * @param request The servlet request. For parameter and path parameter extractions.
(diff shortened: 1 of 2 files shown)


pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/AbstractJsonPostOrPutServlet.java
```diff
@@ -0,0 +1,85 @@
...
+ * Abstract servlet class to provide common behaviour that accepts a json payload in the body of a POST or PUT
+ * web request, parses said payload into a java class, and then operates on the object. All the json
+ * deserialization logic is encapsulated in this class.
+ *
+ * Implementations should override the processPost of processPut methods (both are allowed) to implement
+ * specific logic for either case.
+ *
+ * @author Carlos Munoz
+ */
+public class AbstractJsonPostOrPutServlet<T> extends SlingAllMethodsServlet {
```
aprajshekhar, line 22: I like this idea. However, I have a query - do you see any more servlets in the near term (1-2 sprints) where JSON payload would be expected? If no, my suggestion would be to table this until such a use case comes. Having an extra hierarchy for just one servlet seems non-optimal at the moment. What do you think?

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/AbstractJsonPostOrPutServlet.java
```diff
@@ -0,0 +1,85 @@
...
+    @Override
+    protected final void doPost(SlingHttpServletRequest request, SlingHttpServletResponse response)
+            throws ServletException, IOException {
+        TransformToPojo transformToPojo = new TransformToPojo();
+        T jsonPayload = transformToPojo.fromJson(jsonType, request.getReader());
+
+        try {
+            processPost(request, response, jsonPayload);
+        } catch (Exception e) {
+            throw new ServletException(e);
```
aprajshekhar, line 39: Logging the exception here would help in tracing any exception in splunk.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/AbstractJsonPostOrPutServlet.java
```diff
@@ -0,0 +1,85 @@
...
+    @Override
+    protected final void doPut(SlingHttpServletRequest request, SlingHttpServletResponse response)
+            throws ServletException, IOException {
+        TransformToPojo transformToPojo = new TransformToPojo();
+        T jsonPayload = transformToPojo.fromJson(jsonType, request.getReader());
+
+        try {
+            processPut(request, response, jsonPayload);
+        } catch (Exception e) {
+            throw new ServletException(e);
```
aprajshekhar, line 52: Logging the exception here would help in tracing any exception in splunk.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/AbstractJsonPostOrPutServlet.java
carlosmunoz: Do you mean logging the exception instead of throwing it? or doing both?

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/AbstractJsonPostOrPutServlet.java
carlosmunoz: That's more a question for @benradey, @xdavidson and you really.
I'm happy to keep the PR on hold (somehow?) or close it if you feel we shouldn't add it... just scratching my coding itch :smile:

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/AbstractJsonPostOrPutServlet.java
aprajshekhar: Doing both. When client recieves the exception, having a log of the same on server side helps a lot in case debugging is required  :)

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/AbstractJsonPostOrPutServlet.java
aprajshekhar: Thanks :). I vote for holding it for now. @benradey, @xdavidson what do you think?

benradey: I would vote to bring it in. We already have `AbstractJsonSingleQueryServlet` in the project which only has 1 child based on it currently, so there is some precedent for abstracting things away immediately. Also, if we sit on this PR, we might accidentally forget that it exists when we go to create our 2nd post api (whenever that may be).

aprajshekhar:
@benradey I still think hierarchy with just one child is not an optimal route to go. Since this based on my experience and not statistics, I am fine with merging this :). However, this is the main reason we should start using SonarQube for code and security analysis so that we have analysis reports to make our code  better.

carlosmunoz:
@benradey @aprajshekhar Ok, I had to rebase due to some conflicting changes so please give it another go. I'll go ahead and merge since AP's concerns are non-binding. Lets keep an eye out and if we never need this we can consider removing the extra layer.

carlosmunoz:
@benradey sorry, had to rebase again as more changes conflicted

PR #269, opened 2020-04-13
Add new ACL for status-user

xdavidson:
This PR introduces a new ACL for status-user, whom will be used by hydra acknowledgement. It also 
  * declares a new node type: pant:documentStatus
  * added jcr:lastModified property to status node
  * it creates a status node automatically for each version


pantheon-bundle/src/main/java/com/redhat/pantheon/model/module/DocumentStatus.java
```diff
@@ -1,18 +1,27 @@
...
-    @Named("pant:status")
-    Field<String> status();
+import org.apache.jackrabbit.JcrConstants;
+
+import com.redhat.pantheon.model.api.Field;
+import com.redhat.pantheon.model.api.WorkspaceChild;
+import com.redhat.pantheon.model.api.annotation.JcrPrimaryType;
+
+@JcrPrimaryType("pant:documentStatus")
+public interface DocumentStatus extends WorkspaceChild {
```
aprajshekhar, line 14: One query - since the status represents status of acknowledgement wouldn't AckStatus or something similar be a better name?

xdavidson: AP, thank you for the feedback. this is an effort of co working with ben. We took your advice, and refactored some of the names

PR #271, opened 2020-04-14
Status user group

xdavidson:
This PR introduces
  * create a new pantheon role called pantheon-statusProvider. Providing a role allows us to create multiple users that have the same acls
  * replace the service user called 'status-user'. One downside of status-user is  this user is recreated everytime we build. 
  * add tracking of lastModifiedBy property on ackStatus node.


pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/StatusAcknowledgeServlet.java
```diff
@@ -69,7 +70,7 @@ protected void doPost(@NotNull SlingHttpServletRequest request, @NotNull SlingHt
                 response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Locale other than en_US is not supported");
                 return;
             }
-            processAcknowledgementRequest(acknowledgement, module, moduleLocale);
+            processAcknowledgementRequest(acknowledgement, module, moduleLocale, request);
```
aprajshekhar, line 73: Suggestion: Since only user id/name is needed in createStatusNode method, passing the name would be a better approach than passing the whole request objects. This would make other methods independent of request object and could make refactoring, if required in near future, simpler.

xdavidson: AP, thank you for the suggestion. I've updated the code accordingly.

xdavidson:
> This is an interesting proposal, @xdavidson . I am not sure I understand why you want to _replace_ the status user.
> 
> I would understand if you had the group _and_ the user, and just put the user in the group for the purpose of the ACL.
> 
> Are you proposing that we create the status user by hand? If so, why?

Creating a group will allow us to accommodate future needs for  more than one status-user.
It also avoid having plain password in the code. Yes, we can manually change the user password in the UI.

benradey:
Here's my thought though... if you want to go the group route (which does make sense), let's still follow the pattern of creating the necessary user(s) in the pantheon.txt file. Specify a password that is identical to the user name, which we'll change by hand in production, just like we are already doing for the 'admin' and 'sync-service' users. Unless there is a strong reason to treat this particular use case differently, then let's be consistent. Would you agree with that?

xdavidson:
> Here's my thought though... if you want to go the group route (which does make sense), let's still follow the pattern of creating the necessary user(s) in the pantheon.txt file. Specify a password that is identical to the user name, which we'll change by hand in production, just like we are already doing for the 'admin' and 'sync-service' users. Unless there is a strong reason to treat this particular use case differently, then let's be consistent. Would you agree with that?
Both admin user and sync-service don't belong to any groups. What's special about admin user is it has 'admin' role and all the super power comes that that role. What's special about sync-service user is that we give this user a very generous ACL access.

I created the new group pantheon-statusProvider followed the same pattern as pantheon-authors or pantheon-publishers. Define the group without creating any users in code. Handle user creation and group assignment on the UI.

benradey:
Thanks for doing that. Looks good to me now, other than that the build is failing... :)

PR #273, opened 2020-04-17
Assembly paging prototype

carlosmunoz:
Create a prototype for a potential assembly paging mechanism.
Using the module infrastructure this revision creates a new api endpoint at 
`/api/assembly?id=<identifier>&page=<page num>`

where `id` is a module id (eventually an assembly id) and `page num` is the a page to fetch from the assembly (it's optional and defaults to the first page)

The  api returns the html for a single page, determined by `section` tags. These tags are the result of asciidoctor's processing and they are top level tags in the body of the assembly's html.

This PR also fixes a bug in the `SlingResourceIncludeProcessor` which causes invalid includes for includes at a third level or below.

### DIFF
--- a/pantheon-bundle/src/main/java/com/redhat/pantheon/asciidoctor/extension/SlingResourceIncludeProcessor.java
+++ b/pantheon-bundle/src/main/java/com/redhat/pantheon/asciidoctor/extension/SlingResourceIncludeProcessor.java
@@ -77,7 +77,8 @@ public class SlingResourceIncludeProcessor extends IncludeProcessor {
             log.warn("Could not find include for {}", target);
         }
 
-        reader.push_include(content, target, target, 1, attributes);
+        String includedFilePath = includeResource != null ? includeResource.getPath() : target;
+        reader.push_include(content, includedFilePath, includedFilePath, 1, attributes);
     }
 
     private Resource resolveWithSymlinks(String path, Resource pathParent) {
--- a/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/assembly/AssemblyPageRenderingServlet.java
+++ b/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/assembly/AssemblyPageRenderingServlet.java
@@ -1,0 +1,74 @@
+package com.redhat.pantheon.servlet.assembly;
+
+import com.google.common.base.Charsets;
+import com.redhat.pantheon.conf.GlobalConfig;
+import com.redhat.pantheon.html.Html;
+import com.redhat.pantheon.jcr.JcrQueryHelper;
+import com.redhat.pantheon.model.api.SlingModels;
+import com.redhat.pantheon.model.module.Module;
+import com.redhat.pantheon.model.module.ModuleVersion;
+import com.redhat.pantheon.servlet.ServletUtils;
+import org.apache.sling.api.SlingHttpServletRequest;
+import org.apache.sling.api.SlingHttpServletResponse;
+import org.apache.sling.api.resource.Resource;
+import org.apache.sling.api.servlets.SlingSafeMethodsServlet;
+import org.apache.sling.servlets.annotations.SlingServletPaths;
+import org.jetbrains.annotations.NotNull;
+import org.jsoup.nodes.Document;
+import org.osgi.framework.Constants;
+import org.osgi.service.component.annotations.Component;
+
+import javax.jcr.RepositoryException;
+import javax.servlet.Servlet;
+import javax.servlet.ServletException;
+import java.io.IOException;
+import java.io.Writer;
+import java.util.Optional;
+
+/**
+ * @author Carlos Munoz
+ */
+@Component(
+        service = Servlet.class,
+        property = {
+                Constants.SERVICE_DESCRIPTION + "=Renders an assembly page in html",
+                Constants.SERVICE_VENDOR + "=Red Hat Content Tooling team"
+        })
+@SlingServletPaths(value = "/api/assembly")
+public class AssemblyPageRenderingServlet extends SlingSafeMethodsServlet {
+
+    @Override
+    protected void doGet(@NotNull SlingHttpServletRequest request, @NotNull SlingHttpServletResponse response) throws ServletException, IOException {
+        String assemblyId = ServletUtils.paramValue(request, "id");
+        Long pageNum = ServletUtils.paramValueAsLong(request, "page", 1L);
+
+        try {
+            // First get the assembly
+            // Since there are no assemblies yet, we will get modules
+            JcrQueryHelper helper = new JcrQueryHelper(request.getResourceResolver());
+            Resource assemblyResource = helper.findById(assemblyId);
+            Module assembly = SlingModels.getModel(assemblyResource, Module.class);
+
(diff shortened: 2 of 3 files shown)


pantheon-bundle/src/main/java/com/redhat/pantheon/asciidoctor/extension/SlingResourceIncludeProcessor.java
```diff
@@ -77,7 +77,8 @@ public void process(Document document, PreprocessorReader reader, String target,
             log.warn("Could not find include for {}", target);
         }
 
-        reader.push_include(content, target, target, 1, attributes);
+        String includedFilePath = includeResource != null ? includeResource.getPath() : target;
+        reader.push_include(content, includedFilePath, includedFilePath, 1, attributes);
```
benradey, line 81: Nice find! At first I was confused because I had tried testing this with this document structure:
```
master.adoc
a/level2.adoc
a/b/level3.adoc
```
...and that worked perfectly in both master and your branch.
Then I added 2 more level just to be sure:
```
/a/b/c/level4.adoc
/a/b/c/d/level5.adoc
```
...and lo and behold, I saw the breakage in master, and the fix in this branch! So that was quite a surprise, but good find and good fix. I'm not sure I'm following why this fixes it though - would you care to elaborate?

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/assembly/AssemblyPageRenderingServlet.java
```diff
@@ -0,0 +1,74 @@
...
+            Optional<ModuleVersion> versionOpt = assembly.getReleasedVersion(GlobalConfig.DEFAULT_MODULE_LOCALE);
+            ModuleVersion version = versionOpt.orElse( assembly.getDraftVersion(GlobalConfig.DEFAULT_MODULE_LOCALE).get() );
+            String fullHtmlContent = version.content().get().cachedHtml().get().data().get();
+            // Now split it into the right page
+            String pageHtml = Html.parse(Charsets.UTF_8.name())
+                    .andThen(Document::body)
+                    .andThen(element -> element.select("section.sect1"))
+                    .andThen(elements -> elements.listIterator(pageNum.intValue()-1).next())
+                    .andThen(element -> element.outerHtml())
+                    .apply(fullHtmlContent);
```
benradey, line 64: Interesting technique. This is different from what we discussed, but seems like it accomplishes the same goal. Are you confident that this "sect1" construct is reliable? I would worry that it only exists because of our haml, and that some future unrelated haml change would unexpectedly break this functionality.

benradey: After additional testing, I found out that "sect1" does show up with vanilla asciidoctor (which is probably a good sign) BUT it is tightly tied to the module having a title. (I tested by using `== Some Title` syntax). If the title is at a deeper level, such as `=== Some Subtitle`, or if the module has no title at all, this code causes the assembly "pages" to span multiple modules.

pantheon-bundle/src/main/java/com/redhat/pantheon/asciidoctor/extension/SlingResourceIncludeProcessor.java
carlosmunoz: It doesn't work by itself. The adjustment you made to "fix" the target on line 43 completes the puzzle. It seems that the `getFile` and `getDir` methods from asciidoctorj don't return the right data every time (only at the top 2 levels it seems). They behave correctly however if when doing the include we tell asciidoctor the exact location of these files (hence why I just get the full path of the resource and send it with the include).

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/assembly/AssemblyPageRenderingServlet.java
carlosmunoz: The `section.sect1` selector happens to be the default rendering tag and style for vanilla asciidoctor, and one which we are *not* overriding in our haml templates. It *can* however be overriden, in which case we would have to change our splitting algorithm (e.g. instead of `section` tags we may decide to use `p` tags). The benefit of this approach is that these sections appear at the root of the body, so the splits are very clean. As for the technique you and I discussed, I did try it, but the problem is that the breaks may not happen cleanly (e.g. breaking in the middle of a `p` tag, or worse, a deeply nested structure).

We could explore if Jsoup allows us to "clean up" the html, but I did not go that far.

The scenario you described with the titles is an interesting one for sure. I guess we would need to explore the implications more deeply.

carlosmunoz: Jsoup in theory corrects badly broken html, so we could try to see how it does in this scenario.

pantheon-bundle/src/main/java/com/redhat/pantheon/asciidoctor/extension/SlingResourceIncludeProcessor.java
benradey: Wow, that's quite strange. But, ok! I'll chalk it up as an asciidoctor eccentricity. Thanks for elaborating.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/assembly/AssemblyPageRenderingServlet.java
benradey: @carlosmunoz I can't remember where we left off in our discussions, but let me know if you'd like to merge this in as-is - or if you intend to take it further. Even though there is certainly more research to be done, I think this code already could serve as a good starting point for another dev to start looking at in earnest.

carlosmunoz: So, I added a bit more code to try different assembly paging strategies. I still like the one by section better (the original one). Have a look and let me know what you think. I would like to keep this out of the code for a little while longer still, but we can discuss.

benradey: I was finally able to take a look at this. I see where you're going with the code, and I like it.

I don't know specifically what technique would be more robust - keying off an html comment, or an html tag (that we would add in either case), rather than something that we "hope" is present in the doc, feels like the right way to do it.

When I checked out the code and tried to run it locally, it didn't work. I don't see the `<!-- pantheon-module-whatever -->` tags being represented in the cachedHtml data, so there's nothing for the assembly servlet to split. Am I correct in thinking I ought to see the comments there? I see the include processor adding them at the code level, so it's quite odd that they would simply disappear...

carlosmunoz: Yes, I think you would still see them there. They get inserted at every include of a module, so perhaps you need to make sure in your sample you are including modules (as opposed to just resources).

PR #274, opened 2020-04-17
Custom MockSlingHttpServletRequest for the HttpServlet methods not supported by MockSlingHttpServletRequest

aprajshekhar:
MockSlingHttpServletRequest does not support the some of the methods HttpServletRequest including but not limited to getUserPrincipal. This PR extends the existing MockSlingHttpServletRequest to override those methods. In this PR I have overridden getUserPrincipal. 

@xdavidson this would fix the test failures you are experiencing for https://github.com/redhataccess/pantheon/pull/271


carlosmunoz:
:+1: @aprajshekhar 
I'll let Lisa review as this directly impacts her, but looks good from my side.

PR #280, opened 2020-04-22
update ACL to allow ackStatus node to be created for existing modules

xdavidson:
This PR updates the ACL so that if ackStatus node is not prepopulated, the pantheon-statusProvider user can create it.


xdavidson:
Ideally the glob pattern should be /*/ackStatus , But there is a sling bug.  Current only '*' works. 
The ACLs in the PR is not restrictive enough for pantheon-statusProvider. 
I'm closing this PR. Ben is going to find a way to auto populate ackStatus nodes for existing modules.

PR #281, opened 2020-04-22
Adding debug servlets

benradey:
Drop-dead simple servlets to aid in crafting queries.


benradey:
My rationale for these was twofold.

Firstly, I _thought_ that the repo viewer only supported SQL2. I was looking for a way to toggle to XPath mode and couldn't find one. I see now that simply providing an XPath query "just works." ARGH...

Secondly, using the servlet lets you see the actual data that is being returned, whereas using the viewer only gives you the node path of each result.

So, given that only half of my original rationale holds true, the value here is questionable. I don't think seeing the data helps that much, since in the code we'd simply be transforming the results into their SlingModel counterparts and interacting with them that way. If we prefer to close this PR, I understand.

carlosmunoz:
I'll leave that call to you. Also look at the `DefaultQueryServlet class`. It's a similar concept with some variations. Maybe we can integrate some of these changes on that one.

PR #282, opened 2020-04-23
Adding one-time ackStatus create servlet

benradey:
Designed to be called exactly once in our production environment and then removed later.


carlosmunoz:
This brings up an excellent question: how are we going to deal with content (JCR) structure changes?

This is a quick and dirty approach; introducing a servlet which then needs to be called manually and removed in the next version. One gap in this approach is that it might prevent us from upgrading multiple versions (think from version 1 to 3 where 2 and 3 both involve content changes) because we are removing the servlets as we move.

This approach is fine for Pantheon now because it's in pre-1.0 mode, but I propose we use a sort of "version control" for the JCR structure (ala liquibase. See https://www.liquibase.org/). We could use bundle initializers to execute the changes in order, the changes should be idempotent (much like yours is here) so they could be run multiple times without a problem.

I say we approve this, deploy it, run it, and then swiftly remote it per your plan. Then we can design a longer term approach (if needed).

PR #283, opened 2020-04-24
 CCS-3525: Create POC for validation framework

aprajshekhar:
This PR provides base framework for validations in Pantheon. The framework supports

- Synchronous validation where client instantiates the validators and calls the validation methods themselves.

- Asynchronous validation using Events (Sling Jobs) and Observer/Listener pattern. In this case, the client instantiates the validators needed and pass it on as event payload. Then, the client can register a listener for validation complete event. Once the validators are executed, the assimilated results are passed on to the client via the registered listeners.

The validators themselves extend Validator interface. How the data is passed and what logic has to be applied is left to the individual validator.

In this PR, StatusAcknowledgementServlet acts as client and NotNullValidator is the sample validator. StatusAcknowledgementServlet has example of both synchronous as well asynchronous  validations.
The following sequence diagrams depict how both kinds of Validations work (for brevity StatusAcknowledgementServlet has been depicted as AcknowledgementServlet):

Synchronous Validation flow
Asynchronous Validation flow(raising an event)
Asynchronous Validation flow_ Consuming Validation Event


pantheon-bundle/src/main/java/com/redhat/pantheon/validation/validators/NotNullValidator.java
```diff
@@ -0,0 +1,68 @@
...
+ *      <ol>Returns the constraint violations via {@see Violations} instance</ol>
+ *  </p>
+ */
+@Component( service = NotNullValidator.class,
+        property = {
+            Constants.SERVICE_DESCRIPTION + "=Provides validation services",
+            Constants.SERVICE_VENDOR + "=Red Hat Content Tooling team"
+        }
+)
+public class NotNullValidator implements Validator {
```
benradey, line 27: I imagine adding a few more properties (annotations?) here to somehow list what events this validator cares about, sort of like what we already have in `SlingModelAdaptorFactory`:

```
@Component(
        service = NotNullValidator.class,
        property = {
                EVENT_TYPE + "=com.redhat.pantheon.validation.events.Status_Update",
                EVENT_TYPE + "=com.redhat.pantheon.validation.events.New_Module",
                EVENT_TYPE + "=com.redhat.pantheon.validation.events.Metadata_Changed"
        }
)
```

I made that up and probably the syntax isn't the best, but something along those lines. I'm hoping you can be the OSGi expert and advise on how best to supply the information here.

aprajshekhar:
@benradey thanks for your comments. However, before going into each comments separately, I would like to ensure that we are on the same page regarding what functionality do a validation framework brings to the table.

What I am understanding from the comments is that **validation framework** is being conflated/combined with **integration framework**. Validation framework provides:

- Validators for various use cases 
- A way to enhance existing validators or creating new validators
- A way to ensure that each validator returns validation results in a consistent format e.g. a container class that can be used by each validator to return the result.

It is Integration framework that:

- Provides a way to call the components, in this case the validators, seamlessly 
- Encapsulates how the calls works. This is what you are envisioning
- Provides events or other such integration points for the caller
- Ensure results do not get cross-wired.

I was fine with providing a rudimentary integration framework which lives in **com.redhat.pantheon.validation.events** package. However, looking at the limitations of Sling event framework and what you are envisioning, that is not going to work as most of the comments are about integration.

So, let's do this piecemeal:

1. I will remove com.redhat.pantheon.validation.events from this PR
2. Lets discuss the implementation in com.redhat.pantheon.validation.model and com.redhat.pantheon.validation packages.
3. Use this PR as a base to create validators (here I would like somebody else to create the validators so that it can be enhanced based on the feedback)
4. Document the models that are being validated as well as usage of the validators
5. Based on  4th point discuss and design **Integration Framework** where Validation would be one part of it. Here, we can design a wrapper around events so that Sling's limitations don't hinder us.


Hibernate Validator, Apache commons validator all follow this practice of developing Validators first and integration points at a later stage. The integration points provided by Apache commons validator for Struts and JSF were developed much later after the community provided the details about how validators are being used by these projects and the recurring logic in calling the validators.

If we don't do this piecemeal, with Validators as first piece and Integration framework as a subsequent/later piece, then we won't be able to develop either a good Validation framework or a good Integration framework.

Regarding making validators a separate bundle, that is not in scope of this Jira as it has been already mentioned in the Jira itself. We can tackle it once Validation and Integration frameworks are in place.

Let me know your thoughts.

aprajshekhar:
@benradey , I have removed com.redhat.pantheon.validation.events from this PR so that we can focus on validation. I have created the following JIRA to discuss about integration:
https://projects.engineering.redhat.com/browse/CCS-3586

benradey:
Hey @aprajshekhar ... boy, where do I begin!

First, let me start off with a few thoughts that are scattered. The first one is that I realized there isn't any mechanism right now for persisting Violations to the JCR. I don't remember if we had discussed, but that should ultimately be the goal of this PR... to be able to run validation against a known-bad document, and see additional JCR nodes get created underneath that module in the JCR so that a future PR can build on that and either A.) show the results to the user in a details screen, or B.) prevent a document from publishing.

The next thought is that now that you've removed the `events` package from this PR, it looks like `CombinedViolations` and `ValidationClientDetails` do not currently serve a purpose. Do you want to keep them with the intention of continuing to build out functionality there, or are they obsolete?

Now, for your comments... I think I see what you mean with your comments about "validation framework" vs "integration framework" but in my mind, I disagree that they are necessarily separate. I see "events" (at least the synchronous ones) as tightly coupled to the validation framework - in other words, the upload servlet (or any part of the system that does processing) would fire a validation event, and only validators would be able to listen to the event - nothing else.

From your comments, I think you are envisioning a generic "integration" event-based framework where not only validators might want to listen to events, but other potential future subsystems might want to listen for those same events as well. Perhaps that is why you were trying to hook into the existing Sling event system for your asynchronous events, and in that case that makes perfect sense. At this point in time though, I'm not sure that such a thing is necessary.

I don't understand your concern regarding not being able to mix-and-match validators in an event-based world. IIUC what you mean, then I think the only time we would want to be able to selectively run validators is if we expose such a thing as a user-facing control, and I'm not sure what the purpose would be there, outside of perhaps manually invoking a periodic link-checking validator. (But again, we could fire nightly or weekly events for that too). Otherwise, we would be safe running the predetermined set of validators for a particular event, every time. For example, on "upload" event, run validators A and B; on "publish" event, run validators B, C and D; etc.

If we *really* want to expose dynamic mix-and-match validation to the user at some point, then that feels like a solvable problem regardless of the mechanism that we implement, event-based or not.

I agree that separating validators into separate bundles is out of scope for this PR - but we should still design for that in mind, and we can still use OSGi annotations in our classes, even if ultimately everything winds up in the same bundle for now.

I'd like us to focus on bringing in a simplistic event system to aid in validation as I described, synchronous-only for now, at least as a first step. We can build on it over time, or even potentially replace it wholesale if it comes to that.

Would you agree with that, or would you still have strong objections? Let me know what you think. Thanks AP!

aprajshekhar:
@benradey wow! So many points to tackle :).

Let me tackle them easiest to the most complicated:

**Persisting the results of validation**
We had a conversation on this at the very initial stage. That is the reason for existence of CombinedViolations and ValidationClient classes. Once the data to be saved is agreed upon, I can design the sling models to facilitate the persistence. Again, I would like that to be a different piece and not part of this PR so that following can be handled separately and in-depth:

1. Where would the new node reside - under version or under locale?
2. Should such a node be added to existing modules similar to was done for Status acknowledgement?
3. Validation might be for a module. However, the caller could be the UI or a different component such as a servlet. What all details of the caller should be persisted?
4. Which user group should have access to this node?

**Validators and events**
To use events (as described and supported by Sling), following steps have to be implemented:

1. Register event handler to listen to. 
2. Implement EventHandler interface.
3. Implement JobProcessor interface so that the event data can be retrieved.
4. Pass the data using basic datatypes i.e. persist-able via JCR-Oak.

If the each validator (by itself or via by extending itself from a base class) has to do all the above, it will make writing validators unnecessarily complex. Add to that the limitation enforced by Sling - the event payload can only have base data types. This would mean that each validator may have to have a way to retrieve data via JCR. 

I believe that validator comes under **Single Resonsibility Principle** i.e validator should only validate. The way you are thinking about event handling and the way I have seen it working in Sling is a bit different. I am fully onboard with on the idea that validators should be OSGi injectable. However, one cannot hook into event just be using annotations. And that is the reason, I am batting for a separate *integration framework* so that what *you are envisioning* can come true.

In short, if validators need to listen to events, they will have to hook into Sling event and job subsystem which would make validators do multiple things and not just validation. The event subsystem does not really has an immediate mode. So, events will always be asynchronous. If a separate integration framework is there, we can bypass the Sling event mechanism and its limitations.

So, I would again say, lets start developing validators, if the validation error model looks good for version .01 and based on the different validators and how they are called, design an event based/integration system.

I am not averse to usage of events but given the current state of event handling pattern in Sling, using them as a part of validators themselves, would not be a good idea.   Hence, I am opposed to validators listening to events.

Hope this has brought us more to being on the same page.

aprajshekhar:
Thank you Ben

PR #284, opened 2020-04-29
Apache Karaf build

carlosmunoz:
From the top level just run:

`$ mvn clean install -pl pantheon-bundle,pantheon-karaf-feature,pantheon-karaf-dist`

This will create a tar.gz karaf distribution under `pantheon-karaf-dist/target`

Untar this file and inside the extracted contents run the `bin/karaf` executable to get this instance going. Pantheon will run on `http://localhost:8181`

Right now, the configuration is all static. Take a look at the `pantheon-karaf-feature.xml` file to change the config parameters and run the command above. It needs a running mongo database to work. Will need to figure out how to override these via system properties or environment variables.


aprajshekhar:
@carlosmunoz this is great. Just one question - as this PR approach generates distribution every time the build is run, is this just for the initial deployment with subsequent deployments updating the pantheon bundle or every new deployment would be via distribution?

PR #286, opened 2020-05-06
resolves CORS issue using Filter

xdavidson:
This PR addresses the CORS issue CP raised against our API calls. It uses a SlingServletFilter to allow *.redhat.com to consume our api on GET/HEAD requests

### DIFF
--- a/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/SimpleFilter.java
+++ b/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/SimpleFilter.java
@@ -27,6 +27,11 @@ import javax.servlet.ServletException;
 import javax.servlet.ServletRequest;
 import javax.servlet.ServletResponse;
 
+import org.apache.sling.api.SlingHttpServletRequest;
+import org.apache.sling.api.SlingHttpServletResponse;
+import org.apache.sling.servlets.annotations.SlingServletFilter;
+import org.apache.sling.servlets.annotations.SlingServletFilterScope;
+import org.osgi.service.component.annotations.Component;
 //import org.apache.felix.scr.annotations.Component;
 //import org.apache.felix.scr.annotations.Properties;
 //import org.apache.felix.scr.annotations.Property;
@@ -49,22 +54,30 @@ import org.slf4j.LoggerFactory;
  *     @Property(name="service.ranking", intValue=1)
  * })
  */
-//@SlingFilter(order=1, description="A Simple Filter")
-//@Property(name="service.vendor", value="The Apache Software Foundation")
+@Component
+@SlingServletFilter(scope = {SlingServletFilterScope.REQUEST},
+                    methods = {"GET","HEAD"})
 public class SimpleFilter implements Filter {
     
+    public static final String DOMAIN_ALLOWED = ".redhat.com";
     private final Logger log = LoggerFactory.getLogger(SimpleFilter.class);
 
     public void init(FilterConfig filterConfig) throws ServletException {
     }
 
-    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain) throws IOException,
+    public void doFilter(final ServletRequest req, final ServletResponse res, final FilterChain chain) throws IOException,
             ServletException {
-        log.info("filter invoked - start");
+        final SlingHttpServletRequest request = (SlingHttpServletRequest)req;
+        final SlingHttpServletResponse response = (SlingHttpServletResponse)res;
+        String origin = request.getHeader("Origin");
+        if (origin != null) {
+            if (origin.contains(DOMAIN_ALLOWED)) {
+                 response.addHeader("Access-control-Allow-Origin", origin);
+                 response.addHeader("Access-control-Allow-Methods", "GET, HEAD, OPTIONS");
+            }
+        }
         chain.doFilter(request, response);
-        log.info("filter invoked - end");
     }
-
     public void destroy() {
     }
 
--- a/pantheon-bundle/src/test/java/com/redhat/pantheon/servlet/CorsEnablingFilterTest.java
+++ b/pantheon-bundle/src/test/java/com/redhat/pantheon/servlet/CorsEnablingFilterTest.java
@@ -2,21 +2,15 @@ package com.redhat.pantheon.servlet;
 
 import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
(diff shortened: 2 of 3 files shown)


pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/SimpleFilter.java
```diff
@@ -49,22 +54,30 @@
  *     @Property(name="service.ranking", intValue=1)
  * })
  */
-//@SlingFilter(order=1, description="A Simple Filter")
-//@Property(name="service.vendor", value="The Apache Software Foundation")
+@Component
+@SlingServletFilter(scope = {SlingServletFilterScope.REQUEST},
+                    methods = {"GET","HEAD"})
 public class SimpleFilter implements Filter {
```
carlosmunoz, line 60: Could we call this something like `CorsEnablingFilter`? (or a better name)

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/SimpleFilter.java
```diff
@@ -49,22 +54,30 @@
  *     @Property(name="service.ranking", intValue=1)
  * })
  */
-//@SlingFilter(order=1, description="A Simple Filter")
-//@Property(name="service.vendor", value="The Apache Software Foundation")
```
carlosmunoz, line 53: Lets add some properties here, similar to what the other servlets are using.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/SimpleFilter.java
```diff
@@ -49,22 +54,30 @@
...
+@Component
+@SlingServletFilter(scope = {SlingServletFilterScope.REQUEST},
+                    methods = {"GET","HEAD"})
```
carlosmunoz, line 59: GET  and HEAD will do for now, but we **might** need to extend this to other methods.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/SimpleFilter.java
```diff
@@ -49,22 +54,30 @@
...
+    public void doFilter(final ServletRequest req, final ServletResponse res, final FilterChain chain) throws IOException,
             ServletException {
-        log.info("filter invoked - start");
+        final SlingHttpServletRequest request = (SlingHttpServletRequest)req;
+        final SlingHttpServletResponse response = (SlingHttpServletResponse)res;
+        String origin = request.getHeader("Origin");
+        if (origin != null) {
+            if (origin.contains(DOMAIN_ALLOWED)) {
+                 response.addHeader("Access-control-Allow-Origin", origin);
+                 response.addHeader("Access-control-Allow-Methods", "GET, HEAD, OPTIONS");
```
carlosmunoz, line 76: You are missing the OPTIONS method in your annotation above.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/SimpleFilter.java
```diff
@@ -49,22 +54,30 @@
...
 
-    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain) throws IOException,
+    public void doFilter(final ServletRequest req, final ServletResponse res, final FilterChain chain) throws IOException,
             ServletException {
-        log.info("filter invoked - start");
+        final SlingHttpServletRequest request = (SlingHttpServletRequest)req;
+        final SlingHttpServletResponse response = (SlingHttpServletResponse)res;
+        String origin = request.getHeader("Origin");
+        if (origin != null) {
+            if (origin.contains(DOMAIN_ALLOWED)) {
```
carlosmunoz, line 74: not a blocker, but since this is a security-related change, I would make this condition a bit tighter. Maybe `endsWith` would work, or a regular expression check. 
Maybe I'm being too careful here.. so I'll leave this decision to you.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/SimpleFilter.java
xdavidson: done

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/SimpleFilter.java
xdavidson: done

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/SimpleFilter.java
xdavidson: i added pattern value to limit when this filter will run:
@SlingServletFilter(scope = {SlingServletFilterScope.REQUEST},
                    pattern = "/api/.*",
                    methods = {"GET","HEAD", "OPTIONS"})

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/SimpleFilter.java
xdavidson: updated

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/SimpleFilter.java
xdavidson: updated

pantheon-bundle/src/test/java/com/redhat/pantheon/servlet/CorsEnablingFilterTest.java
```diff
@@ -0,0 +1,73 @@
...
+        // set headers
+        request.addHeader("Origin", "https://www.redhat.com");
+
+        CorsEnablingFilter filter = new CorsEnablingFilter();
+
+        // Execute the method with the mocks we want to test
+        filter.doFilter(request, response, chain);
+
+        // Verify that chain.doFilter() was called
+        verify(chain).doFilter(request, response);
```
carlosmunoz, line 71: I would add the verification that the headers are being set. I'll approve the PR, but consider adding this to the unit test otherwise the only thing being tested is that the `doFilter` method is invoked, which actually happens on every invocation.

```suggestion
        verify(chain).doFilter(request, response);
        verify(response).addHeader("Access-control-Allow-Origin", "https://www.redhat.com");
        verify(response).addHeader("Access-control-Allow-Methods", "GET, HEAD, OPTIONS");
```

xdavidson: thank you for the suggestion!

PR #287, opened 2020-05-09
add a more compact null-safe way to traverse jcr structures using models

carlosmunoz:
Add a more compact null-safe way to traverse jcr structures using models.
This revision deprecates the existing `Child.map` method and adds a similar but more compact way to do complex JCR structure traversals and avoid `null` elements in the path. 

See the unit tests for the way to use it.

### DIFF
--- a/pantheon-bundle/src/main/java/com/redhat/pantheon/model/api/util/SafeResourceTraversal.java
+++ b/pantheon-bundle/src/main/java/com/redhat/pantheon/model/api/util/SafeResourceTraversal.java
@@ -1,0 +1,91 @@
+package com.redhat.pantheon.model.api.util;
+
+import com.redhat.pantheon.model.api.Child;
+import com.redhat.pantheon.model.api.Field;
+import com.redhat.pantheon.model.api.SlingModel;
+
+import javax.annotation.Nullable;
+import java.util.Optional;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+import static java.util.Optional.ofNullable;
+
+/**
+ * Utility class which offers a null-safe way to traverse deep {@link SlingModel}
+ * structures. Traversals represent paths across multiple nodes ending in deeper
+ * child nodes, or properties. If any of the intermediary nodes is not present, the whole traversal
+ * will still conclude, with a final result of null.<br/>
+ * <br/>
+ * To use it, invoke the {@link SafeResourceTraversal#start(SlingModel)} method with a
+ * model to be traversed. From there on, the whole structure may be traversed all the
+ * way down to specific fields.
+ *
+ * @author Carlos Munoz
+ */
+public class SafeResourceTraversal<T extends SlingModel> implements Supplier<T> {
+
+    private static final SafeResourceTraversal<?> EMPTY = new SafeResourceTraversal<>(null);
+
+    private final Optional<T> currentResource;
+
+    private SafeResourceTraversal(T resource) {
+        this.currentResource = ofNullable(resource);
+    }
+
+    /**
+     * Starts a safe traversal.
+     * @param model The {@link SlingModel} to traverse.
+     * @param <U>
+     * @return A traversal object starting from the given {@link SlingModel}. If the model
+     * is null, traversals will still conclude but will always yield null results.
+     */
+    public static final <U extends SlingModel> SafeResourceTraversal<U> start(@Nullable U model) {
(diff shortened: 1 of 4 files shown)


carlosmunoz:
@xdavidson I was working on this as part of the JCR structure changes, but I realized this could be integrated sooner. It doesn't change any existing code as I want to avoid conflicts with the JCR structure changes, but it does deprecate one of the methods in use. My goal is to get this integrated into the main branch so we may be able to use it in the JCR restructure branch. Let me know what you think.

pantheon-bundle/src/main/java/com/redhat/pantheon/model/api/util/SafeResourceTraversal.java
```diff
@@ -0,0 +1,91 @@
...
+ * <br/>
+ * To use it, invoke the {@link SafeResourceTraversal#start(SlingModel)} method with a
+ * model to be traversed. From there on, the whole structure may be traversed all the
+ * way down to specific fields.
+ *
+ * @author Carlos Munoz
+ */
+public class SafeResourceTraversal<T extends SlingModel> implements Supplier<T> {
+
+    private static final SafeResourceTraversal<?> EMPTY = new SafeResourceTraversal<>(null);
```
xdavidson, line 28: I have a question. Why are we using an unbounded wildcard type for EMPTY instead of `<T>`?

xdavidson: never mind. Further down  i see we have genetic types of `<T>` or `<U>` or `<P>`

pantheon-bundle/src/main/java/com/redhat/pantheon/model/api/util/SafeResourceTraversal.java
```diff
@@ -0,0 +1,91 @@
...
+
+    /**
+     * Traverses to a field in the current traversed node. This represents and end
+     * to the traversal as there is nothing to traverse from a field.
+     * @param fieldAccessor A function that returns a {@link Field} from the current resource.
+     * @param <P>
+     * @return An optional containing the value of the field. If the field is not present, or if
+     * any of the intermediary nodes in the traversal was not present, this optional is empty.
+     */
+    public <P> Optional<P> field(Function<? super T, Field<P>> fieldAccessor) {
```
xdavidson, line 69: can we name the generic type from `<P>` to `<F>`? I think the convention is to take the First letter of the type parameter.

pantheon-bundle/src/main/java/com/redhat/pantheon/model/api/util/SafeResourceTraversal.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    /**
+     * Starts a safe traversal.
+     * @param model The {@link SlingModel} to traverse.
+     * @param <U>
+     * @return A traversal object starting from the given {@link SlingModel}. If the model
+     * is null, traversals will still conclude but will always yield null results.
+     */
+    public static final <U extends SlingModel> SafeResourceTraversal<U> start(@Nullable U model) {
```
xdavidson, line 43: Can we rename to generic `<U>` to `<M>`? Since the type parameter if model...

pantheon-bundle/src/main/java/com/redhat/pantheon/model/api/util/SafeResourceTraversal.java
carlosmunoz: It's unbounded because it's a singleton (private too) and at this point we don't know a specific type parameter to provide. Further down in the `empty()` method you'll see that there is a cast to the specific parameterized type. Since these are all erased types (just for compilation), it all just works.

pantheon-bundle/src/main/java/com/redhat/pantheon/model/api/util/SafeResourceTraversal.java
carlosmunoz: done.

pantheon-bundle/src/main/java/com/redhat/pantheon/model/api/util/SafeResourceTraversal.java
carlosmunoz: done.

PR #306, opened 2020-06-03
FCCEUD-19 Describe products

stoobie:


stoobie:
@redhataccess/pantheon-developers Please review for technical completeness and accuracy.

pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/con_about_products.adoc
```diff
@@ -0,0 +1,8 @@
+[id='about-products_{context}']
+= About Products in {ProductName}
+
+A _product_ in Pantheon is a category that corresponds to a product in the conventional meaning of the word. For example: Red Hat Enterprise Linux (RHEL) is a Red Hat product. There is also a product in {ProductName} named Red Hat Enterprise Linux that contains the modules and assemblies for RHEL documentation.
+
+You assign the product name and version in the metadata of a module or assembly. That module or assembly then appears in the search results for that product.
```
benradey, line 6: This may be true but I'm not 100% positive. We do send the product information to solr. Whether solr does anything with that information - I'm less clear about that.

If you know for sure that this is accurate, then that's good enough for me, we don't need to dig into it further.

Otherwise, this looks good!

benradey: I know who to ask. I'll start an email thread internally with you on it and we'll find out!

benradey: Looks like we got our answer via the email thread. I'll mark this as resolved.

stoobie:
@redhataccess/pantheon-developers Ready to merge.

PR #307, opened 2020-06-04
sling servlet suffix utility

carlosmunoz:
Add a utility for dealing with Sling path suffixes. In this revision the utility allows for template-based extraction of path parameters, enabling building complex, non-static API paths.

### DIFF
--- a/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/util/SlingPathSuffix.java
+++ b/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/util/SlingPathSuffix.java
@@ -36,9 +36,14 @@ public class SlingPathSuffix {
     private Map<String, String> parameterMap;
     private final Pattern pattern;
 
-    public SlingPathSuffix(final String parameterTemplate) {
+    /**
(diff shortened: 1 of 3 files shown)


pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/util/SlingPathSuffix.java
```diff
@@ -0,0 +1,88 @@
...
+ *
+ * The suffix needs a template or pattern to be provided in order to be able to extract
+ * information from it. It may be a completely static pattern or contain placeholders for
+ * named parameters.
+ *
+ * @author Carlos Munoz
+ */
+public class SlingPathSuffix {
+
+    private static final Pattern PARAMETER_PATTERN = Pattern.compile("(\\{[a-zA-Z0-9]+})");
```
xdavidson, line 34: do we allow '-' in the sufix pattern?

carlosmunoz: **Yes** in the pattern (the url itself). **No** for the parameter names.

PR #312, opened 2020-06-09
Module uploads will now build canonical variant

benradey:
Variant selected is either the first match if any variant definitions have the canonical flag set, otherwise it builds the "DEFAULT" variant for modules.

One thing this does NOT have is the logic to copy metadata from any old released version onto the draft version. Is that functionality we still want? @xdavidson @carlosmunoz

### DIFF
--- a/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/ModuleVersionUpload.java
+++ b/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/ModuleVersionUpload.java
@@ -127,7 +127,13 @@ public class ModuleVersionUpload extends AbstractPostOperation {
                 asciidoctorService.getModuleHtml(module, localeObj, module.getWorkspace().getCanonicalVariantName(),
                         true, context, true);
 
-                Metadata metadata = moduleLocale.variants().getOrCreate().variant(moduleLocale.getWorkspace().getCanonicalVariantName()).getOrCreate().draft().getOrCreate().metadata().getOrCreate();
+                Metadata metadata = moduleLocale
+                        .variants().getOrCreate()
+                        .variant(
+                                moduleLocale.getWorkspace().getCanonicalVariantName())
+                        .getOrCreate()
+                        .draft().getOrCreate()
+                        .metadata().getOrCreate();
                 metadata.dateModified().set(Calendar.getInstance());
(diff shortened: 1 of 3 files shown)


pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/ModuleVersionUpload.java
```diff
@@ -103,13 +121,19 @@ protected void doRun(SlingHttpServletRequest request, PostResponse response, Lis
...
 
-                // TODO Html can no longer be generated on upload since there might be too many variants
-                // TODO This will need to be re-thought since metadata now lies on the variant
+                resolver.commit();
+
+                Map<String, Object> context = asciidoctorService.buildContextFromRequest(request);
+                asciidoctorService.getModuleHtml(module, localeObj, module.getWorkspace().getCanonicalVariantName(),
+                        true, context, true);
+
+                Metadata metadata = moduleLocale.variants().getOrCreate().variant(moduleLocale.getWorkspace().getCanonicalVariantName()).getOrCreate().draft().getOrCreate().metadata().getOrCreate();
```
carlosmunoz, line 130: lets break this up into multiple lines, both for readability and in case we are hit by NPEs

pantheon-bundle/src/test/java/com/redhat/pantheon/servlet/module/ModuleVersionUploadTest.java
```diff
@@ -49,7 +74,7 @@ void createFirstVersion() throws Exception {
         // Then
         assertEquals(HttpServletResponse.SC_CREATED, response.getStatusCode());
         assertNotNull(slingContext.resourceResolver().getResource("/content/repositories/test_workspace/entities/new/proc_module/es_ES/source/draft/jcr:content"));
-        assertNull(slingContext.resourceResolver().getResource("/content/repositories/test_workspace/entities/new/proc_module/es_ES/variants"));
+        assertNotNull(slingContext.resourceResolver().getResource("/content/repositories/test_workspace/entities/new/proc_module/es_ES/variants"));
```
carlosmunoz, line 77: a comment for all these tests:
we should at least add `verify` calls to ensure the appropriate methods are being called in the `asciidoctorService` and `serviceResourceResolvedProvider` objects

benradey:
@carlosmunoz Updated; please take another look.

PR #321, opened 2020-06-17
(CCS-3230) Single page assembly preview

rednitish:
Changes  corresponding to segregating assemblies/ modules and rendering on it,


pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/RenderingServlet.java
```diff
@@ -0,0 +1,77 @@
...
+ * query parameter of name 'ctx_product' provided in the url.
+ */
+@Component(
+        service = Servlet.class,
+        property = {
+                Constants.SERVICE_DESCRIPTION + "=Servlet which transforms asciidoc content into html",
+                Constants.SERVICE_VENDOR + "=Red Hat Content Tooling team"
+        })
+@SlingServletResourceTypes(
+        resourceTypes = { "pantheon/module" },
```
carlosmunoz, line 50: You are missing the `pantheon/assembly` resource type here if you want to use the same servlet to preview both.

pantheon-bundle/src/test/java/com/redhat/pantheon/servlet/AssemblyRenderingServletTest.java
```diff
@@ -0,0 +1,259 @@
...
+import java.util.Locale;
+
+import static com.redhat.pantheon.util.TestUtils.registerMockAdapter;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.ArgumentMatchers.*;
+import static org.mockito.Mockito.*;
+
+@ExtendWith({SlingContextExtension.class, MockitoExtension.class})
+public class AssemblyRenderingServletTest {
```
carlosmunoz, line 25: A better name for this test would be `AssemblyRenderingTest`

pantheon-bundle/src/test/java/com/redhat/pantheon/servlet/ModuleRenderingServletTest.java
```diff
@@ -19,11 +20,10 @@
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.ArgumentMatchers.*;
-import static org.mockito.Mockito.lenient;
-import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.*;
 
 @ExtendWith({SlingContextExtension.class, MockitoExtension.class})
-public class AsciidocRenderingServletTest {
+public class ModuleRenderingServletTest {
```
carlosmunoz, line 26: Same here regarding class naming: `ModuleRenderingTest` (since you are testing the `ModuleRendering` class)

pantheon-bundle/src/test/java/com/redhat/pantheon/servlet/ModuleRenderingServletTest.java
```diff
@@ -58,11 +59,13 @@ public void testGenerateHtmlFromAsciidoc() throws Exception {
                 .thenReturn("A generated html string");
 
         // Test class
-        AsciidocRenderingServlet servlet = new AsciidocRenderingServlet(asciidoctorService);
-        servlet.init();
+/*        RenderingServlet servlet = new RenderingServlet(asciidoctorService);
+        servlet.init();*/
```
carlosmunoz, line 63: Lets get rid of this comment

pantheon-bundle/src/test/java/com/redhat/pantheon/servlet/RenderingServletTest.java
```diff
@@ -0,0 +1,134 @@
...
+                .resource("/repo/entities/module/en_US/variants/DEFAULT/released/metadata")
+                .resource("/repo/entities/module/en_US/variants/DEFAULT/released/cached_html/jcr:content",
+                        "jcr:data", "A generated html string")
+                .commit();
+            registerMockAdapter(Module.class, slingContext);
+        Resource resource = slingContext.resourceResolver().getResource("/repo/entities/module");
+        slingContext.request().setResource(resource);
+        AsciidoctorService asciidoctorService = mock(AsciidoctorService.class);
+        ModuleRendering moduleRendering = mock(ModuleRendering.class);
+//        when(moduleRendering.getRenderedHTML(slingContext.request(), slingContext.response())).then()
```
carlosmunoz, line 54: lets get rid of this comment

pantheon-bundle/src/test/java/com/redhat/pantheon/servlet/RenderingServletTest.java
```diff
@@ -0,0 +1,134 @@
...
+                .resource("/repo/entities/module/en_US/variants/DEFAULT/released/metadata")
+                .resource("/repo/entities/module/en_US/variants/DEFAULT/released/cached_html/jcr:content",
+                        "jcr:data", "A generated html string")
+                .commit();
+        registerMockAdapter(Module.class, slingContext);
+        Resource resource = slingContext.resourceResolver().getResource("/repo/entities/module");
+        slingContext.request().setResource(resource);
+        AsciidoctorService asciidoctorService = mock(AsciidoctorService.class);
+        ModuleRendering moduleRendering = mock(ModuleRendering.class);
+//        when(moduleRendering.getRenderedHTML(slingContext.request(), slingContext.response())).then()
```
carlosmunoz, line 105: lets get rid of this comment

pantheon-bundle/src/main/java/com/redhat/pantheon/model/Rendering.java
```diff
@@ -0,0 +1,23 @@
...
+import java.io.IOException;
+import java.util.Calendar;
+
+/**
+ * Represents rendering as an interface to be implemented.
+ *
+ */
+
+public interface Rendering {
+    public  void getRenderedHTML(SlingHttpServletRequest request, SlingHttpServletResponse resposne) throws IOException;
```
carlosmunoz, line 22:
```suggestion
    public  void renderHtml(SlingHttpServletRequest request, SlingHttpServletResponse resposne) throws IOException;
```
since the method doesn't return anything

pantheon-bundle/src/main/java/com/redhat/pantheon/model/Rendering.java
```diff
@@ -0,0 +1,23 @@
...
+import javax.inject.Named;
+import java.io.IOException;
+import java.util.Calendar;
+
+/**
+ * Represents rendering as an interface to be implemented.
+ *
+ */
+
+public interface Rendering {
```
carlosmunoz, line 21: I like this idea, but I would move it to the `servlet` package

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/AssemblyRendering.java
```diff
@@ -0,0 +1,108 @@
...
+                    .toChild(ModuleLocale::source)
+                    .toChild(SourceContent::released)
+                    .getAsOptional();
+        }
+
+
+        if(!moduleVariantSource.isPresent()) {
+            response.sendError(HttpServletResponse.SC_NOT_FOUND, (draft ? "Draft " : "Released ")
+                    + "source content not found for " + variantName +  " module variant at "
+                    + request.getResource().getPath());
```
aprajshekhar, line 90: Rather than having an else you can simply return

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/RenderingServlet.java
```diff
@@ -0,0 +1,77 @@
...
+    @Activate
+    public RenderingServlet(
+            @Reference AsciidoctorService asciidoctorService) {
+        this.asciidoctorService = asciidoctorService;
+    }
+
+    @Override
+    protected void doGet(SlingHttpServletRequest request,
+            SlingHttpServletResponse response) throws ServletException, IOException {
+        if(request.getResource().getResourceType().equals("pantheon/module")){
```
aprajshekhar, line 69: Switch/case would be a better option than if/else

pantheon-bundle/src/test/java/com/redhat/pantheon/servlet/AssemblyRenderingServletTest.java
```diff
@@ -0,0 +1,259 @@
...
+                .resource("/repo/entities/module/en_US/source/draft/jcr:content",
+                        "jcr:data", "Some source content (irrelevant)")
+                .resource("/repo/entities/module/en_US/variants/DEFAULT/draft/metadata")
+                .resource("/repo/entities/module/en_US/variants/DEFAULT/draft/cached_html/jcr:content",
+                        "jcr:data", "A generated html string")
+                .commit();
+        registerMockAdapter(Module.class, slingContext);
+        Resource resource = slingContext.resourceResolver().getResource("/repo/entities/module");
+        slingContext.request().setResource(resource);
+        slingContext.request().getParameterMap().put(AssemblyRendering.PARAM_RERENDER, new String[]{"true"});
```
aprajshekhar, line 102: Sling recommends using setParameterMap than getParameterMap().put().
```slingContext.request().setParameterMap(
                ImmutableMap.<String, Object>builder()
                                .put(AsciidocRenderingServlet.PARAM_RERENDER, "true")
                        .build()
        );

File: pantheon-bundle/src/test/java/com/redhat/pantheon/servlet/AssemblyRenderingServletTest.java [hunk 12]
```diff
@@ -0,0 +1,259 @@
...
+        slingContext.request().getParameterMap().put(AssemblyRendering.PARAM_DRAFT, new String[]{"true"});
+        lenient().when(
```
aprajshekhar, line 104: Sling recommends using setParameterMap than getParameterMap().put().

```ImmutableMap.<String, Object>builder()
       .put(AsciidocRenderingServlet.PARAM_RERENDER, "true")
        .build()
 );
This and the above one can be merged as ImmutableMap builder supports multiple puts.

File: pantheon-bundle/src/test/java/com/redhat/pantheon/servlet/AssemblyRenderingServletTest.java [hunk 13]
```diff
@@ -0,0 +1,259 @@
...
+                .resource("/repo/entities/module/en_US/variants/variant1/draft/metadata")
+                .resource("/repo/entities/module/en_US/variants/variant1/draft/cached_html/jcr:content",
+                        "jcr:data", "A generated html string")
+                .commit();
+        registerMockAdapter(Module.class, slingContext);
+        Resource resource = slingContext.resourceResolver().getResource("/repo/entities/module");
+        slingContext.request().setResource(resource);
+        slingContext.request().getParameterMap().put(AssemblyRendering.PARAM_RERENDER, new String[]{"true"});
+        slingContext.request().getParameterMap().put(AssemblyRendering.PARAM_DRAFT, new String[]{"true"});
+        slingContext.request().getParameterMap().put(AssemblyRendering.PARAM_VARIANT, new String[]{"variant1"});
```
aprajshekhar, line 154: Same as my previous comment on use of getParameterMap.

pantheon-bundle/src/test/java/com/redhat/pantheon/servlet/AssemblyRenderingServletTest.java
```diff
@@ -0,0 +1,259 @@
...
+                .resource("/module/en_US/variants/DEFAULT/draft/cachedHtml/jcr:content",
+                        "jcr:data", "A generated html string")
+                .resource("/module/en_US/variants/DEFAULT/draft/metadata")
+                .commit();
+        registerMockAdapter(Module.class, slingContext);
+        Resource resource = slingContext.resourceResolver().getResource("/module");
+        slingContext.request().setResource(resource);
+        slingContext.request().getParameterMap().put(AssemblyRendering.PARAM_RERENDER, new String[]{"true"});
+        slingContext.request().getParameterMap().put(AssemblyRendering.PARAM_DRAFT, new String[]{"true"});
+        slingContext.request().getParameterMap().put(AssemblyRendering.PARAM_VARIANT, new String[]{"non_existing"});
```
aprajshekhar, line 201: Same as my previous comment on  use of getParameterMap.

pantheon-bundle/src/test/java/com/redhat/pantheon/servlet/ModuleRenderingServletTest.java
```diff
@@ -96,8 +100,8 @@ public void testGenerateHtmlFromReleasedAsciidocWithRerender() throws Exception
         registerMockAdapter(Module.class, slingContext);
         Resource resource = slingContext.resourceResolver().getResource("/repo/entities/module");
         slingContext.request().setResource(resource);
-        slingContext.request().getParameterMap().put(AsciidocRenderingServlet.PARAM_RERENDER, new String[]{"true"});
-        slingContext.request().getParameterMap().put(AsciidocRenderingServlet.PARAM_DRAFT, new String[]{"true"});
+        slingContext.request().getParameterMap().put(ModuleRendering.PARAM_RERENDER, new String[]{"true"});
+        slingContext.request().getParameterMap().put(ModuleRendering.PARAM_DRAFT, new String[]{"true"});
```
aprajshekhar, line 104: Same as my previous comment on  use of getParameterMap.

pantheon-bundle/src/test/java/com/redhat/pantheon/servlet/ModuleRenderingServletTest.java
```diff
@@ -147,9 +150,12 @@ public void testGenerateHtmlFromDraftAsciidocWithRerender() throws Exception {
         registerMockAdapter(Module.class, slingContext);
         Resource resource = slingContext.resourceResolver().getResource("/repo/entities/module");
         slingContext.request().setResource(resource);
-        slingContext.request().getParameterMap().put(AsciidocRenderingServlet.PARAM_RERENDER, new String[]{"true"});
-        slingContext.request().getParameterMap().put(AsciidocRenderingServlet.PARAM_DRAFT, new String[]{"true"});
-        slingContext.request().getParameterMap().put(AsciidocRenderingServlet.PARAM_VARIANT, new String[]{"variant1"});
+        slingContext.request().getParameterMap().put(ModuleRendering.PARAM_RERENDER, new String[]{"true"});
+        slingContext.request().getParameterMap().put(ModuleRendering.PARAM_DRAFT, new String[]{"true"});
+        slingContext.request().getParameterMap().put(ModuleRendering.PARAM_VARIANT, new String[]{"variant1"});
+
```
aprajshekhar, line 156: Same as my previous comment on use of getParameterMap.

pantheon-bundle/src/test/java/com/redhat/pantheon/servlet/ModuleRenderingServletTest.java
```diff
@@ -234,9 +236,9 @@ public void renderHtmlForInvalidLocale() throws Exception {
         registerMockAdapter(Module.class, slingContext);
         Resource resource = slingContext.resourceResolver().getResource("/module");
         slingContext.request().setResource(resource);
-        slingContext.request().getParameterMap().put(AsciidocRenderingServlet.PARAM_RERENDER, new String[]{"true"});
-        slingContext.request().getParameterMap().put(AsciidocRenderingServlet.PARAM_DRAFT, new String[]{"true"});
-        slingContext.request().getParameterMap().put(AsciidocRenderingServlet.PARAM_LOCALE, new String[]{"ja_JP"});
+        slingContext.request().getParameterMap().put(ModuleRendering.PARAM_RERENDER, new String[]{"true"});
+        slingContext.request().getParameterMap().put(ModuleRendering.PARAM_DRAFT, new String[]{"true"});
+        slingContext.request().getParameterMap().put(ModuleRendering.PARAM_LOCALE, new String[]{"ja_JP"});
```
aprajshekhar, line 241: Same as my previous comment on use of getParameterMap.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/AssemblyRendering.java
```diff
@@ -0,0 +1,108 @@
...
+        this.asciidoctorService = asciidoctorService;
+    }
+
+    @Override
+    public void getRenderedHTML(SlingHttpServletRequest request,
+            SlingHttpServletResponse response) throws IOException {
+        String locale = paramValue(request, PARAM_LOCALE, DEFAULT_MODULE_LOCALE.toString());
+        boolean draft = paramValueAsBoolean(request, PARAM_DRAFT);
+        boolean reRender = paramValueAsBoolean(request, PARAM_RERENDER);
+        String variantName = paramValue(request, PARAM_VARIANT, DEFAULT_VARIANT_NAME);
```
aprajshekhar, line 65: Since the locale, draft, rerender variables are being used in ModuleRender as well, wouldn't it be a better approach to move these statements to a helper/util?

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/RenderingServlet.java
benradey: Are we sure we want to use the same servlet for both?

In @xdavidson's recent PR, we agreed that our gut was that the upload servlets would likely be sufficiently different to warrant them being separate. I wonder if the same doesn't hold true for the rendering servlets?

I admit that off the top of my head, I'm not sure why the two rendering servlets might look different, at least from a client perspective, especially since this is for single-page assemblies. But I'm interested to hear other opinions on whether we should have one render servlet or two separate.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/AssemblyRendering.java
benradey: This also doesn't sit well with me. I would much rather see the `getRenderedHTML` method take locale, draft, etc as explicit parameters, rather than simply having it take the request object and extracting what it needs by itself. Then the `paramValueAsX` calls get combined into the parent class rather than duplicated in the child classes.

pantheon-bundle/src/main/java/com/redhat/pantheon/model/Rendering.java
benradey: Also consider renaming it to "DocumentRenderer" - class/interface names that end in "-ing" are unusual, and adding "Document" to the front of it makes it a bit more specific.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/AssemblyRendering.java
```diff
@@ -0,0 +1,108 @@
...
+ * Renders an HTML preview for a single module.
+ * To provide parameters to the asciidoc generation process, provide the parameters with their name prefixed
+ * with "ctx_".
+ *
+ * For example, if an asciidoc attribute of name 'product' needs to be passed, there will need to be a
+ * query parameter of name 'ctx_product' provided in the url.
+ */
+
+
+public class AssemblyRendering implements Rendering {
```
benradey, line 42: Rename to "AssemblyRenderer"

pantheon-bundle/src/main/java/com/redhat/pantheon/model/Rendering.java
benradey: Or... now that I think about it... maybe something more like "DocumentServletHandler" since this doesn't just render, but also takes care of populating the servlet response as well. That would make the name even more targeted toward its actual function.

pantheon-bundle/src/test/java/com/redhat/pantheon/servlet/AssemblyRenderingServletTest.java
rednitish: As per Ben's  and Carlos's suggestion Changed name to noun ( AssemblyRenderer)

pantheon-bundle/src/main/java/com/redhat/pantheon/model/Rendering.java
rednitish: @benradey  This is an interface, we can name implementation class as handler, but renaming it to DocumentRedererHandler doesn't sound good.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/RenderingServlet.java
rednitish: I have tried to create separate rendering servlet, to resolve this issue, considering, we have different resourcetype  as well for Module and assembly

pantheon-bundle/src/test/java/com/redhat/pantheon/servlet/ModuleRenderingServletTest.java
rednitish: Updated.

pantheon-bundle/src/test/java/com/redhat/pantheon/servlet/ModuleRenderingServletTest.java
rednitish: Copy that.. :)

pantheon-bundle/src/test/java/com/redhat/pantheon/servlet/RenderingServletTest.java
rednitish: Copy that.

pantheon-bundle/src/test/java/com/redhat/pantheon/servlet/RenderingServletTest.java
rednitish: Deprecated

pantheon-bundle/src/main/java/com/redhat/pantheon/model/Rendering.java
rednitish: Deprecated.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/AssemblyRendering.java
rednitish: resolved

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/AssemblyRendering.java
rednitish: I have made 2 separate servlet to handle above comment.

rednitish:
Thanks Carlos, Ben and AP. Updated the code, with review comments provided.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/RenderingServlet.java
benradey: Ok. I think that works for now. We can revisit the decision later.

benradey: (if necessary)

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/AssemblyRendering.java
benradey: I don't think this is resolved. It looks like you're missing a `return;` statement after the call to `response.sendError`.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/RenderingServlet.java
benradey: I'm going to mark this is resolved now that Nitish has split this into separate servlets.

benradey:
Hey @rednitish, I went through and made sure most of the old issues were addressed. Other than than the one comment that I left open about a missing return statement, this looks good enough to merge.

One thing that I do notice is that the tests are failing - both here with Travis, and locally on my machine. It looks like both the `ModuleRenderServletTest` and `AssemblyRenderServletTest` classes are throwing a NPE during testing. Please resolve that, and then once both those have been remedied and the return statement above has been added, then we can merge this.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/AssemblyRenderServlet.java
```diff
@@ -0,0 +1,121 @@
...
+                break;
+
+            case "false":
+                moduleVariantSource = module.moduleLocale(localeObj)
+                        .traverse()
+                        .toChild(ModuleLocale::source)
+                        .toChild(SourceContent::released)
+                        .getAsOptional();
+                break;
+        }
```
benradey, line 98:
```suggestion
        moduleVariantSource = module.moduleLocale(localeObj)
                .traverse()
                .toChild(ModuleLocale::source)
                .toChild(draft ? SourceContent::draft : SourceContent::released)
                .getAsOptional();
```

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/ModuleRendererServlet.java
```diff
@@ -62,36 +58,40 @@
...
+                        break;
+
+            case "false":
+                moduleVariantSource = module.moduleLocale(localeObj)
                     .traverse()
                     .toChild(ModuleLocale::source)
                     .toChild(SourceContent::released)
                     .getAsOptional();
+                break;
         }
```
benradey, line 95:
```suggestion
        moduleVariantSource = module.moduleLocale(localeObj)
                .traverse()
                .toChild(ModuleLocale::source)
                .toChild(draft ? SourceContent::draft : SourceContent::released)
                .getAsOptional();
```

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/AssemblyRenderServlet.java
```diff
@@ -0,0 +1,121 @@
...
+
+
+        if(!moduleVariantSource.isPresent()) {
+            response.sendError(HttpServletResponse.SC_NOT_FOUND, (draft ? "Draft " : "Released ")
+                    + "source content not found for " + variantName +  " module variant at "
+                    + request.getResource().getPath());
+        }
```
benradey, line 105:
```suggestion
        if(!moduleVariantSource.isPresent()) {
            response.sendError(HttpServletResponse.SC_NOT_FOUND, (draft ? "Draft " : "Released ")
                    + "source content not found for " + variantName +  " module variant at "
                    + request.getResource().getPath());
            return;
        }
```

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/ModuleRendererServlet.java
```diff
@@ -100,19 +100,18 @@ protected void doGet(SlingHttpServletRequest request,
                     + "source content not found for " + variantName +  " module variant at "
                     + request.getResource().getPath());
         }
```
benradey, line 102:
```suggestion
                    + "source content not found for " + variantName +  " module variant at "
                    + request.getResource().getPath());
            return;
        }
```

benradey:
@rednitish I made suggestions that ought to resolve the remaining concerns. Let me know what you think.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/AssemblyRenderServlet.java
rednitish: I have a query, here as we are using sendError method, shoudn't it automatically returns control back to parent class?
     * After using this method, the response should be considered
     * to be committed and should not be written to.

rednitish:
@benradey   Thanks for comments, i have them fixed in last commit. Please take a look :)

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/AssemblyRenderServlet.java
benradey: Hi @rednitish, the answer to your question is no, Java does not work that way. There is no way to call a method, and then that child method halts further execution of its parent method, outside of throwing an exception (which is not what's happening here). You must add an explicit `return` in order to stop your method from continuing to execute once the child finishes.

What the javadoc is saying is that once you call response.sendError, you should not attempt to call any further methods on the response object, because they will fail.

benradey:
@aprajshekhar @carlosmunoz Looks like Nitish isn't able to merge until you guys change your review statuses. Please take another look and decide whether further changes are necessary or whether this can be merged. Thanks.

PR #333, opened 2020-07-06
CCS-3231 Assembly search

benradey:
Makes assemblies show up in module search results.

Forked from #321 so I'm not sure if we can merge this in prior to that one.


pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/ModuleRendererServlet.java
```diff
@@ -0,0 +1,105 @@
...
+
+        moduleVariantSource = module.moduleLocale(localeObj)
+                .traverse()
+                .toChild(ModuleLocale::source)
+                .toChild(draft ? SourceContent::draft : SourceContent::released)
+                .getAsOptional();
+
+
+        if(!moduleVariantSource.isPresent()) {
+            response.sendError(HttpServletResponse.SC_NOT_FOUND, (draft ? "Draft " : "Released ")
```
aprajshekhar, line 87: String.format would be a better option than string concatenation.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/ModuleListingServlet.java
```diff
@@ -209,12 +210,16 @@ protected String getQuery(SlingHttpServletRequest request) {
         SimpleDateFormat sdf = new SimpleDateFormat("MMM dd, yyyy HH:mm");
         
         //logic for file name is present in ModuleVersionUpload.java
-        if(draftMetadata.isPresent() && draftMetadata.get().moduleType().get()!=null){
-            m.put("moduleType",draftMetadata.get().moduleType().get());
-        }else if(releasedMetadata.isPresent() && releasedMetadata.get().moduleType().get()!=null){
-            m.put("moduleType",releasedMetadata.get().moduleType().get());   
-        }else{
-            m.put("moduleType","-");
+        if ("pantheon/assembly".equals(resource.getResourceType())) {
```
aprajshekhar, line 213: The string values can be made into constants. As we are already doing at other places, this will make code consistent.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/AssemblyRenderServlet.java
```diff
@@ -78,41 +77,33 @@ protected void doGet(SlingHttpServletRequest request,
...
+        // collect a list of parameter that traverseFrom with 'ctx_' as those will be used as asciidoctorj
+        // parameters
+        Map<String, Object> context = asciidoctorService.buildContextFromRequest(request);
+
+        // only allow forced rerendering if this is a draft version. Released and historical revs are written in stone.
+        String html = asciidoctorService.getModuleHtml(
+                module, localeObj, variantName, draft, context, reRender && draft);
+
+        response.setContentType("text/html");
+        Writer w = response.getWriter();
```
aprajshekhar, line 104:
```suggestion
        response.getWriter().write(html);
```
would eliminate the need for temporary variable.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/AssemblyRenderServlet.java
```diff
@@ -78,41 +77,33 @@ protected void doGet(SlingHttpServletRequest request,
...
-        Optional<HashableFileResource> moduleVariantSource;
+        Optional<HashableFileResource> moduleVariantSource = null;
 
-        if(draft) {
-            moduleVariantSource = module.moduleLocale(localeObj)
+        moduleVariantSource = module.moduleLocale(localeObj)
                 .traverse()
                 .toChild(ModuleLocale::source)
-                .toChild(SourceContent::draft)
+                .toChild(draft ? SourceContent::draft : SourceContent::released)
```
aprajshekhar, line 85: :thumbsup:

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/ModuleRendererServlet.java
benradey: Better in what sense? String.format is my least favorite option.

If you are talking about "faster," that is not the case. String.format is the slowest option to use.
https://stackoverflow.com/a/2087703
https://redfin.engineering/java-string-concatenation-which-way-is-best-8f590a7d22a8

I would note that in older versions of java, string concatenation was handled poorly and was slow. However in newer versions of java, concatenation uses StringBuilder behind the scenes and is now fast and efficient.

If you have further thoughts then I would welcome the feedback, but otherwise I would prefer to leave this code block as-is.

benradey:
I'm closing this PR as it appears that the relevant code has already been merged into master. I'm not quite sure how this happened, but it must have been part of an offshoot branch that was eventually merged.

PR #339, opened 2020-07-14
Configuration for logging request data

aprajshekhar:
This PR adds configuration that would log request details in <karaf_home>/data/logs/request.log. The changes is to configure Sling's RequestLoggerService[1] to output specific details of the request.

[1] - https://sling.apache.org/documentation/development/client-request-logging.html


carlosmunoz (approved):
Approving. This targets the request log, what about the error log? do we need something similar?

aprajshekhar:
> Approving. This targets the request log, what about the error log? do we need something similar?
Thanks.

The Karaf log already has error log. I was planning to add thies even before karaf migration. However, Karaf migration just made it easier.

PR #340, opened 2020-07-14
add github builds

carlosmunoz:
This should add functionality to build using github. It replicates the exact same build process we have on travis with the addition of an exception when the changes happen solely on the `docs` folder.


carlosmunoz:
@aprajshekhar we'd need to remove the "mandatory" flag for travis builds if we want to merge this in

.github/workflows/maven.yml
```diff
@@ -0,0 +1,61 @@
...
+    - name: Cache maven dependencies
+      uses: actions/cache@v2
+      with:
+        path: ~/.m2
+        key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+        restore-keys: |
+          ${{ runner.os }}-maven-
+    
+    - name: Build with Maven
+      run: ./mvnw clean install
```
aprajshekhar, line 39: Providing  modules' list for building would make knowing errors, if encountered, easy to trace. 
```suggestion
      run: ./mvnw clean install -pl pantheon-bundle,pantheon-karaf-feature,pantheon-karaf-dist
```

carlosmunoz: I'm fine with this change, as long as we understand that if we ever add new maven modules to the larger project they will be ignored by the builder.

aprajshekhar: If we ever add a new module this would have to be updated as well. I am ok with clean install. Only thought was if a new module is added, then its easily understandable from the workflow file.

PR #344, opened 2020-07-21
update role name pant:abstract

xdavidson:
This PR update the role name for pant:abstract. On metadata node, the abstract is stored as pant:abstract attribute


carlosmunoz:
The code looks good but there is a failing unit test.

xdavidson:
> The code looks good but there is a failing unit test.

Failing test is fixed.

PR #349, opened 2020-07-22
add mixin type to help with fast querying of documents

carlosmunoz:
Added the `pant:document` mixin type to help with searches as Sling has a very performant index for primary types and mixin types.


carlosmunoz:
Added the `pant:document` mixin type to help with searches as Sling has a very performant index for primary types and mixin types.

PR #350, opened 2020-07-23
CCS-3716: Add more logging for golang app

aprajshekhar:
This PR:

1. Logs  server, directory, and user being used for calling uploader in GoApp.
2. Checks whether error is null or not before logging error in GoApp.
3. Logs the response received from Pantheon in uploader so that GoApp can log the actual response text from pantheon rather than reason (OK/Internal Server Error etc.)

### DIFF
--- a/tools/git2pantheon/git_uploader.go
+++ b/tools/git2pantheon/git_uploader.go
@@ -33,7 +33,7 @@ func gitClone(repository string, branch string, directory string) {
 
 func getUploader() {
 	if _, err := os.Stat("./pantheon.py"); os.IsNotExist(err) {
-		const uploader_url = "https://raw.githubusercontent.com/redhataccess/pantheon/master/uploader/pantheon.py"
+		const uploader_url = "https://raw.githubusercontent.com/aprajshekhar/pantheon/ap_request_data_log/uploader/pantheon.py"
 		args := []string{"-o", "./pantheon.py", uploader_url}
 		cmd := exec.Command("curl", args...)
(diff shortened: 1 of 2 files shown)


tools/git2pantheon/git_uploader.go
```diff
@@ -33,7 +33,7 @@ func gitClone(repository string, branch string, directory string) {
 
 func getUploader() {
 	if _, err := os.Stat("./pantheon.py"); os.IsNotExist(err) {
-		const uploader_url = "https://raw.githubusercontent.com/redhataccess/pantheon/master/uploader/pantheon.py"
+		const uploader_url = "https://raw.githubusercontent.com/aprajshekhar/pantheon/ap_request_data_log/uploader/pantheon.py"
```
carlosmunoz, line 36: Is this change intentional?

carlosmunoz: Your latest commit answered my question 😄

aprajshekhar: It was unintentional. Thanks for pointing out. Pushed commit fixing it.

xdavidson:
AP, great work on improving the logging. Can you also update the _print_response in createVariant method? It's important see some logging around variant creation.

aprajshekhar:
> AP, great work on improving the logging. Can you also update the _print_response in createVariant method? It's important see some logging around variant creation.

Lisa, thanks. I have updated the _print_response in createVariant method.

aprajshekhar:
Lisa, thanks :)

PR #351, opened 2020-07-23
Adds Assembly classes to our Sling Model for future development work

benradey:


pantheon-bundle/src/main/java/com/redhat/pantheon/asciidoctor/extension/MetadataExtractorTreeProcessor.java
```diff
@@ -136,18 +137,24 @@ private void extractAbstract(List<StructuralNode> allNodes) {
      */
     private void extractModuleType(Document document) {
         Object attValue = document.getAttribute(MODULE_TYPE_ATT_NAME);
+        ModuleMetadata mm = null;
+        if (documentMetadata instanceof ModuleMetadata) {
```
carlosmunoz, line 141: I dislike `instanceof` 😢 
I wonder if extracting the module type is something we should externalize to a different asciidoctor extension. One that specializes in `Module`s. That way we build the extension set based on the type of document that we need to generate.

pantheon-bundle/src/main/java/com/redhat/pantheon/model/assembly/AssemblyVariant.java
```diff
@@ -0,0 +1,36 @@
...
+
+    Child<AssemblyVersion> draft();
+
+    Child<AssemblyVersion> released();
+
+    @Override
+    AssemblyVariants getParent();
+
+    // TODO Not sure we need this
+    default AssemblyVersion getOrCreateVersion(String name) {
```
carlosmunoz, line 22: I don't think we need this.

pantheon-bundle/src/main/java/com/redhat/pantheon/model/assembly/AssemblyVariant.java
```diff
@@ -0,0 +1,36 @@
...
+        return child(name, AssemblyVersion.class).get();
+    }
+
+    // TODO Not sure we need this
+    default AssemblyVersion createNextVersion() {
```
carlosmunoz, line 27: I don't think we need this.

pantheon-bundle/src/main/java/com/redhat/pantheon/model/assembly/AssemblyVariants.java
```diff
@@ -0,0 +1,36 @@
...
+
+    // The default variant
+    @Named(DEFAULT_VARIANT_NAME)
+    Child<AssemblyVariant> defaultVariant();
+
+    default Stream<AssemblyVariant> getVariants() {
+        return this.as(AssemblyVariant.class);
+    }
+
+    default AssemblyVariant getOrCreateVariant(String name) {
```
carlosmunoz, line 26: this method is not needed either.

pantheon-bundle/src/main/java/com/redhat/pantheon/asciidoctor/extension/MetadataExtractorTreeProcessor.java
benradey: Per our gchat conversation, I feel that this is a strength of the code, not a weakness. I would like to leave it as-is for now. We could revisit the decision later if things get more complicated.

pantheon-bundle/src/main/java/com/redhat/pantheon/model/assembly/AssemblyVariant.java
benradey: Removed.

pantheon-bundle/src/main/java/com/redhat/pantheon/model/assembly/AssemblyVariant.java
benradey: Removed.

pantheon-bundle/src/main/java/com/redhat/pantheon/model/assembly/AssemblyVariants.java
benradey: Removed.

PR #353, opened 2020-07-27
CCS-3718: Intermittent 500 Server error for module variants while uploading using git import

aprajshekhar:
This PR adds logic to:

- Check whether variant and workspace has been created successful.
- If create variant or workspace have not succeeded, then exit with error message. Since, creation of workspace and variant are the first operations, if it does not succeed, uploader would exit.

### DIFF
--- a/uploader/pantheon.py
+++ b/uploader/pantheon.py
@@ -344,7 +344,7 @@ def process_workspace(path):
 
     else:
         data = {'DEFAULT': {}}
-    createVariant(data, path, url, workspace)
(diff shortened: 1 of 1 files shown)


aprajshekhar:
> I like the fact we do error checking on process_workspace. It would be nice to return the actual error of failure in addition to "Variants could not be created...."

Thanks. I had not done that because we are already logging the complete error response just before the exiting. However, I like the idea that you have proposed. Hence, made the changes.

PR #354, opened 2020-07-27
CCS-3717: Updates to frontend build

wesruv:
This branch is to allow the React dev process to work without rebuilding any Karaf or Java resources. This enables the webpack server compile JUST the react/frontend components on when they're saved.

This also resolves CORS errors by running setting up a proxy with Spandx.

Testing instructions:
* Have Karaf built/working
* Follow "Getting Started" instructions from: `pantheon-bundle/frontend/README.md`
* Make a change in a file in React, confirm that dev server reloads browser and adds change
* Make sure I didn't mess up any existing way that you all are loading/using the React frontend


README.md
```diff
@@ -22,9 +22,9 @@ and built on top of Apache sling.
 
 ### Prerequisites
 
-* You must have an account on GitHub.  
+* You must have an account on GitHub.
```
benradey, line 25: Thanks, invisible whitespace bugs me. :)

README.md
```diff
@@ -182,27 +182,26 @@ This will install the code in this project on the running Sling instance, where
...
 ```sh
 echo fs.inotify.max_user_watches=65535 | sudo tee -a /etc/sysctl.conf && sudo sysctl -p
 ```
 
-```sh
-chromium-browser --disable-web-security --user-data-dir=/home/user/anywhere/chromeDev/ &
-```
-
-This works because there is code in app.tsx that preempts all fetch() calls and checks if the app is being served from localhost. If so, it modifies the request to point to localhost:8080 specifically, rather than localhost:9000 which is where webpack-dev-server serves the frontend code from.
-
```
benradey, line 205: Glad you got rid of this, it always felt like a hack.

pantheon-bundle/frontend/src/app/app.tsx
```diff
@@ -27,12 +27,16 @@ class App extends Component<any, IAppState> {
...
     const realFetch = self.fetch
     self.fetch = (input: RequestInfo, init?: RequestInit | undefined) => {
-      let newInput = input.toString()
-      if (window.location.host.startsWith('localhost') && input.toString().startsWith('/')) {
+      let newInput = input.toString();
+      if (
+        window.location.host.startsWith('localhost')
+        // Only changing requests Sling's old port
+        && parseInt(window.location.port) === 8080
+        && input.toString().startsWith('/')) {
```
benradey, line 35: I do not recall why we originally needed this... are you sure it is still necessary? Did you try and remove it completely?

(I'm the one who wrote this method in the first place... I don't recall if it was for a frontend development workaround, or if it was so that our yarn tests would pass... if only I had included some comments about this...)

pantheon-bundle/frontend/src/app/app.tsx
```diff
@@ -27,12 +27,16 @@ class App extends Component<any, IAppState> {
...
         newInput = 'http://localhost:8181' + input.toString()
       }
       console.log('Development fetch', input, '=>', newInput)
-      return realFetch(newInput, init)
+      return realFetch(newInput, init);
```
benradey, line 39: :exploding_head:

README.md
wesruv: Agree!! I just realized my editor hadn't been removing trailing white space for me this week.

Thanks!

pantheon-bundle/frontend/src/app/app.tsx
wesruv: All this JS without colons makes me itchy :joy:

pantheon-bundle/frontend/src/app/app.tsx
wesruv: It's required if you want it to work inside of Karaf. I'm not sure how all that stuff works, so I thought it'd be fine to leave it.

What's really weird is that the code is trying to reach `:8080` in Karaf, when it should be `:8181`, if we fix that then we can definitely get rid of this code. No idea what's causing it.

wesruv:
> @wesruv This looks good and works on my local machine. I notice that our tests are now failing, so please fix that. Also, I left one comment for you w.r.t. the realFetch method, let me know if you have any insight on that.
> 
> Otherwise, thanks for this nice contribution!

Rut roh, I'll take a look at the tests.

I fixed a lot of test-related dependencies that weren't working previously, so there's a chance that the tests should have never been passing, but aren't now because the dependencies are working :grimacing:

pantheon-bundle/frontend/src/app/app.tsx
benradey: I did some quick testing without the code and it doesn't look like it makes any difference anymore - so please take this opportunity to get rid of it.

pantheon-bundle/frontend/spandx.config.js
```diff
@@ -0,0 +1,47 @@
...
+    "/system": {
+      host: {
+        local: "http://localhost:8181"
+      }
+    },
+    "/pantheon": {
+      host: {
+        local: "http://localhost:8181"
+      }
+    }
```
benradey, line 40: One thing I noticed while I was playing with this is this - while the application 99% works in "hot frontend" mode, the login screen does not. The login request results in a 404. I'm wondering if we need to add an additional route here to account for the login api endpoint?

Here's a screenshot that might help:
image

wesruv: Ahh, yes! I can capture that login request to make sure it goes to Karaf.

Good catch!

pantheon-bundle/frontend/src/app/productDetails.tsx
```diff
@@ -28,17 +28,17 @@ class ProductDetails extends Component<IProps, IState> {
         }
     }
 
-    public componentWillReceiveProps(nextProps) {
+    public componentDidUpdate() {
         // allow page load from productDetails to products listing
-        if (nextProps.productName !== undefined && nextProps.productName.trim() !== '') {
+        if (this.props.productName !== undefined && this.props.productName.trim() !== '') {
```
benradey, line 33: @wesruv what is the thought behind this change? I believe this causes the user to be unable to actually navigate into product details page. I'm not sure why - I just noticed that clicking the 'product details' widget just takes you nowhere now.

benradey:
@wesruv This looks great, login/logout works great and it looks like you addressed my other comments as well. Awesome! I did just notice one bit of breakage related to product details, which I made a comment on, so please add your perspective on that. Thanks!

pantheon-bundle/frontend/src/app/productDetails.tsx
wesruv: BLARG.

`componentWillReceiveProps ` is deprecated, I think it might have caused one of the test failures, but maybe I was mistaken. I'll try and revert it and rerun the tests.

See: https://github.com/redhataccess/pantheon/pull/354/files/b2e686dfa25a3ee42126782e0c3b18841ceae66e#diff-275ba574586207026ee0741897867bdbR117

It's definitely something we'll have to address at some point.

wesruv: Looks like it wasn't breaking the tests, but it does throw an error.

benradey: Yeah, it's been awhile, but I vaguely recall being the person who wrote that and needing to specifically hunt down just the right method to make the logic work. Sucks that it's going away, I guess we'll cross that bridge when we come to it.

PR #355, opened 2020-07-28
improve management of resources during uploads

carlosmunoz:
a. Use InputStreams when dealing with content
b. restore the checksum validation to prevent unnecesary load on the JCR repository if there are no changes

### DIFF
--- a/pantheon-bundle/src/main/java/com/redhat/pantheon/jcr/JcrResources.java
+++ b/pantheon-bundle/src/main/java/com/redhat/pantheon/jcr/JcrResources.java
@@ -4,11 +4,14 @@ package com.redhat.pantheon.jcr;
 import com.google.common.base.Charsets;
 import com.google.common.hash.HashCode;
 import com.google.common.hash.Hashing;
+import com.google.common.hash.HashingInputStream;
 import org.apache.sling.api.resource.Resource;
 import org.apache.sling.api.resource.ResourceResolver;
 
 import javax.jcr.RepositoryException;
 import javax.jcr.Session;
+import java.io.IOException;
+import java.io.InputStream;
 
 public class JcrResources {
     private JcrResources() {
@@ -20,11 +23,24 @@ public class JcrResources {
         session.move(resource.getPath(), resource.getParent().getPath() + "/" + newName);
     }
 
-    /*
-     * calculates a hash for a string
+    /**
+     * calculates a checksum hash for a string using the adler32 algorithm
      *
      */
     public static HashCode hash(String str) {
         return Hashing.adler32().hashString(str == null ? "" : str, Charsets.UTF_8);
+    }
+
+    /**
+     * Calculates a checksum hash for the contents for an input stream usng the adler32 algorithm.
+     * @param is The input stream. The stream is NOT closed after reading from it.
+     * @return A {@link HashCode} for the contents of the given InputStream
+     * @throws IOException If there is a problem accessing the contents of the InputStream
+     */
+    public static HashCode hash(InputStream is) throws IOException {
+        try (HashingInputStream his = new HashingInputStream(Hashing.adler32(), is)) {
+            while (his.read() != -1) ;
+            return his.hash();
+        }
     }
 }
--- a/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/assembly/AssemblyVersionUpload.java
+++ b/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/assembly/AssemblyVersionUpload.java
@@ -2,6 +2,7 @@ package com.redhat.pantheon.servlet.assembly;
 
 import com.google.common.hash.HashCode;
 import com.redhat.pantheon.conf.GlobalConfig;
+import com.redhat.pantheon.jcr.JcrResources;
 import com.redhat.pantheon.model.api.SlingModels;
 import com.redhat.pantheon.model.assembly.Assembly;
 import com.redhat.pantheon.model.assembly.AssemblyLocale;
@@ -81,7 +82,7 @@ public class AssemblyVersionUpload extends AbstractPostOperation {
                     .draft().getOrCreate();
 
             // Check if the content is the same as what is hashed already
-            HashCode incomingSrcHash = hash(asciidocContent);
+            HashCode incomingSrcHash = JcrResources.hash(asciidocContent);
             String storedSrcHash = draftSrc.hash().get();
             // If the source content is the same, don't update it
             if (incomingSrcHash.toString().equals(storedSrcHash)) {
--- a/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/ModuleVersionUpload.java
+++ b/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/ModuleVersionUpload.java
@@ -3,11 +3,12 @@ package com.redhat.pantheon.servlet.module;
 import com.google.common.hash.HashCode;
 import com.redhat.pantheon.asciidoctor.AsciidoctorService;
 import com.redhat.pantheon.conf.GlobalConfig;
+import com.redhat.pantheon.jcr.JcrResources;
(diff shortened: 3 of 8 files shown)


pantheon-bundle/src/main/java/com/redhat/pantheon/jcr/JcrResources.java
```diff
@@ -20,11 +23,24 @@ public static void rename(Resource resource, String newName) throws RepositoryEx
...
+
+    /**
+     * Calculates a checksum hash for the contents for an input stream usng the adler32 algorithm.
+     * @param is The input stream. The stream is NOT closed after reading from it.
+     * @return A {@link HashCode} for the contents of the given InputStream
+     * @throws IOException If there is a problem accessing the contents of the InputStream
+     */
+    public static HashCode hash(InputStream is) throws IOException {
+        try (HashingInputStream his = new HashingInputStream(Hashing.adler32(), is)) {
+            while (his.read() != -1) ;
```
benradey, line 42: This is interesting. I suppose this is to load the entire stream into memory as a prerequisite to hash. I'm curious if this would necessitate a call to Stream.reset() later or something?

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/assembly/AssemblyVersionUpload.java
```diff
@@ -81,7 +82,7 @@ protected void doRun(SlingHttpServletRequest request, PostResponse response, Lis
                     .draft().getOrCreate();
 
             // Check if the content is the same as what is hashed already
-            HashCode incomingSrcHash = hash(asciidocContent);
+            HashCode incomingSrcHash = JcrResources.hash(asciidocContent);
```
benradey, line 85: Interesting that you chose to leave `asciidocContent` as a String here, rather than reworking it to be an InputStream instead. Was that intentional?

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/ModuleVersionUpload.java
```diff
@@ -106,14 +105,29 @@ protected void doRun(SlingHttpServletRequest request, PostResponse response, Lis
...
                 responseCode = HttpServletResponse.SC_NOT_MODIFIED;
             } else {
-                draftSrc.jcrContent().getOrCreate()
-                        .jcrData().set(asciidocContent);
+                ServletUtils.handleParamAsStream(request, "asciidoc", encoding,
+                        inputStream -> {
+                            Session session = resolver.adaptTo(Session.class);
+                            draftSrc.jcrContent().getOrCreate()
+                                    .jcrData().toFieldType(InputStream.class)
+                                    .set(inputStream);
```
benradey, line 127: Isn't this fancy! What's going on here? Is this a magic way to convert the Stream to a String so that it can be persisted in the jcr?

Also, are you sure this works? My concern is that the Stream has already been consumed by the hashing method, and nothing ever reset it?

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/assembly/AssemblyVersionUpload.java
carlosmunoz: The change to use a single checksum hash method is intentional, but I focused on the module uploading initially and not the assembly upload. Perhaps that is the next step after this is reviewed.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/ModuleVersionUpload.java
carlosmunoz: It's not that magical (you can see that it's just calling some JCR utilities), but it is exactly for the purpose of what you described.

It does work (at least tested locally), but now that you ask the question, the input stream is not being closed anywhere so perhaps there is still a memory leak here.

pantheon-bundle/src/main/java/com/redhat/pantheon/jcr/JcrResources.java
carlosmunoz: The use of a `HashingInputStream` is to **avoid** loading the entire content into memory; it hashes as it consumes.

But you bring up a good point: the stream is potentially useless after it has been consumed. The way we are using it right now is safe as it is being closed, but a comment in the javadoc wouldn't hurt.

carlosmunoz: Never mind... I did have a comment there already 😊

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/ModuleVersionUpload.java
carlosmunoz: reviewed again... there are no problems because the input stream is being passed as the argument to the function, and it's being closed in the method after being used up by the function. All good 😊

pantheon-bundle/src/main/java/com/redhat/pantheon/jcr/JcrResources.java
benradey: > The use of a HashingInputStream is to avoid loading the entire content into memory; it hashes as it consumes.

Well, I understand that's the goal, but `while (his.read() != -1) ;` makes me think that that's not what's happening. The stream reads until there is nothing left, thereby loading everything into memory, *then* it hashes. Is that not what's happening?

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/ModuleVersionUpload.java
benradey: Can you point me to "the function" that you are referring to? I am genuinely curious how this works. AFAIK, Sling has to resolve this stream to some base type in order to be able to persist it, but how does it do that? It has no hint in this code that the stream here represents a string. It could be a stream of bytes, or anything, right? Did *we* write the method that resolves this and *we* are making that assumption? Or is something else going on?

pantheon-bundle/src/main/java/com/redhat/pantheon/jcr/JcrResources.java
carlosmunoz: It does read the whole stream, but not into memory. It hashes as it reads, modifying the hash as it consumes the contents. The end result is the same as if the entire content were hashed.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/ModuleVersionUpload.java
carlosmunoz: You are correct, the input stream can be of anything, and because we are 'setting' an InputStream into a jcr field this results in jackrabbit slurping the contents of the stream into the node's field. 
Now we (as in the caller) determines what stream is being passed to the 'set' method, in this case it's a character stream (UTF-8 or whatever encoding we decided on) directly from the sling request. 
Because the content is being pushed to the sling API as an `nt:file` sling smartly creates a temporary file where it stores the content, and in practice this is the stream that we are getting.

pantheon-bundle/src/main/java/com/redhat/pantheon/jcr/JcrResources.java
benradey: Ok, so you are asserting that the hashing actually happens in the `read()` method, and that the subsequent call to `hash()` doesn't actually do any work, it just returns the precomputed hash. If that's the case, then I'm good with it.

carlosmunoz: correct. 
But this is not because of the HashingInputStream only, it also depends on the hashing function behind it, in this case adler32, which happens to operate in that way you described (confirmed by stepping through it).

PR #356, opened 2020-07-29
PnT JIRA (CCS-3683) unable to publish module when logged in as a demo…

rednitish:
Fix to provided access to users other than admin, to publish the modules.

### DIFF
--- a/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/PublishDraftVersion.java
+++ b/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/PublishDraftVersion.java
@@ -86,7 +86,7 @@ public class PublishDraftVersion extends AbstractPostOperation {
         }
         log.debug("Operation Publishinging draft version,  completed");
         long elapseTime = System.currentTimeMillis() - startTime;
-        log.info("Total elapsed http request/response time in milliseconds: " + elapseTime);
+        log.debug("Total elapsed http request/response time in milliseconds: " + elapseTime);
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/PublishDraftVersion.java
```diff
@@ -75,64 +84,74 @@ public void run(SlingHttpServletRequest request, PostResponse response, SlingPos
             asciidoctorService.getModuleHtml(module, locale, variant, false, new HashMap(), true);
             events.fireEvent(new ModuleVersionPublishedEvent(moduleVersion), 15);
         }
+        log.debug("Operation Publishinging draft version,  completed");
+        long elapseTime = System.currentTimeMillis() - startTime;
+        log.info("Total elapsed http request/response time in milliseconds: " + elapseTime);
```
aprajshekhar, line 89: Do we need this as info rather than debug? Just curious.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/PublishDraftVersion.java
```diff
@@ -75,64 +84,74 @@ public void run(SlingHttpServletRequest request, PostResponse response, SlingPos
...
-            Optional<HashableFileResource> releasedSource = traverseFrom(module)
-                    .toChild(m -> module.locale(locale))
-                    .toChild(ModuleLocale::source)
-                    .toChild(sourceContent -> sourceContent.released())
-                    .getAsOptional();
-            if (draftSource != null) {
-                if (releasedSource.isPresent()) {
+            // Get the draft version, there should be one
+            Optional<ModuleVersion> versionToRelease = module.getDraftVersion(locale, variant);
+            if (!versionToRelease.isPresent()) {
```
aprajshekhar, line 102: Rather than if/else if,  if and return  would be better as you don't want the process to go further after setting response.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/UnpublishVersion.java
```diff
@@ -81,52 +91,62 @@ public void run(SlingHttpServletRequest request, PostResponse response, SlingPos
...
+            Module module = serviceResourceResolver.getResource(request.getResource().getPath()).adaptTo(Module.class);
+            Locale locale = getLocale(request);
+            String variant = getVariant(request);
 
+            // Get the released version, there should be one
+            Optional<? extends DocumentVersion> foundVariant = module.getReleasedVersion(locale, variant);
+
+            if(!foundVariant.isPresent()) {
+                response.setStatus(HttpServletResponse.SC_PRECONDITION_FAILED,
+                        "The module is not released (published)");
```
aprajshekhar, line 112: Return would be better as processing need not to run after setting the status.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/PublishDraftVersion.java
rednitish: Changed level of logger to debug,

benradey:
Code looks great - I didn't test this, but I presume it functions correctly. Nitish, thanks for taking the time to rework the solution! :)

PR #357, opened 2020-07-30
Assembly relationships + table of contents

benradey:


pantheon-bundle/src/main/resources/SLING-INF/nodetypes/nodetypes.cnd
```diff
@@ -84,6 +84,9 @@
 [pant:assemblyVersion] > nt:unstructured, sling:Resource, mix:referenceable, mix:created, mix:lastModified
     - sling:resourceType (string) = 'pantheon/assemblyVersion' mandatory autocreated
 
+[pant:assemblyPage] > nt:unstructured, sling:Resource, mix:referenceable, mix:created, mix:lastModified
```
xdavidson, line 87: what's the purpose of having pant:assemblyPage?

benradey: As you know, long-term we are going to want to break up large assemblies. For now, we've agreed the assemblies are small enough to not need it, but the thought is to future-proof the data structure. So all of our assemblies today will just have "one page" (the entire assembly). But it will be stored in such a way that we can have future multi-page assemblies and our data structure will not need to change to accommodate it.

Does that answer the question or were you trying to ask something else?

pantheon-bundle/src/main/java/com/redhat/pantheon/model/api/WorkspaceChild.java
```diff
@@ -7,7 +7,7 @@
 
     default Workspace getWorkspace() {
         Resource workspace = this.getParent();
-        while (workspace != null && !workspace.getResourceType().equals("pantheon/workspace")) {
+        while (workspace != null && !"pantheon/workspace".equals(workspace.getResourceType())) {
```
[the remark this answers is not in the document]
benradey: I don't recall the exact reason why but I thiiiiink I was trying to write unit tests and I was having trouble with this line throwing an NPE.

Generally, `stringVar.equals("someConstant")` is dangerous, but `"someConstant".equals(stringVar)` is equivalent and safer.

PR #362, opened 2020-08-06
CCS-3661: Json API for assembly

aprajshekhar:
This PR adds:

- URL for assembly data
- Response for assembly data identified by assembly uuid and variant 

Sample response has been attached to the corresponding JIRA ticket.

### DIFF
--- a/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/assembly/AssemblyJsonServlet.java
+++ b/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/assembly/AssemblyJsonServlet.java
@@ -1,0 +1,183 @@
+package com.redhat.pantheon.servlet.assembly;
+
+import com.google.common.base.Charsets;
+import com.redhat.pantheon.html.Html;
+import com.redhat.pantheon.model.ProductVersion;
+import com.redhat.pantheon.model.api.FileResource;
+import com.redhat.pantheon.model.assembly.Assembly;
+import com.redhat.pantheon.model.assembly.AssemblyMetadata;
+import com.redhat.pantheon.model.assembly.AssemblyVersion;
+import com.redhat.pantheon.servlet.AbstractJsonSingleQueryServlet;
+import com.redhat.pantheon.servlet.ServletUtils;
+import org.apache.sling.api.SlingHttpServletRequest;
+import org.apache.sling.api.resource.Resource;
+import org.apache.sling.servlets.annotations.SlingServletPaths;
+import org.jetbrains.annotations.NotNull;
+import org.osgi.framework.Constants;
+import org.osgi.service.component.annotations.Component;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nonnull;
+import javax.jcr.RepositoryException;
+import javax.servlet.Servlet;
+import java.util.*;
+
+import static com.redhat.pantheon.conf.GlobalConfig.DEFAULT_MODULE_LOCALE;
+import static com.redhat.pantheon.model.assembly.AssemblyVariant.DEFAULT_VARIANT_NAME;
+import static com.redhat.pantheon.servlet.ServletUtils.paramValue;
+import static com.redhat.pantheon.servlet.ServletUtils.paramValueAsLocale;
+import static javax.servlet.http.HttpServletResponse.SC_OK;
+
+/**
+ * Get operation to render a Released Assembly data in JSON format.
+ * Only two parameters are expected in the Get request:
+ * 1. locale - Optional; indicates the locale that the assembly content is in, defaulted to en-US
+ * 2. assembly_id - indicates the uuid string which uniquely identifies an assembly
+ *
+ * The url to GET a request from the server is /api/module
+ * Example: <server_url>/api/assembly?locale=en-us&module_id=xyz&variant=abc
+ * The said url is accessible outside of the system without any authentication.
+ *
+ * @author A.P. Rajjshekhar
+ */
+@Component(
+        service = Servlet.class,
+        property = {
+                Constants.SERVICE_DESCRIPTION + "=Servlet to facilitate GET operation which accepts locale and assembly uuid to output module data",
+                Constants.SERVICE_VENDOR + "=Red Hat Content Tooling team"
+        })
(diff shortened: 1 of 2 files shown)


pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/assembly/AssemblyJsonServlet.java
```diff
@@ -0,0 +1,183 @@
...
+        //assemblyMap.put("description", releasedMetadata.get().description().get());
+        assemblyMap.put("content_type", "assembly");
+        assemblyMap.put("date_published", releasedMetadata.get().getValueMap().containsKey("pant:datePublished") ? releasedMetadata.get().datePublished().get().toInstant().toString() : "");
+        assemblyMap.put("status", "published");
+
+        // Assume the path is something like: /content/<something>/my/resource/path
+        assemblyMap.put("assembly_url_fragment", resourcePath.substring("/content/repositories/".length()));
+
+        // Striping out the jcr: from key name
+        String module_uuid = (String) assemblyMap.remove("jcr:uuid");
```
xdavidson, line 115: can we rename module_uuid to assembly_uuid?

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/assembly/AssemblyJsonServlet.java
```diff
@@ -0,0 +1,183 @@
...
+        assemblyMap.put("assembly_uuid", module_uuid);
```
xdavidson, line 116: same as above

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/assembly/AssemblyJsonServlet.java
```diff
@@ -0,0 +1,183 @@
...
+        // Convert date string to UTC
+        Date dateModified = new Date(resource.getResourceMetadata().getModificationTime());
+        assemblyMap.put("date_modified", dateModified.toInstant().toString());
+        // Return the body content of the module ONLY
```
xdavidson, line 120: please replace "module" with "assembly"

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/assembly/AssemblyJsonServlet.java
```diff
@@ -0,0 +1,183 @@
...
+        if (searchKeywords != null && !searchKeywords.isEmpty()) {
+            assemblyMap.put(SEARCH_KEYWORDS, searchKeywords.split(", *"));
+        }
+        else {
+            assemblyMap.put(SEARCH_KEYWORDS, new String[] {});
+        }
+
+        // Process view_uri
+        if (System.getenv(PORTAL_URL) != null) {
+            String view_uri = System.getenv(PORTAL_URL) + "/topics/" + ServletUtils.toLanguageTag(locale) + "/" + module_uuid;
```
xdavidson, line 163: please rename module_uuid to assembly_uuid

aprajshekhar:
Thanks!. This is interesting. I had done these changes before reverse merge with master. Will make the changes

pantheon-bundle/src/test/java/com/redhat/pantheon/servlet/assembly/AssemblyJsonServletTest.java
```diff
@@ -0,0 +1,141 @@
...
+        assertDoesNotThrow(() -> slingContext.resourceResolver().queryResources(query, Query.JCR_SQL2));
+    }
+
+    @Test
+    void getQuery() {
+        // Given
+        AssemblyJsonServlet servlet = new AssemblyJsonServlet();
+        Map<String, Object> map = newHashMap();
+        map.put("locale", ServletUtils.toLanguageTag(Locale.US));
+        map.put("module_id", "jcr:uuid");
```
xdavidson, line 51: i wonder if we should call it something more generic so that both module api and assembly api can share the same name. Hydra team maps this field to id field in solr. The fact is this is assembly_id not module_id. Both are uuid of a node. we could call it "uuid", which will make the mapping more straightforward for hydra team.

aprajshekhar: That is a good idea. So, rename (assembly, module)_uuid to just uuid in the response only or also in the url as well?

xdavidson: I would rename in both the response body. I am not sure which url you refer to. 
    the url to module api looks like this: https://<HOST>/api/module/variant.json/b537ef3c-5c7d-4280-91ce-e7e818e6cc11
    the url to assembly api looks like this: https://<HOST>/api/assembly.json/bc3ae91d-fbf1-4dc5-9bec-edfaa03d29fa

xdavidson: if you do decide to rename line 51 to uuid instead of "module_id", it would be great if you can add this to VariantJsonServlet.java:
       // TODO: remove module_uuid after Hydra team releases UNIFIED-6570
        variantMap.put("module_uuid", variant_uuid);
       // add the following to the map for hdyra to consume
        variantMap.put("uuid", variant_uuid);

aprajshekhar: Ah.. ok, my bad regarding the URI. I am fine with renaming in assembly but module response is being used by  teams other than Hydra. So, it could introduce bugs. Hence, I am hesitant and not in favor of making change  to module response in the scope of this JIRA/PR. Maybe we can have another JIRA for that where other teams can be notified?

xdavidson: yes, that works for me.

aprajshekhar: I have updated the test case accordingly. Will raise another JIRA for common/generic name for uuid

PR #369, opened 2020-08-12
 (CCS-3550) Publishing or republishing an assembly

rednitish:
(CCS-3550) Publishing or republishing an assembly

If a Publisher publishes or republishes an assembly:

    The assembly draft is published.
    If an assembly cannot be published (ex: missing required metadata), then show a warning and do not publish the assembly.
    Pantheon will tell Hydra to publish the assembly from search.


pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/PublishDraftVersion.java
```diff
@@ -114,46 +129,145 @@ protected void doRun(SlingHttpServletRequest request, PostResponse response, Lis
...
+                || versionToRelease.get().metadata().getOrCreate().productVersion().get().isEmpty()) {
+            // Check if productVersion is set
+            response.setStatus(HttpServletResponse.SC_PRECONDITION_FAILED,
+                    "The version to be released doesn't have productVersion metadata");
+            return ;
+        } else if (versionToRelease.get().metadata().getOrCreate().urlFragment().get() == null
+                || versionToRelease.get().metadata().getOrCreate().urlFragment().get().isEmpty()) {
+            // Check if urlFragment is set
+            response.setStatus(HttpServletResponse.SC_PRECONDITION_FAILED,
+                    "The version to be released doesn't have urlFragment metadata");
```
xdavidson, line 179: urlFragment is an optional field

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/PublishDraftVersion.java
```diff
@@ -114,46 +129,145 @@ protected void doRun(SlingHttpServletRequest request, PostResponse response, Lis
...
                 // Check if productVersion is set
                 response.setStatus(HttpServletResponse.SC_PRECONDITION_FAILED,
                         "The version to be released doesn't have productVersion metadata");
-                return;
+                return ;
             } else if (versionToRelease.get().metadata().getOrCreate().urlFragment().get() == null
                     || versionToRelease.get().metadata().getOrCreate().urlFragment().get().isEmpty()) {
                 // Check if urlFragment is set
                 response.setStatus(HttpServletResponse.SC_PRECONDITION_FAILED,
                         "The version to be released doesn't have urlFragment metadata");
```
xdavidson, line 250: urlFragment is an optional field

pantheon-bundle/frontend/src/app/assemblyDisplay.tsx
```diff
@@ -143,6 +144,7 @@ class AssemblyDisplay extends Component<any, any, any> {
                         <Card>
                             <Versions
                                 modulePath={this.state.modulePath}
+                                type="assembly"
```
xdavidson, line 147: please create an entry for assembly in pantheon-bundle/frontend/src/app/Constants.tsx

pantheon-bundle/frontend/src/app/moduleDisplay.tsx
```diff
@@ -143,6 +144,7 @@ class ModuleDisplay extends Component<any, any, any> {
                         <Card>
                             <Versions
                                 modulePath={this.state.modulePath}
+                                type="module"
```
xdavidson, line 147: please create an entry for module in pantheon-bundle/frontend/src/app/Constants.tsx

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/PublishDraftVersion.java
rednitish: @xdavidson, urlFragment is not a part of this ticket.

rednitish:
@xdavidson  I have resolved comments given by you, Can you please take a look at it again.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/PublishDraftVersion.java
xdavidson: Agreed. But it would be super awesome if you can remove the validation on requiring urlFragment.

pantheon-bundle/src/main/java/com/redhat/pantheon/extension/events/assembly/AssemblyVersionPublishStateEvent.java
```diff
@@ -0,0 +1,25 @@
...
+
+    private final String moduleVersionPath;
+
+    protected AssemblyVersionPublishStateEvent(@Nonnull AssemblyVersion assemblyVersion) {
+        this.moduleVersionPath = assemblyVersion.getPath();
+    }
+
+    public String getModuleVersionPath() {
+        return moduleVersionPath;
+    }
```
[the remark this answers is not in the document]
rednitish: @benradey  There is a separate ticket to handle hydra messaging part, I added this class, as it had a dependency on changes.

pantheon-bundle/src/main/java/com/redhat/pantheon/extension/HydraIntegration.java
```diff
@@ -96,7 +93,7 @@ public boolean canProcessEvent(Event event) {
     }
 
     /**
-     * Process ModuleVersionPublishedEvent. It sends a simple text message to the Message Broker.
+     * Process AssemblyVersionPublishedEvent. It sends a simple text message to the Message Broker.
```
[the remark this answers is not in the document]
rednitish: @benradey please check below answer

pantheon-bundle/src/main/java/com/redhat/pantheon/extension/events/assembly/AssemblyVersionPublishStateEvent.java
```diff
@@ -0,0 +1,24 @@
...
+import javax.annotation.Nonnull;
+
+/**
+ * Event fired when a assembly version has been published.
+ * Includes the assembly version path so it can be re-fetched in the
+ * handlers if necessary.
+ */
+public class AssemblyVersionPublishStateEvent implements Event {
+
+    private final String moduleVersionPath;
```
benradey, line 15: Still need to rename this to Assembly (vs module)

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/PublishDraftVersion.java
```diff
@@ -114,46 +120,40 @@ protected void doRun(SlingHttpServletRequest request, PostResponse response, Lis
...
-                response.setStatus(HttpServletResponse.SC_PRECONDITION_FAILED,
-                        "The version to be released doesn't have urlFragment metadata");
-                return;
             } else {
                 // Draft becomes the new released version
-                ModuleVariant moduleVariant = traverseFrom(module)
-                        .toChild(m -> module.locale(locale))
-                        .toChild(ModuleLocale::variants)
+                DocumentVariant docVariant = traverseFrom(document.get())
+                        .toChild(m -> document.get().locale(locale))
```
benradey, line 141: @rednitish I realize you did not write this code originally, but now that I'm looking at it, I see that the variable `m` here is just unused. That has me wondering if the initial call to `traverseFrom` is unnecessary, or if this technique isn't being used in the wrong way.

@carlosmunoz Since you're the mastermind behind most of the model api, can you comment on whether this is the best way to get from Document down to ultimately DocumentVariant? Am I right in thinking that the discarding of `m` is a code smell? And what's the difference between using ResourceTraversal vs using the model api directly, such as `document.get().locale(whatever).get().variants().get().variant("variant").get()`?

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/PublishDraftVersion.java
```diff
@@ -114,46 +120,40 @@ protected void doRun(SlingHttpServletRequest request, PostResponse response, Lis
             if(canPublish) {
                 serviceResourceResolver = serviceResourceResolverProvider.getServiceResourceResolver();
             }
-            Module module = serviceResourceResolver.getResource(request.getResource().getPath()).adaptTo(Module.class);
+            Optional<? extends Document> document = getDocument(request, serviceResourceResolver);
```
benradey, line 123: You shouldn't need a call to a new `getDocument()` method here, you should just be able to use the original line and just replace `Module` with `Document`.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/PublishDraftVersion.java
```diff
@@ -179,4 +179,14 @@ protected void doRun(SlingHttpServletRequest request, PostResponse response, Lis
...
+
+    @NotNull
+    private Optional<? extends Document> getDocument(SlingHttpServletRequest request, ResourceResolver serviceResourceResolver) {
+        if(serviceResourceResolver.getResource(request.getResource().getPath()).adaptTo(Document.class).getResourceType().equalsIgnoreCase("pantheon/assembly")){
+            Assembly assembly = serviceResourceResolver.getResource(request.getResource().getPath()).adaptTo(Assembly.class);
+            return Optional.ofNullable(assembly);
+        }
+            Module module = serviceResourceResolver.getResource(request.getResource().getPath()).adaptTo(Module.class);
+            return Optional.ofNullable(module);
+    }
```
benradey, line 191: I have a bunch of thoughts about this method.

- It doesn't make a whole lot of sense to use both `@NotNull` and the `Optional` construct together. The whole point of Optional is to wrap something that may be null, so if your method returns an Optional, then of course the reference to the Optional itself is never going to be null (even if the thing that it wraps might be). Having said that, Sling makes sure that your servlet is only called when your Resource actually exists, so we don't need to worry about a null check. If I weren't about to recommend that you just remove this method, then I would recommend that you just drop both constructs because it's not worth worrying about.
- The `if ()` statement here is a bit long. If I were writing this method, here's how I'd do it instead:
```
Resource r = serviceResourceResolver.getResource(request.getResource().getPath());
if ("pantheon/assembly".equals(r.getResourceType()) {
    return r.adaptTo(Assembly.class);
} else {
    return r.adaptTo(Module.class);
}
```
- However, you really don't need the method at all. Just call `.adaptTo(Document.class)` on line 123 and use that for the entire servlet. Since you need to know what type of event to fire on line 97, just check the resource type there. The rest of the servlet can be completely generic.
- If you get rid of the unnecessary Optional construct, you don't need `.get()` cluttering up the code.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/PublishDraftVersion.java
```diff
@@ -80,16 +81,21 @@ public void run(SlingHttpServletRequest request, PostResponse response, SlingPos
...
+            asciidoctorService.getModuleHtml(document.get(), locale, variant, false, new HashMap(), true);
 
-            // Regenerate the module once more
-            asciidoctorService.getModuleHtml(module, locale, variant, false, new HashMap(), true);
-            events.fireEvent(new ModuleVersionPublishedEvent(moduleVersion), 15);
+            if(documentVersion.get() instanceof AssemblyVersion){
+                events.fireEvent(new AssemblyVersionPublishedEvent((AssemblyVersion) documentVersion.get()), 15);
+            }else{
+                events.fireEvent(new ModuleVersionPublishedEvent((ModuleVersion)documentVersion.get()), 15);
+            }
```
benradey, line 98: This is the only place where you need to know whether it's assembly vs module. I would not use `instanceof` here - rather, I would just use `if ("pantheon/assemblyVersion".equals(documentVersion.getResourceType())` or something similar. Let me know if you don't understand why, we can talk about the difference on chat.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/UnpublishVersion.java
```diff
@@ -88,15 +91,19 @@ public void run(SlingHttpServletRequest request, PostResponse response, SlingPos
...
-                    .draft().get();
+                    .draft().get());
 
             // TODO We need to change the event so that the right variant is processed
-            events.fireEvent(new ModuleVersionUnpublishedEvent(moduleVersion), 15);
+            if(documentVersion.get() instanceof AssemblyVersion){
+                events.fireEvent(new AssemblyVersionPublishedEvent((AssemblyVersion) documentVersion.get()), 15);
+            }else{
+                events.fireEvent(new ModuleVersionPublishedEvent((ModuleVersion)documentVersion.get()), 15);
+            }
```
benradey, line 106: Same comment here as publish servlet.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/UnpublishVersion.java
```diff
@@ -88,15 +91,19 @@ public void run(SlingHttpServletRequest request, PostResponse response, SlingPos
         if (response.getError() == null) {
             // call the extension point
             Locale locale = getLocale(request);
-            Module module = getModule(request);
+            Optional<? extends Document> document = getDocument(request, request.getResourceResolver());
             String variant = getVariant(request);
-            ModuleVersion moduleVersion = module.locale(locale).get()
+            Optional<? extends DocumentVersion> documentVersion = Optional.of(document.get().locale(locale).get()
```
benradey, line 96: Same comments here, I am betting this `Optional` construct is unnecessary.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/UnpublishVersion.java
```diff
@@ -121,32 +128,41 @@ protected void doRun(SlingHttpServletRequest request, PostResponse response, Lis
...
+            Optional<? extends Document> document;
+            if(serviceResourceResolver.getResource(request.getResource().getPath()).adaptTo(Document.class).getResourceType().equalsIgnoreCase("pantheon/assembly")){
+                Assembly assembly = serviceResourceResolver.getResource(request.getResource().getPath()).adaptTo(Assembly.class);
+                document = Optional.ofNullable(assembly);
+            }
+            else{
+                Module module = serviceResourceResolver.getResource(request.getResource().getPath()).adaptTo(Module.class);
+                document = Optional.ofNullable(module);
+
+            }
```
benradey, line 140: Again, unnecessary, you should just be able to `.adaptTo(Document.class)` and the code should still function.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/UnpublishVersion.java
```diff
@@ -173,4 +189,13 @@ protected void doRun(SlingHttpServletRequest request, PostResponse response, Lis
...
     }
+    @NotNull
+    private Optional<? extends Document> getDocument(SlingHttpServletRequest request, ResourceResolver serviceResourceResolver) {
+        if(serviceResourceResolver.getResource(request.getResource().getPath()).adaptTo(Document.class).getResourceType().equalsIgnoreCase("pantheon/assembly")){
+            Assembly assembly = serviceResourceResolver.getResource(request.getResource().getPath()).adaptTo(Assembly.class);
+            return Optional.ofNullable(assembly);
+        }
+        Module module = serviceResourceResolver.getResource(request.getResource().getPath()).adaptTo(Module.class);
+        return Optional.ofNullable(module);
+    }
```
benradey, line 200: Same comment as publish servlet, please remove this method.

pantheon-bundle/src/main/java/com/redhat/pantheon/extension/events/assembly/AssemblyVersionPublishStateEvent.java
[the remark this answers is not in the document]
benradey: Ok, I am not sure I understand. Can you please explain why the field names we use in Java has an impact on the JSON payload that we product for Hydra to consume? Theoretically we should be able to rename this variable to `private final String bigYellowBanana` and Hydra should not care... am I mistaken?

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/UnpublishVersion.java
rednitish: @benradey we have below method structure for method getReleasedVersion(), so DocumentVersion instance wont be able to call below method in case we remove optional construct.

default Optional<? extends DocumentVersion> getReleasedVersion(@Nonnull final Locale locale,@Nonnull final String variantName) {

pantheon-bundle/src/main/java/com/redhat/pantheon/asciidoctor/AsciidoctorService.java
```diff
@@ -176,10 +177,10 @@ private String buildDocument(@Nonnull Document base, @Nonnull Locale locale, @No
                     + ",variant: " + variantName + ", draft: " + isDraft);
         }
 
-        // Use a service-level resource resolver to build the module as it will require write access to the resources
+        // Use a service-level resource resolver to build the module or assemblies as it will require write access to the resources
         try (ResourceResolver serviceResourceResolver = serviceResourceResolverProvider.getServiceResourceResolver()) {
 
-            Class cls = base instanceof Module ? Module.class : Assembly.class;
+            Class cls = base.getResourceType().equals(PantheonConstants.RESOURCETYPE_ASSEMBLY) ? Assembly.class : Module.class;
             Document serviceDocument = (Document) SlingModels.getModel(serviceResourceResolver, base.getPath(), cls);
```
benradey, line 184: This is interesting. It looks like we're putting effort into figuring out whether the resource is a module or an assembly, but then we're ultimately discarding the answer and just casting to a generic document class for the rest of the method. I bet this could be removed.

I know you didn't write this originally, but since it looks like you have a conflict to resolve anyway, bonus points for taking care of this at the same time. :)

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/PublishDraftVersion.java
```diff
@@ -60,16 +63,16 @@ public PublishDraftVersion(@Reference Events events,
...
     }
 
     private Locale getLocale(SlingHttpServletRequest request) {
         return paramValueAsLocale(request, "locale", GlobalConfig.DEFAULT_MODULE_LOCALE);
     }
 
     private String getVariant(SlingHttpServletRequest request) {
-        return paramValue(request, "variant", ModuleVariant.DEFAULT_VARIANT_NAME);
+        return paramValue(request, "variant", DocumentVariant.DEFAULT_VARIANT_NAME);
     }
```
benradey, line 76: Not your code, but I worry that we have a bug here, so also tagging @carlosmunoz @xdavidson @aprajshekhar.

I don't think it's appropriate to assume that if the caller doesn't specify a variant parameter for publishing, that we assume DEFAULT_VARIANT_NAME.

Better options:
- Assume `document.getWorkspace().getCanonicalVariantName()`
- Throw an exception, because if the caller is publishing something, they should be explicit about which variant they want rather than us potentially guessing wrong.

Thoughts / other opinions?

benradey: For the record, the same method appears in UnpublishVersion.java as well.

aprajshekhar: @benradey since this logic appears at multiple files, I would say let's table it till MS3 is done. Creating a JIRA to track this is what I would suggest and lets discuss the options there. Discussing it here would need updating the JIRA and that would be a overhead. Just my two cents.

pantheon-bundle/src/main/java/com/redhat/pantheon/asciidoctor/AsciidoctorService.java
rednitish: @benradey  I have created a separate ticket, to track this code quality issue, we can work on it separately.
https://projects.engineering.redhat.com/browse/CCS-3772

PR #371, opened 2020-08-13
Basic Xref functionality

benradey:
If the author is writing moduleA.adoc and includes text like `xref:moduleB.adoc[label]`, and moduleB.adoc is present in Pantheon at build time, then Pantheon will now rewrite the xref to `xref:ABCD1234-EFGH-ETC[label]` automatically, which will ultimately become a link to `<a href="ABCD1234-EFGH-ETC.html">label</a>`.

This PR also adds UUID-based preview servlets that allow these links to resolve. If you start a preview session at `/pantheon/preview/latest/MODULE-A-UUID`, and follow an xref to module B, that should resolve successfully via the new servlets.

Our React UI will need to be refactored to use these new servlets, and triangle-bracket xrefs are not yet handled.


pantheon-bundle/src/main/java/com/redhat/pantheon/asciidoctor/extension/SlingResourceIncludeProcessor.java
```diff
@@ -79,9 +79,13 @@ public void process(Document document, PreprocessorReader reader, String target,
                         .toChild(FileResource::jcrContent)
                         .toField(FileResource.JcrContent::jcrData)
                         .get();
-                content = new StringBuilder(":pantheon_module_id: ")
+                content = new StringBuilder()
+                        .append(":pantheon_module_id: ")
```
xdavidson, line 83: I wonder if we should change pantheon_module_id to something more generic? :pantheon_document_id?

benradey: That's probably a solid idea but we'll have to bring @wesruv in to comment on that because it would have downstream impact.

pantheon-bundle/src/main/java/com/redhat/pantheon/asciidoctor/extension/XrefPreprocessor.java
```diff
@@ -0,0 +1,122 @@
...
+        reader.restoreLines(preprocess(reader.readLines()));
+    }
+
+    private String processLineWithPattern(String line, Pattern pattern) {
+        Matcher matcher = pattern.matcher(line);
+        StringBuffer sb = new StringBuffer();
+        while (matcher.find()) {
+            String originalTarget = matcher.group("filepath");
+            // Assume it's a relative path to a file in the same repo for now
+            Resource desiredTarget = documentVariant.getResourceResolver().getResource(documentVariant.getParentLocale().getParent().getParent().getPath() + "/" + originalTarget);
```
xdavidson, line 54: if would be super if we can add some line wrap here to make it more readable :-)

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/DocumentVariantRenderServlet.java
```diff
@@ -0,0 +1,119 @@
...
+        extensions = "preview")
+public class DocumentVariantRenderServlet extends SlingSafeMethodsServlet {
+
+    private static final Set<String> RELEASED_SUFFIXES = new HashSet<>();
+    static {
+        RELEASED_SUFFIXES.add("/released");
+        RELEASED_SUFFIXES.add("/");
+        RELEASED_SUFFIXES.add("");
+        RELEASED_SUFFIXES.add(null);
+    }
```
xdavidson, line 57: are those allowed values for RELEASED_SUFFICES?

pantheon-bundle/src/main/java/com/redhat/pantheon/asciidoctor/extension/SlingResourceIncludeProcessor.java
wesruv: I'm not picky, if it's not specific to modules, I'm for it. Better to get in now than later.

benradey: Sorry! I don't know what I was thinking when I commented yesterday, but yes this *is* module-specific. So let's keep it as-is for now.

(I saw "for now" because when we support assembly-of-assembly, that may change... or it may not... but we'll see when we get there.)

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/DocumentVariantRenderServlet.java
benradey: Yes, the idea here is that the suffix is optional. If you don't specify it, we assume you want the released version. This covers all the cases: [document, document/, document/released]

pantheon-bundle/src/main/java/com/redhat/pantheon/asciidoctor/extension/XrefPreprocessor.java
benradey: Okie dokie, I broke up the line a bit with the latest commit.

PR #373, opened 2020-08-18
UI change for assebmly details and Patternfly/React upgrade

xdavidson:
  *  Upgrade Patternfly 3.16 to 4.32
  *  Introduce the new UI for assembly
  * Update module details screen


pantheon-bundle/frontend/src/app/moduleDisplay.tsx
```diff
@@ -335,6 +360,31 @@ class ModuleDisplay extends Component<any, any, any> {
...
+            .then(responseJSON => {
+                if (responseJSON["pant:attributesFilePath"] !== undefined) {
+                    this.setState({ attributesFilePath: responseJSON["pant:attributesFilePath"] })
+                }
+            })
+            .catch((error) => {
+                console.log(error)
+            })
+    }
+
```
benradey, line 387: Tagging this at the bottom, but this entire file looks substantially similar to the assemblyDisplay.tsx file. Can we combine them to reduce duplication?

pantheon-bundle/frontend/src/app/versions.test.tsx
```diff
@@ -304,39 +283,39 @@ describe('Versions tests', () => {
...
+    //     sinon.assert.called(spy)
+    // })
+
+    // it('test changePublishState function', () => {
+    //     const wrapper = renderer.create(<Versions {...props} />)
+    //     const inst = wrapper.getInstance()
+    //     const spy = sinon.spy(inst, 'changePublishState')
+    //     inst.changePublishState("ds")
+    //     sinon.assert.called(spy)
+    // })
```
benradey, line 320: It looks like you've commented out this entire file. Was that intentional? Will the commented code be coming back at some point?

IIRC, versions.tsx is responsible for displaying most of the document details pages. I would guess that you've reworked most of that and that the old tests are now obsolete. If that's true, please remove them rather than commenting them.

xdavidson: versions.test.tsx has been updated.

pantheon-bundle/frontend/src/app/moduleDisplay.tsx
xdavidson: I agree we can use a shared component for assemblyDisplay and moduleDisplay. I would like it to be a separate ticket.  This is just the initial work of introducing assembly on the UI.

benradey: Ok, we'll agree that this is tech debt and we'll bring in this PR as-is.

PR #375, opened 2020-08-20
CCS-3762: Error in git2pantheonpod while gitimport: http: superfluous  response.WriteHeader

aprajshekhar:
This PR:

1. Adds the logic to return the response if parsing fails with correct code (400) withe detail rather than just 500.
2. Fixes the http: superfluous  response.WriteHeader.
3. Changes the error response given by go app  to {"error": "<error detail>"}. This is to make responses standardized.

### DIFF
--- a/tools/git2pantheon/main.go
+++ b/tools/git2pantheon/main.go
@@ -33,20 +33,30 @@ func cloneBranch(w http.ResponseWriter, r *http.Request) {
 		var repo repository_and_branch
 		err = json.Unmarshal(body, &repo)
 
-		log.Println(repo.Repo)
-		log.Println(repo.Branch)
+		//check parsing error. If error is not null, log and return 400 if parse failed
+		if err != nil {
+			log.Printf("Error in unmarshalling request body %s", err)
+			//setting the response in JSON format for standardizing error responses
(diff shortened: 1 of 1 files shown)


tools/git2pantheon/main.go
```diff
@@ -33,20 +33,30 @@ func cloneBranch(w http.ResponseWriter, r *http.Request) {
...
 		if repository == "" {
-			http.Error(w, "Error reading repository url", http.StatusInternalServerError)
-			error_status = true
+			//setting the response in JSON format for standardizing error responses
+			http.Error(w, "{\"error\" : \"Error reading repository url\"}", http.StatusInternalServerError)
+			// return after setting error response to avoid superfluous error header in logs
+			return
 		}
 		if !strings.Contains(strings.ToUpper(repository), "GIT") {
-			http.Error(w, "The repository entered does not look like a git repo.", http.StatusInternalServerError)
```
xdavidson, line 46: I recall we currently only support git repo.

aprajshekhar: I do not have context about why this check is there. The main focus of this PR is to add error handling, providing response for each error that is caught, and make the error response standard. Hence, I did not change the check. Anyways we are going to rewrite the git2pantheon later, we can make discuss it in a JIRA  related to that. Sounds good?

PR #376, opened 2020-08-20
print filepath on 304 response code

xdavidson:
Currently filepath is not printed out as part of standard output when Uploader receives a http response code  304. This PR allows us to see the filepath on stdout for 304.


aprajshekhar:
@xdavidson it will be great if you could add the same change to https://github.com/redhataccess/pantheon-uploader as well. Thanks in advance.

PR #380, opened 2020-08-24
CCS-3661: Json API for assembly. Adding variant based API

aprajshekhar:
This PR provides:

1. Variant based URI for Assembly - /api/assembly/variant.json/${variantUuid}
2. The data is same as that available on the corresponding JIRA.


aprajshekhar:
@xdavidson thanks. Please approve.

PR #381, opened 2020-08-24
Tweak to haml template

benradey:
Tweak from Wes that enables anchor hyperlinks to function with webcomponents.


benradey:
Closing since @wesruv says this is no longer needed.

PR #382, opened 2020-08-25
rename assembly_uuid and module_uuid to uuid

carlosmunoz:
This change introduces a new `uuid` field in the module and assembly APIs which will contain the same value as the `module_uuid` and `assembly_uuid` respectively.
The previous fields are left in for backwards compatibility but are now deprecated.


carlosmunoz:
> the changes looks good for assembly APIs. I think we also need update
> pantheon/servlet/module/VariantJsonServlet.java

Good catch, I have committed the change to include this. Please review.

PR #383, opened 2020-08-25
Making 'variant' required for publish servlets

benradey:


pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/PublishDraftVersion.java
```diff
@@ -72,12 +73,17 @@ private Locale getLocale(SlingHttpServletRequest request) {
     }
 
     private String getVariant(SlingHttpServletRequest request) {
-        return paramValue(request, "variant", DocumentVariant.DEFAULT_VARIANT_NAME);
+        return paramValue(request, "variant", null);
```
rednitish, line 76: @benradey returning null directly in code, is not a good practice.

Also, as per our discussion on JCR changes, we are under impression, in case of no variant, it should create a default or return a default variant.
@xdavidson  your thoughts ?

Also, As a coding standard we should move, our hard coded variables in Constant Class.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/UnpublishVersion.java
```diff
@@ -82,12 +83,17 @@ private Locale getLocale(SlingHttpServletRequest request) {
     }
 
     private String getVariant(SlingHttpServletRequest request) {
-        return paramValue(request, "variant", DocumentVariant.DEFAULT_VARIANT_NAME);
+        return paramValue(request, "variant", null);
```
rednitish, line 86: same comment as above

rednitish (commented):
@benradey I have few concerns, in this PR, let me know, if you disagree, we can have a discussion on it.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/PublishDraftVersion.java
benradey: Hi Nitish,

Can you clarify what you mean by "returning null is not a good practice?" That is a vague statement and in this scenario, returning null is exactly the appropriate thing to do.

Also, my argument is exactly that returning a default variant is inappropriate. Publishing is a very important operation, and if the user does not specify which variant they are trying to publish, then it's wrong and dangerous to think we can correctly guess what they want. It is better to throw an error and force the caller to be explicit about what they want.

rednitish: @benradey 

If we are to pass a null in this line in place of "DocumentVariant.DEFAULT_VARIANT_NAME", com.redhat.pantheon.servlet.ServletUtils.paramValue will return null directly.

The rationale behind not returning null is code should not follow a different path based on the return value. There is a Design pattern "" Null Object Pattern "" which provides more information on this.

if I were to do the same, I would typically prefer to return an empty collection  or String rather than null as it means my client code is cleaner.

Regarding my second query tagging  @xdavidson   to provide some more inputs.

benradey: Thanks for clarifying, Nitish.

I think I understand what you're describing, but I have two thoughts.

1st and more relevant: the code is not any different and/or cleaner if we replace a null check with an empty string check. It is practically identical.

2nd and less relevant, but still related: saying that something is a "good design pattern" or "bad design pattern" generally may be true, but that doesn't mean it's the right or wrong tool for the job in any specific circumstance. I would agree that if we were in some hypothetical example where we were building a collection, then returning an empty collection would be more appropriate than returning null, and the pattern you refer to might apply nicely to that. However, the situation we have here is different. We are asking for input that the user supplied, and the value of `null` has the special meaning of "the user did not supply anything" which would be distinct from "the user supplied a value of nothing."

Ultimately, this is splitting hairs. Avoiding the use of null does not gain us anything tangible here, so I'm not in favor of making that code change.

rednitish: Thanks Ben & Thanks again for putting forward your thoughts, But i still strongly disagree with point 1 you mentioned.
 and yes, we opt for different design patterns according to our requirement, and I can see a lot of points where we should include them. I will tag them all going forwards and lets have a cleaner code going forward :)

benradey: @rednitish I understand that you strongly disagree, but I still maintain that your request is not functionally different from what is already here and that this code is correct as-is.

If you have no other concerns, then I would ask that you approve this PR with the code as I have it.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/PublishDraftVersion.java
```diff
@@ -72,12 +73,17 @@ private Locale getLocale(SlingHttpServletRequest request) {
...
     }
 
     @Override
     public void run(SlingHttpServletRequest request, PostResponse response, SlingPostProcessor[] processors) {
         logger.debug("Operation Publishinging draft version started");
+        String variant = getVariant(request);
```
rednitish, line 82: @benradey  Can you please add a test case for this block in PublishDraftVersionTest.java

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/UnpublishVersion.java
```diff
@@ -82,12 +83,17 @@ private Locale getLocale(SlingHttpServletRequest request) {
...
     }
 
     @Override
     public void run(SlingHttpServletRequest request, PostResponse response, SlingPostProcessor[] processors) {
         logger.debug("Operation UnPublishinging draft version started");
+        String variant = getVariant(request);
+        if (variant == null) {
```
rednitish, line 93: add test case for below piece of code in UnpublishVersionTest.java

rednitish (changes requested):
@benradey  I saw the coverage got dropped a little due to changes, would request you to add a case to cover the added condition.

rednitish (approved):
> @benradey I saw the coverage got dropped a little due to changes, would request you to add a case to cover the added condition.

As per my discussion with Lisa, This fix is a tech debt and not a priority for now, we can fix it later stage, so approving this PR.

PR #392, opened 2020-08-28
Fixes to leveloffset code

benradey:
This PR works around a bug in Asciidoctor involving the following scenario:
- The author has set a document-wide leveloffset using :leveloffset: X
- The author has also set an include-specific leveloffset using include:whatever.adoc[leveloffset=+Y]
- A preprocessor is involved in the asciidoctor build process

When all of the above are true, Asciidoctor does not resolve leveloffset correctly. It attempts to resolve the local leveloffset by temporarily overwriting the document-wide value, but fails to reset the document-wide value to its correct value at the conclusion of the include.

The code contained in this PR works around that issue.


xdavidson (approved):
test locally. it works for me.

PR #393, opened 2020-08-30
eliminate the ResourceTraversal class

carlosmunoz:
This change simplifies the mechanism for navigating through a model structure in a null-safe manner. It basically removes the `ResourceTraversal` class and instead uses the existing `Child` interface to accommodate the navigation functionality. It also removes the deprecated `map` method, which was the first attempt to offer a similar mechanism.

Most of the changes are about adapting to the new, simpler API. The core changes can be found in the following classes:
`Child`
`Field`
`NullObjects`
`SlingModels`


carlosmunoz:
@benradey I picked this up from a while ago. I've rebased and made improvements. Please have a look and let me know what you think.

pantheon-bundle/src/main/java/com/redhat/pantheon/model/api/Child.java
```diff
@@ -37,28 +41,92 @@
...
+     * @return A {@link Child} (may be non-existent) as indicated by the accessor.
      */
-    @Deprecated
-    <R> Optional<R> map(Function<? super T, ? extends R> func);
+    default <R extends SlingModel> Child<R> toChild(Function<? super T, Child<R>> childAccessor) {
+        if(isPresent()) {
+            return childAccessor.apply(get());
+        }
+        return (Child<R>) NullObjects.nullChild();
+    }
```
benradey, line 68: I think this warrants some further explanation. It's easy to think that the `isPresent()` call applies to the child, when I think this must apply to the parent.

I realized this when I was trying to get to the bottom of why your changes help to make things null-safe. For example, if you have this chain:
```
Child.from(A)
.toChild(B)
.toChild(C)
.toChild(D)
.toChild(E)
.getAsOptional()
````
...and let's say that you get a null on C, then what the heck stops us from getting an NPE on the call to D?

I guess the answer must be that there's some code that makes sure that once C becomes a NullChild, then the call to get D must also return a NullChild, and the same for the call down into E. The code that I have attached this comment to must be what does that. But like I said, it's easy to misread at first glance, so if I'm right then I'd like to request *something* that makes this look less innocuous and more critically important (which I believe it is). Maybe some enhanced javadoc, or some `// INLINE COMMENTS IN ALL CAPS` or something - whatever you think would be most fitting.

**Review by benradey — CHANGES_REQUESTED** [review]
@carlosmunoz - looks good - this was hard to follow at first, I think I sorted through the logic, but I did have one comment that I'd like your thoughts on. Thanks!

File: pantheon-bundle/src/main/java/com/redhat/pantheon/model/api/Child.java (hunk 1 above)
Comment (carlosmunoz) [reply pantheon-bundle/src/main/java/com/redhat/pantheon/model/api/Child.java:68]: You are correct in your interpretation. This code is applying the Null Object pattern (https://en.wikipedia.org/wiki/Null_object_pattern) to ensure the null-safety. So let me make a few points which may clarify your questions:

1. In line 64 above, the call to `isPresent()` is done on the parent (aka the current node in the navigation). Perhaps it would be more self-explanatory to replace that with `this.isPresent()`.
2. Line 67 above is where the "magic" begins. If the current node is **not present**, then any child will inevitably not exist either, and so a `NullChild` is returned every time. 
3. The code never throws an NPE because it **never** returns a `null`, instead it returns a `NullChild` which has some specific behaviours to represent the fact that it is an empty node.

Happy to add more context and javadocs, just let me know where so I can do it in the right places where things were not clear. It's hard to identify those when you're the author.

**Comment by carlosmunoz** [comment]
@benradey ping

File: pantheon-bundle/src/main/java/com/redhat/pantheon/model/api/Child.java [hunk 2]
```diff
@@ -37,28 +41,92 @@
...
-     * @param func The function to apply to the value
+     * Convert this Child to an {@link Optional}
+     * @return An {@link Optional} with the contained value.
+     */
+    default Optional<T> asOptional() {
+        return Optional.ofNullable(get());
+    }
+
+    /**
+     * Navigates to the {@link Child} provided by an accessor function.
```
benradey, line 57:
```suggestion
     * Navigates to the {@link Child} provided by an accessor function.
     * Uses the Null Object pattern to avoid throwing an NPE in a long chain of navigation.
     * For example:
     * Child.from(A)
     * .toChild(B)
     * .toChild(C)
     * .toChild(D)
     * .toChild(E)
     * .getAsOptional()
     * If C does not exist, the chain will not throw an exception because the nullChild will be returned for the remainder of the calls.
```

pantheon-bundle/src/main/java/com/redhat/pantheon/model/api/Child.java
```diff
@@ -37,28 +41,92 @@
...
-     * returns null, or if the value of this child was not present in the first place,
-     * this returns an empty Optional
-     * @deprecated Use {@link ResourceTraversal#traverseFrom(SlingModel)}
-     * for safe resource traversals
+     * @return A {@link Child} (may be non-existent) as indicated by the accessor.
      */
-    @Deprecated
-    <R> Optional<R> map(Function<? super T, ? extends R> func);
+    default <R extends SlingModel> Child<R> toChild(Function<? super T, Child<R>> childAccessor) {
+        if(isPresent()) {
```
benradey, line 64:
```suggestion
        if(isPresent()) {  // <-- Applies to the parent (aka the current node in navigation)
```

benradey (approved):
Thanks for the contribution @carlosmunoz - this looks good and I think it will help accelerate our development efforts.

PR #399, opened 2020-09-03
Ccs 3703 schema support

xdavidson:
* On module variant API, you will see "isPartOf" and the assembly/assemblies uuid and url that this module is included in
* On assembly variant API, you will see "hasPart" and the published module(s) info that this assembly includes


aprajshekhar (approved):
Looks good to me. Approved.

PR #402, opened 2020-09-06
split the base model adapter factory and the custom implementation

carlosmunoz:
So that the actual model classes can be separated from the base class. This means the model classes can now be used completely independently from the rest of the source code.


carlosmunoz:
@xdavidson for context, this is so we can use these classes in cupmatic as they are.

xdavidson (approved):

PR #403, opened 2020-09-07
CCS-3785: error on git import when "Repository URL is invalid" needs to change on Pantheon UI

aprajshekhar:
This PR
1. Parses the exception message coming from Git2Pantheon
2. Makes one of the left out error message into standard JSON error format.


rednitish (approved):
LGTM

PR #405, opened 2020-09-08
Reintroducing Wes' tweak to haml for anchors

benradey:
This was supposed to be part of the original xref feature branch but somehow didn't make it into master.


carlosmunoz (approved):
Seems targeted enough

PR #407, opened 2020-09-09
mvn test issue fix

aprajshekhar:
One of the poms was still referring to 0.1.1 rather than 0.1.0 which results in mvn test failing. This PR fixes that.


rednitish (commented):
LGTM

rednitish (approved):

PR #414, opened 2020-09-10
Added a pantheon2.yml file to manage the end-user documentation.

adahms:
Added a Pantheon configuration file so that the end-user documentation team can import the content they are creating into Pantheon and start reviewing it in its intended format.

### DIFF
--- a/pantheon2.yml
+++ b/pantheon2.yml
@@ -1,0 +1,10 @@
+server: https://pantheon.corp.redhat.com
+repository: Pantheon
+variants:
+  - name: main
+    path: pantheon-bundle/src/main/resources/SLING-INF/content/docs/attributes.adoc
+    canonical: true
+assemblies:
+  - pantheon-bundle/src/main/resources/SLING-INF/content/docs/assemblies/*.adoc
+modules:
+  - pantheon-bundle/src/main/resources/SLING-INF/content/docs/modules/*.adoc


adahms:
@redhataccess/pantheon-developers - Please review for technical completeness and accuracy.

benradey:
Interesting idea... I'm curious where you're going with this? :)

Is it just to get the docs into Pantheon so the preview links can be shared, and the process stops there? That's the takeaway I get from your opening comment.

I imagine the intention is not to go "all the way" and publish these docs to the customer portal?

Purely asking for curiosity, not trying to shut this down by any stretch.

adahms:
Hi @benradey - that's exactly it.

Early on in the project, we envisioned there being a custom mechanism of some kind for the EUD team to arbitrarily push updates to the documentation that would appear under the '? > Help' menu in Pantheon V2. 

The requirements for this are laid out in CCS-3630.

In recent months, Dawn told us that engineering would like us to take a dogfooding approach instead - use Pantheon V2 to manage and render the content. In line with this approach, we were told that we would need to create a pantheon2.yml file and product in Pantheon V2 like any other repository, and that once this content was in there, the scope of CCS-3630 would be to link the '? > Help' menu to whatever assembly acted as our main file.

That's the understanding I'm operating under in raising this PR - that engineering asked us to do this, and that this is part of our expected workflow in preparing the content to render ahead of MS3.

If that's not the case, we need to figure out how our content is going to get into Pantheon V2 and be rendered so that we can start adjusting the content to suit that model in time. We're now at the stage where we need to see how it's going to appear before we can make some decisions and get everything ready.

adahms:
Thanks, @benradey - glad this is still the way to go.

Removing the 'WIP' prefix from this PR; you're welcome to merge when engineering feels this is ready.

benradey (approved):

PR #415, opened 2020-09-10
CCS-3790 Product name and version url fragment missing from the view

rednitish:
CCS-3790 Product name and version url fragment missing from the view


aprajshekhar (approved):
LGTM

PR #417, opened 2020-09-11
Ccs 3743 change module type fixes

xdavidson:
This PR updates the label in the sort filter. It changes the label from  "Module type" to "Content type" so that it matches the column name.

### DIFF
--- a/pantheon-bundle/frontend/src/app/__snapshots__/searchFilter.test.tsx.snap
+++ b/pantheon-bundle/frontend/src/app/__snapshots__/searchFilter.test.tsx.snap
@@ -67,9 +67,9 @@ exports[`SearchFilter tests should render default Search component 1`] = `
       />
     </FormSelect>
     <FormSelect
-      aria-label="FormSelect ModuleType"
+      aria-label="FormSelect cType"
       className="small-margin"
-      id="moduleTypeForm"
+      id="contentTypeForm"
       isDisabled={false}
       isRequired={false}
       onBlur={[Function]}
@@ -151,10 +151,10 @@ exports[`SearchFilter tests should render default Search component 1`] = `
       />
       <FormSelectOption
         isDisabled={false}
-        key="Module type"
-        label="Module type"
+        key="Content type"
+        label="Content type"
         required={false}
-        value="Module type"
+        value="Content type"
       />
     </FormSelect>
     <Button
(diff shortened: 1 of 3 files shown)


pantheon-bundle/frontend/src/app/__snapshots__/searchFilter.test.tsx.snap
```diff
@@ -152,7 +152,7 @@ exports[`SearchFilter tests should render default Search component 1`] = `
       <FormSelectOption
         isDisabled={false}
         key="Module type"
-        label="Module type"
+        label="Content type"
         required={false}
         value="Module type"
```
rednitish, line 157: Can we also consider changing other module instances changes within this PR?

pantheon-bundle/frontend/src/app/__snapshots__/searchFilter.test.tsx.snap
```diff
@@ -152,7 +152,7 @@ exports[`SearchFilter tests should render default Search component 1`] = `
       <FormSelectOption
         isDisabled={false}
         key="Module type"
```
rednitish, line 154: Can we also consider changing other module instances changes within this PR?

pantheon-bundle/frontend/src/app/searchFilter.tsx
```diff
@@ -57,7 +57,7 @@ class SearchFilter extends Component<any, any> {
       { value: 'Uploaded date', label: 'Uploaded date', disabled: false },
       { value: 'Title', label: 'Title', disabled: false },
       { value: 'Updated date', label: 'Updated date', disabled: false },
-      { value: 'Module type', label: 'Module type', disabled: false }
+      { value: 'Module type', label: 'Content type', disabled: false }
```
rednitish, line 60: Can we also consider changing other module instances changes within this PR?

rednitish (changes requested):
I have few queries below, let me know, if we have those in separate change list.

pantheon-bundle/frontend/src/app/__snapshots__/searchFilter.test.tsx.snap
xdavidson: that would be a different effort. This ticket I'm working on is weighted as 0.5

pantheon-bundle/frontend/src/app/__snapshots__/searchFilter.test.tsx.snap
xdavidson: changing the key value breaks the sort.

pantheon-bundle/frontend/src/app/searchFilter.tsx
xdavidson: changing the key value breaks the sort. We are going to re-do the search page in the near future. For this ticket, the goal is to just update the label.

rednitish:
LGTM

rednitish (changes requested):
LGTM

rednitish (approved):

PR #419, opened 2020-09-14
CCS-3723: Adding inline JS to enable warning message for inline styles

wesruv:
Since the &lt;head> tag isn't brought into prod just adding JS to add attribute that allows warning
Updated version of cp-documentation in web-assets build


benradey (approved):
It's greek to me, I assume you know what you're doing :)

PR #421, opened 2020-09-16
Ccs 3323 send assembly data

xdavidson:
   * remove deprecated "module_uuid", "variant_uuid" and "assembly_uuid" from module/assembly variant apis. Going forward, we will only have "uuid". Hydra team is aware of this action. They are going to prod with this change tomorrow.
   * Update HydraIntegration to allow messages to be sent to Hydra when we publish/unpublish Assembly


aprajshekhar (approved):
buildEventMessage can be refactored into smaller methods. However, that can be taken up later.
Other than that LGTM :).

Approved.

PR #422, opened 2020-09-18
Pantheon xref URLs are now dynamically rewritten for the customer portal.

benradey:
Links should work for both Pantheon preview (where they are unmodified) and in the Customer Portal (where they are dynamically rewritten in the API as part of building the json payload).

### DIFF
--- a/pantheon-bundle/src/main/java/com/redhat/pantheon/extension/url/CustomerPortalUrlProvider.java
+++ b/pantheon-bundle/src/main/java/com/redhat/pantheon/extension/url/CustomerPortalUrlProvider.java
@@ -1,0 +1,28 @@
+package com.redhat.pantheon.extension.url;
+
+import com.redhat.pantheon.helper.PantheonConstants;
+import com.redhat.pantheon.model.document.DocumentVariant;
+import org.apache.sling.api.resource.Resource;
+import org.apache.sling.api.resource.ResourceResolver;
+
+import java.util.Optional;
+
+public abstract class CustomerPortalUrlProvider implements UrlProvider {
+
+    protected static final String URL_PREFIX = "/documentation/";
+
+    protected String getDocumentType(DocumentVariant variant) {
+        return PantheonConstants.RESOURCE_TYPE_ASSEMBLYVARIANT.equals(variant.getResourceType()) ? "guide" : "topic";
+    }
+
+    protected String getLocale(DocumentVariant variant) {
+        return variant.getParentLocale().getName();
+    }
+
+    public String getHost(ResourceResolver resolver) {
+        Optional<Resource> conf = Optional.ofNullable(resolver.getResource("/conf/pantheon"));
+        return conf.map(Resource::getValueMap)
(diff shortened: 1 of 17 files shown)


pantheon-bundle/src/main/java/com/redhat/pantheon/extension/url/CustomerPortalUrlProvider.java
```diff
@@ -0,0 +1,28 @@
...
+
+    protected String getLocale(DocumentVariant variant) {
+        return variant.getParentLocale().getName();
+    }
+
+    public String getHost(ResourceResolver resolver) {
+        Optional<Resource> conf = Optional.ofNullable(resolver.getResource("/conf/pantheon"));
+        return conf.map(Resource::getValueMap)
+                .map(m -> m.get("pant:portalUrl", String.class))
+                .orElse("");
```
carlosmunoz, line 26: @sfloess you were asking about `Optional` the other day, here's a nice use of it 😃

sfloess: @carlosmunoz - yes!  \o/  This is awesome!  <3   Thanks!!!

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/DocumentCustomerPortalUrlServlet.java
```diff
@@ -0,0 +1,51 @@
...
+        property = {
+                Constants.SERVICE_DESCRIPTION + "=Servlet which provides a Customer Portal URL for any document",
+                Constants.SERVICE_VENDOR + "=Red Hat Content Tooling team"
+        })
+@SlingServletResourceTypes(
+        resourceTypes = { "pantheon/moduleVariant", "pantheon/assemblyVariant", "pantheon/module", "pantheon/assembly" },
+        methods = "GET",
+        selectors = "url",
+        extensions = "txt")
+public class DocumentCustomerPortalUrlServlet extends SlingSafeMethodsServlet {
```
carlosmunoz, line 33: I would try to use more generic names than "Customer Portal". These specific classes should eventually live in a different bundle.

benradey: I partially agree. Someday, these should live in a redhat-specific bundle, and I tried to split the code between "generic definition" and "specific implementation" as much as I could (although this is a servlet so it doesn't apply here). But because that is the ultimate goal, I think it's ok to use specific names in these and other classes that are destined to get broken out from the main codebase. What do you think?

carlosmunoz: Yes, we're not going to split them up now, so as long as we are aware of the future plan I'm good. Perhaps a comment describing the purpose of the class would help remind us of that.

benradey:
Closing this PR since it looks like #425 was based off this and was merged.

PR #424, opened 2020-09-22
CCS-3781: Adding styles for related/additional content

wesruv:
Also added example in ascii-doc-styleguide.html for testing


xdavidson (approved):
LGTM

PR #426, opened 2020-09-24
CCS-3400: page styling for modules

wesruv:
Will have to generate new HTML to see changes, AND may need to clear your caches, to see the update you have to get the newest version of cp-documentation.js from access.redhat.com (which is likely cached).

There should be an underline under the document header, e.g. 
image


xdavidson (approved):
LGTM

PR #427, opened 2020-09-25
Ccs 3768

rednitish:
non ascii character text is getting corrupted on preview of the module"

### DIFF
--- a/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/assembly/AssemblyVersionUpload.java
+++ b/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/assembly/AssemblyVersionUpload.java
@@ -60,13 +60,6 @@ public class AssemblyVersionUpload extends AbstractPostOperation {
 
         try {
             String locale = ServletUtils.paramValue(request, "locale", GlobalConfig.DEFAULT_MODULE_LOCALE.toString());
-//            String contentType = ServletUtils.paramValue(request, "type", "assembly");
-
-            String encoding = request.getCharacterEncoding();
-            if (encoding == null) {
-                encoding = StandardCharsets.UTF_8.name();
-            }
-
             String path = request.getResource().getPath();
 
             log.debug("Pushing new version at: " + path + " with locale: " + locale);
@@ -109,7 +102,7 @@ public class AssemblyVersionUpload extends AbstractPostOperation {
             if (incomingSrcHash.toString().equals(storedSrcHash)) {
                 responseCode = HttpServletResponse.SC_NOT_MODIFIED;
             } else {
-                ServletUtils.handleParamAsStream(request, "asciidoc", encoding,
+                ServletUtils.handleParamAsStream(request, "asciidoc",
                         inputStream -> {
                             Session session = resolver.adaptTo(Session.class);
                             draftSrc.jcrContent().getOrCreate()
--- a/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/ModuleVersionUpload.java
+++ b/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/ModuleVersionUpload.java
@@ -73,7 +73,7 @@ public class ModuleVersionUpload extends AbstractPostOperation {
 
         try {
             String locale = ServletUtils.paramValue(request, "locale", GlobalConfig.DEFAULT_MODULE_LOCALE.toString());
-            String encoding = ServletUtils.paramValue(request, "Content-Type", StandardCharsets.UTF_8.name());
+            String encoding = ServletUtils.paramValue(request, "encoding", StandardCharsets.UTF_8.name());
             String path = request.getResource().getPath();
 
             log.debug("Pushing new module version at: " + path + " with locale: " + locale);
--- a/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/ModuleVersionUpload.java
+++ b/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/ModuleVersionUpload.java
@@ -73,11 +73,7 @@ public class ModuleVersionUpload extends AbstractPostOperation {
 
         try {
             String locale = ServletUtils.paramValue(request, "locale", GlobalConfig.DEFAULT_MODULE_LOCALE.toString());
-            String encoding = request.getCharacterEncoding();
-            if (encoding == null) {
-                encoding = StandardCharsets.UTF_8.name();
-            }
-
+            String encoding = ServletUtils.paramValue(request, "Content-Type", StandardCharsets.UTF_8.name());
             String path = request.getResource().getPath();
 
             log.debug("Pushing new module version at: " + path + " with locale: " + locale);
--- a/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/ModuleVersionUpload.java
+++ b/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/ModuleVersionUpload.java
@@ -73,7 +73,6 @@ public class ModuleVersionUpload extends AbstractPostOperation {
 
         try {
             String locale = ServletUtils.paramValue(request, "locale", GlobalConfig.DEFAULT_MODULE_LOCALE.toString());
-            String encoding = ServletUtils.paramValue(request, "encoding", StandardCharsets.UTF_8.name());
             String path = request.getResource().getPath();
 
             log.debug("Pushing new module version at: " + path + " with locale: " + locale);
@@ -115,7 +114,7 @@ public class ModuleVersionUpload extends AbstractPostOperation {
             if(incomingSrcHash.toString().equals( storedSrcHash )) {
                 responseCode = HttpServletResponse.SC_NOT_MODIFIED;
             } else {
-                ServletUtils.handleParamAsStream(request, "asciidoc", encoding,
+                ServletUtils.handleParamAsStream(request, "asciidoc",
                         inputStream -> {
                             Session session = resolver.adaptTo(Session.class);
                             draftSrc.jcrContent().getOrCreate()
--- a/pantheon-bundle/src/test/java/com/redhat/pantheon/servlet/module/ModuleVersionUploadTest.java
+++ b/pantheon-bundle/src/test/java/com/redhat/pantheon/servlet/module/ModuleVersionUploadTest.java
@@ -89,56 +89,6 @@ class ModuleVersionUploadTest {
     }
 
     @Test
-    void createFirstVersionUnicodeIso() throws Exception {
-        // Given
-        slingContext.build()
-                .resource("/content/repositories/test_workspace",
-                        "jcr:primaryType", "pant:workspace")
-                .commit();
-
-        lenient().when(
-                asciidoctorService.getDocumentHtml(
-                        any(Module.class), any(Locale.class), anyString(), anyBoolean(), anyMap(), anyBoolean()))
-                .thenReturn("A generated html string");
-        registerMockAdapter(Workspace.class, slingContext);
-
-        ModuleVersionUpload upload = new ModuleVersionUpload(asciidoctorService);
-        Map<String, Object> params = newHashMap();
-        params.put("locale", Locale.SIMPLIFIED_CHINESE.toString());
-        params.put("asciidoc", "å\u008D\u0097äº¬é\u0098²ç\u0096«ç\u008E°å\u009Cº");
-        params.put("encoding", StandardCharsets.ISO_8859_1.toString());
-        slingContext.request().setParameterMap(params);
-        slingContext.request().setResource(new NonExistingResource(slingContext.resourceResolver(), "/content/repositories/test_workspace/entities/new/proc_module"));
-        HtmlResponse response = new HtmlResponse();
-
-        // when
-        upload.doRun(slingContext.request(), response, null);
-
-        // Then
-        assertEquals(HttpServletResponse.SC_CREATED, response.getStatusCode());
-
-        Module module =
-                SlingModels.getModel(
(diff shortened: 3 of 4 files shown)


carlosmunoz:
Quick question @rednitish, does this change now require all content to be UTF-8?
If so, what is the reason for abandoning the other encodings?, which pantheon seemed to be currently supporting.

rednitish:
> Quick question @rednitish, does this change now require all content to be UTF-8?
> If so, what is the reason for abandoning the other encodings?, which pantheon seemed to be currently supporting.

@carlosmunoz I was about to discuss it with AP. But thanks for asking.Currently pantheon is supporting ISO-8859-1 which gets picked up by default. but for content in other encoding like utf-8, pantheon expects it in same way.

I couldn't find encoding attribute in adoc files, to identify content type.

so, 1.) either we need to update content writers to add encoding type in case we have encoding other than ASCII
or 2.) we can switch over to utf, for covering content in all formats.

Let me know, your thoughts .

pantheon-bundle/src/test/java/com/redhat/pantheon/servlet/module/ModuleVersionUploadTest.java
```diff
@@ -87,57 +87,6 @@ void createFirstVersion() throws Exception {
         );
         verify(asciidoctorService).getDocumentHtml(any(Module.class), any(Locale.class), anyString(), eq(true), anyMap(), eq(true));
     }
-
-    @Test
```
aprajshekhar, line 91: This test can be used for utf-8 testing.

aprajshekhar:
@carlosmunoz since UTF-8 supports all the charsets, what would be the use case for ISO based charsets?

aprajshekhar:
Once the test case is added back, I can approve the PR.

aprajshekhar:
As discussed, please discuss with Ben and Lisa as well.

pantheon-bundle/src/test/java/com/redhat/pantheon/servlet/module/ModuleVersionUploadTest.java
xdavidson: I see we have a test case for CJK characters using simplified chinese locale, I also see a test case for es_ES locale. If we want to use UTF-8 encoding for all uploads, I don't see a need for testing  a string as ascii content like "å\u008D\u0097äº¬é\u0098²ç\u0096«ç\u008E°å\u009Cº".   As that's not how the writers create their content. 
If we want to be thorough, we could add a test case for accented characters like
    "À È Í Ó Ü Ÿ"

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/assembly/AssemblyVersionUpload.java
```diff
@@ -62,10 +62,7 @@ protected void doRun(SlingHttpServletRequest request, PostResponse response, Lis
             String locale = ServletUtils.paramValue(request, "locale", GlobalConfig.DEFAULT_MODULE_LOCALE.toString());
 //            String contentType = ServletUtils.paramValue(request, "type", "assembly");
 
-            String encoding = request.getCharacterEncoding();
-            if (encoding == null) {
-                encoding = StandardCharsets.UTF_8.name();
-            }
+            String encoding = StandardCharsets.UTF_8.name();
```
benradey, line 65: @rednitish I'm worried that we're breaking support for non-UTF-8 encodings by doing this. Can you share your thought process behind this change?

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/ModuleVersionUpload.java
```diff
@@ -73,10 +73,7 @@ protected void doRun(SlingHttpServletRequest request, PostResponse response, Lis
 
         try {
             String locale = ServletUtils.paramValue(request, "locale", GlobalConfig.DEFAULT_MODULE_LOCALE.toString());
-            String encoding = request.getCharacterEncoding();
-            if (encoding == null) {
-                encoding = StandardCharsets.UTF_8.name();
-            }
+            String encoding = StandardCharsets.UTF_8.name();
```
benradey, line 76: (same comment as above)

pantheon-bundle/src/test/java/com/redhat/pantheon/servlet/module/ModuleVersionUploadTest.java
```diff
@@ -105,10 +105,10 @@ void createFirstVersionUnicodeIso() throws Exception {
         ModuleVersionUpload upload = new ModuleVersionUpload(asciidoctorService);
         Map<String, Object> params = newHashMap();
         params.put("locale", Locale.SIMPLIFIED_CHINESE.toString());
-        params.put("asciidoc", "å\u008D\u0097äº¬é\u0098²ç\u0096«ç\u008E°å\u009Cº");
+        params.put("asciidoc", "d'agua per toles partes, por exemplu: Sicilia y Cuba; al otrudía  decatáronse  d'ú  veníen  les  voces:  de  dientro  la  casa;  nun  foi  aaguantar aquel perru: vendiólu.");
         slingContext.request().setParameterMap(params);
         slingContext.request().setResource(new NonExistingResource(slingContext.resourceResolver(), "/content/repositories/test_workspace/entities/new/proc_module"));
-        slingContext.request().setCharacterEncoding(StandardCharsets.ISO_8859_1.toString());
+        slingContext.request().setCharacterEncoding(StandardCharsets.UTF_8.toString());
```
benradey, line 111: Echoing my previous concern - it's nice that we have a test to confirm that UTF-8 works, but we should also have tests to make sure that other encodings work as well. It seems like with these changes, we're forcing all uploads to use UTF-8 and I don't think that's the correct design decision for the system.

pantheon-bundle/src/test/java/com/redhat/pantheon/servlet/module/ModuleVersionUploadTest.java
benradey: @xdavidson - although I agree with you when you say "authors don't write their content as å\u008D\u0097äº¬é\u0098²ç\u0096«ç\u008E°å\u009Cº" - I *do* think that it is appropriate to test using that format. The reason is that from the author's perspective, they will write "南京防疫现场" in their document, BUT as we have seen in the past, it's quite possible that those characters will be sent using the wrong encoding and the string will get mangled. We need to be sure that sling will unmangle the content correctly, which is the rationale for structuring the test the way it was.

I think that by rewriting the test the way that it is proposed in this PR, we are weakening our assurance that the system handles such scenarios correctly.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/assembly/AssemblyVersionUpload.java
rednitish: @benradey if you check encoding, after running uploader script, you will find, it's always ISO_8859_1, There is no use of this conditional block with current code, as it will never execute.

secondly, if you check com.redhat.pantheon.servlet.ServletUtils#handleParamAsStream(org.apache.sling.api.SlingHttpServletRequest, java.lang.String, java.util.function.Function<java.io.InputStream,R>) , line no 184, you can see, we are parsing content in UTF_8 while storing it in nodes.

Accordingly, i understand, we can go ahead by removing this block, to fix the encoding issue.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/ModuleVersionUpload.java
rednitish: @benradey if you check encoding, after running uploader script, you will find, it's always ISO_8859_1, There is no use of this conditional block with current code, as it will never execute.

secondly, if you check com.redhat.pantheon.servlet.ServletUtils#handleParamAsStream(org.apache.sling.api.SlingHttpServletRequest, java.lang.String, java.util.function.Function<java.io.InputStream,R>) , line no 184, you can see, we are parsing content in UTF_8 while storing it in nodes.

Accordingly, i understand, we can go ahead by removing this block, to fix the encoding issue.

pantheon-bundle/src/test/java/com/redhat/pantheon/servlet/module/ModuleVersionUploadTest.java
rednitish: @benradey Thanks for this comment,
This test case was added for accented characters, we can push negative test case, to have your concern resolved.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/assembly/AssemblyVersionUpload.java
benradey: I think you have struck at the heart of the bug with the lines you reference, but I think that the mechanisms at play are working a bit differently than you describe.

Take a look at how the code used to look in this class:
https://github.com/redhataccess/pantheon/blame/e672d87df1f7f221367aa06fa45daca1f23289ca/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/assembly/AssemblyVersionUpload.java

The relevant line is 53 - it's very clear: "decode the string of bytes that I got from the request using the specified encoding, and give me a string re-encoded in UTF-8 because that's what I want to work with."

As you point out, the specified encoding was always ISO_8859_1 (but still, the 'if' block left the door open for other possible encodings). This code worked, but it was inefficient.

What you're proposing is to assume that the string of bytes coming from the request is encoded in UTF-8 from the start, which I believe is just not true. It "appears" to work but ignoring the specified encoding is unsafe. We really need to rely on the encoding that the request specifies being accurate.

But, you point at line 184 in ServletUtils as treating the input as UTF-8 by default. Honestly I can't really tell what's going on in the code. Method A provide Function B to Method C, which provides an InputStreamReader to Function B and executes it by wrapping a stream with a reader with a stream, but then Function B takes that stream and wraps it in another reader wrapped by another stream before finally wrapping it with another reader, *and then finally* applying that to the function that the developer specified, and the net result is...? Your guess is as good as mine. But you rightly point out that somewhere along the way, UTF-8 is brought in as a hardcoded default. I think THAT's what's mucking with our encoding and THAT's what needs to change.

@carlosmunoz since you wrote the code in question, can you help untangle what's actually going on in the ServletUtils class and whether this is likely where our encoding bug actually lies?

Also tagging @xdavidson and @aprajshekhar in case they can help sort this out.

carlosmunoz: So, there are 2 methods in that ServletUtil class: one takes a charset parameter, the other one doesn't (uses UTF-8 by default). This is the character set to use when reading the stream.
Does that make sense?

benradey: @rednitish Please remove the `encoding` variable completely and all subsequent references to it.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/ModuleVersionUpload.java
benradey: @rednitish Same, please remove this variable and all subsequent references.

pantheon-bundle/src/test/java/com/redhat/pantheon/servlet/module/ModuleVersionUploadTest.java
benradey: @rednitish, rather than overwriting the ISO_8859_1 test, can you copy/paste it so that your new test for the accented characters becomes an additional test?

benradey (changes requested):
@rednitish - after investigating some things locally, I discovered that with @carlosmunoz's changes in #432, the stream/reader objects that we work with handle the encodings automatically for us.

So, regardless of whether the upload used ISO_8859_1 or UTF-8, we don't need to worry about it, things "just work."

I made a few updated requests to reflect this new discovery. Thanks for your patience while we worked through this!

rednitish:
@benradey @carlosmunoz I have pushed few changes to retain iso encoding and improvised test cases. Please take a look

rednitish:
> @rednitish - after investigating some things locally, I discovered that with @carlosmunoz's changes in #432, the stream/reader objects that we work with handle the encodings automatically for us.
> 
> So, regardless of whether the upload used ISO_8859_1 or UTF-8, we don't need to worry about it, things "just work."
> 
> I made a few updated requests to reflect this new discovery. Thanks for your patience while we worked through this!

Hey Ben,

I tested the changes, and it was not working, so i have pushed small changes to fix it please do take a look.

uploader/pantheon.py
```diff
@@ -250,6 +250,7 @@ def process_file(path, filetype):
         data = _generate_data(jcr_primary_type, base_name, path.name, asccidoc_type='nt:file')
         # This is needed to add a new module version, otherwise it won't be handled
         data[':operation'] = 'pant:newModuleVersion'
+        data['Content-Type'] = 'UTF-8'
```
aprajshekhar, line 253: This approach is good for short term approach until we find how to pass charset as header with multipart data. However, use some other name rather than `content-type`. My suggestion is to use `encoding`

rednitish: renamed the parameter name to encoding

benradey (approved):
Looks good to me, Nitish!

I started wondering whether the different test cases that we have now actually mean anything. We don't have any encoding code of our own now, so we're effectively testing system functionality, which would be outside of our scope.

But I think keeping the test cases doesn't hurt anything, so let's leave them exactly the way you have them.

Thanks for being patient as we worked through the issue!

PR #428, opened 2020-09-28
Ccs 3798 modules included

xdavidson:
  * create a servlet for the frontend to fetch modules(published/unpublished) that's included in an assembly.
  * update the UI to show a list of modules included in an assembly


aprajshekhar (approved):
Looks good to me. The only improvement for the future that I see is consolidating documentsIncluded and assembliesData into one entity.

PR #432, opened 2020-10-02
fix the incorrect encoding of a stream parameter when a character encoding is specified

carlosmunoz:
This should fix the problem in #427 and still allow the api clients to use the encoding of their choice.

### DIFF
--- a/pantheon-bundle/pom.xml
+++ b/pantheon-bundle/pom.xml
@@ -465,6 +465,12 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.junit.jupiter</groupId>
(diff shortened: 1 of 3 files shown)


benradey:
Hi @carlosmunoz, thank you for submitting this PR. The code looks better, although I think there's still a problem here. When I upload a file with these contents:
`【感染者为0的城市--南京】日本导演镜头下最真实的南京防疫现场`
...that is still being read by the stream/reader classes as:
`?????0???--????????????????????`
So, something is still not quite right. I'm doing my best to debug this on my end for now, but if you have any ideas, I'm all ears.

benradey (approved):
Approving this because it's a step in the right direction for fixing our encoding issue, as we discussed over chat.

PR #434, opened 2020-10-13
Reworking portal url handling

benradey:
ACL issues were causing QA to see partial URLs when calling APIs unauthenticated, this fixes it to read the value directly from the environment variable which is not ACL'ed.


xdavidson (approved):
LGTM

rednitish (commented):
LGTM

PR #436, opened 2020-10-14
CCS-3812: Module details page needs to be refreshed for the "View on portal" link to show-up

aprajshekhar:
This PR
- Calls the url provider endpoint from publish event handler in modules component.
- Adds a wait of 1 sec to the above call so that JCR changes are propagated. This is needed because even though publish endpoint sends 200, it takes nearly a second for changes to propagate in JCR structure.

### DIFF
--- a/pantheon-bundle/frontend/src/app/assemblyDisplay.tsx
+++ b/pantheon-bundle/frontend/src/app/assemblyDisplay.tsx
@@ -272,6 +272,9 @@ class AssemblyDisplay extends Component<any, IAssemblyDisplayState> {
 
     private onPublishEvent = () => {
         this.getVersionUUID(this.props.location.pathname)
+        setTimeout(()=> {
+            this.getPortalUrl(this.props.location.pathname.substring(PathPrefixes.MODULE_PATH_PREFIX.length), this.state.variant)
+        }, 500)
     }
 
     private getVersionUUID = (path) => {
--- a/pantheon-bundle/frontend/src/app/moduleDisplay.tsx
+++ b/pantheon-bundle/frontend/src/app/moduleDisplay.tsx
@@ -302,6 +302,10 @@ class ModuleDisplay extends Component<any, IModuleDisplayState> {
 
     private onPublishEvent = () => {
         this.getVersionUUID(this.props.location.pathname)
+
+        setTimeout(()=> {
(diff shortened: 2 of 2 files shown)


pantheon-bundle/frontend/src/app/assemblyDisplay.tsx
```diff
@@ -254,6 +254,10 @@ class AssemblyDisplay extends Component<any, any, any> {
 
     private onPublishEvent = () => {
         this.getVersionUUID(this.props.location.pathname)
+        // wait for 1 second so that JCR data gets updated
```
xdavidson, line 257: assemblyDisplay doesn't have issue with  ""View on Customer Portal" link when publishing. This code is not neccessary.

pantheon-bundle/frontend/src/app/moduleDisplay.tsx
```diff
@@ -306,6 +306,10 @@ class ModuleDisplay extends Component<any, IModuleDisplayState> {
 
     private onPublishEvent = () => {
         this.getVersionUUID(this.props.location.pathname)
+        // wait for 1 second so that JCR data gets updated
```
xdavidson, line 309: I don't believe this fix the underlining problem. The reason on Module display, "View on Customer Portal" link doesn't show up when publishing is because, you need add this method call in public componentDidMount() method:
  this.getPortalUrl()

aprajshekhar: This was based on the existing code and not the new code that has just been merged. Will make changes accordingly. Also, there is a time gap between 200 OK received and the actual JCR updates happening in the backend. I will setup a meeting on Monday to discuss this PR.

pantheon-bundle/frontend/src/app/assemblyDisplay.tsx
aprajshekhar: I am seeing this issue on QA for assembly when I am publishing the assembly for the first time.

xdavidson (approved):
The fix works for local environment. Hopefully it also works on OHC where it's slower than local. I'm not a big fan with the setTimeout call. Would like to see a better fix in the future.

aprajshekhar:
Thanks for the approval. I am also not fan of the setTimeOut. However, any other approach would require a major change. So, this is a compromise.

PR #439, opened 2020-10-20
CCS-3812: Module details page needs to be refreshed for the "View on portal" link to show-up. 

aprajshekhar:
This PR implements a non-elegent fix since useEffect hook cannot be used with class based components.


rednitish (approved):
LGTM

PR #440, opened 2020-10-20
CCS-3812: Module details page needs to be refreshed for the "View on portal" link to show-up. Updating for the changes in master.

aprajshekhar:
Reverting the commit


rednitish (approved):
approving as it's revert.

PR #444, opened 2020-10-20
CCS-3839: Fixing preview issues loading JS, CSS and styling page

wesruv:
JS for loading cp-documentation stopped working for some reason
CSS was moved to the web component folder and link needed to be changed here
rhdocs is necessary on a wrapper for no-JS version to be styled correctly


xdavidson (approved):
In the future, we should config the portal url per environment instead of pointing to prod directly.

PR #447, opened 2020-10-23
Ccs 3839 images breaking content area

wesruv:


xdavidson (approved):
Tested in pantheon.corp.dev. The fix looks good.

PR #449, opened 2020-10-23
Ccs 3842 assembly warning alert

xdavidson:
This PR fixes the wording of Publish Warning for assembly. currently when an assembly failed to publish, the warning message states:
  
  Module versions
  Module failed to publish. Check the following:
     * Are you logged in as a publisher?
     * Does the module have all required metadata?

The new message looks like this:
  Publishing Document
  Assembly failed to publish. Check the following:
     * Are you logged in as a publisher?
     * Does the assembly have all required metadata?

### DIFF
--- a/pantheon-bundle/frontend/src/app/versions.tsx
+++ b/pantheon-bundle/frontend/src/app/versions.tsx
@@ -35,6 +35,7 @@ export interface IProps {
 }
 
 interface IState {
+    alertTitle: any
     allProducts: any
     allProductVersions: any
     canChangePublishState: boolean
@@ -69,6 +70,7 @@ class Versions extends Component<IProps, IState> {
     constructor(props) {
         super(props)
         this.state = {
+            alertTitle: "",
             allProducts: [],
             // tslint:disable-next-line: object-literal-sort-keys
             allProductVersions: [],
@@ -136,7 +138,7 @@ class Versions extends Component<IProps, IState> {
                 {this.state.publishAlertVisible && <div className="notification-container pant-notification-container-md">
                     <Alert
                         variant="warning"
-                        title="Publishing Document"
+                        title={this.state.alertTitle}
                         actionClose={<AlertActionCloseButton onClose={this.hidePublishAlert} />}
                     >
                         {this.capitalize(this.props.contentType)} failed to publish. Check the following:
@@ -518,6 +520,7 @@ class Versions extends Component<IProps, IState> {
                     } else {
                         console.log(buttonText + " failed " + response.status)
                         this.setState({ publishAlertVisible: true })
+                        this.getAlertTitle()
                     }
                     this.fetchVersions()
                 });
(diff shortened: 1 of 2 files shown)


pantheon-bundle/frontend/src/app/versions.tsx
```diff
@@ -744,6 +744,13 @@ class Versions extends Component<IProps, IState> {
                 })
         }
     }
+
```
rednitish, line 747: I have one query here, as we already know, content type can be either module or assembly, we can compare if content type is one of them other wise return. what do you suggest ?

pantheon-bundle/frontend/src/app/versions.tsx
```diff
@@ -136,13 +136,13 @@ class Versions extends Component<IProps, IState> {
                 {this.state.publishAlertVisible && <div className="notification-container pant-notification-container-md">
                     <Alert
                         variant="warning"
-                        title="Module Versions"
+                        title="Publishing Document"
```
rednitish, line 139: Can we write instead as :

"Publishing" {this.props.contentType}  ?

rednitish (commented):
I have few suggestions, let me know, if you differ with my opinion.

pantheon-bundle/frontend/src/app/versions.tsx
xdavidson: I tried the above suggestions. {this.props.contentType} or {...this.props.contentType} renders empty. I ended up creating a state variable to display the Alert title. That seems to reflect the contentType for publishing alerts.

pantheon-bundle/frontend/src/app/versions.tsx
```diff
@@ -744,6 +747,18 @@ class Versions extends Component<IProps, IState> {
...
     }
+
+    private capitalize = (str) => {
+        if (str === undefined || str.trim().length === 0) {
+            return
+        }
+        return str.charAt(0).toUpperCase() + str.slice(1)
+    }
+
+    private getAlertTitle = () => {
```
benradey, line 758: Could I please ask that this method be renamed? I would expect that a method named "getXXX" is a "safe"/read-only operation, but inside of it you are calling a setState. Maybe "resetAlertTitle" or "determineAlertTitle" or something else if you have a better idea.

pantheon-bundle/frontend/src/app/versions.tsx
```diff
@@ -35,6 +35,7 @@ export interface IProps {
 }
 
 interface IState {
+    alertTitle: any
```
benradey, line 38: Why `any` and not `string`?

pantheon-bundle/frontend/src/app/versions.tsx
xdavidson: updated

benradey:
Cool, thanks for the updates @xdavidson. I think this looks good, but since @rednitish was the first to comment, I'll let him provide the approving review once he's ready.

pantheon-bundle/frontend/src/app/versions.tsx
rednitish: Looks cleaner :+1:

rednitish (approved):
LGTM, Thanks Lisa for quick changes :+1:

PR #454, opened 2020-10-27
CCS-3841  500 Server error when publishing an assembly which includes a published module

benradey:


xdavidson (approved):
Deployed the branch on dev. The fix looks good!

PR #457, opened 2020-10-28
CCS-3849: Empty value in included_in_guides

aprajshekhar:
This PR fixes the issue where included_in_guides has empty value when both draft and released versions are present.


rednitish (approved):
LGTM

PR #459, opened 2020-10-29
CCS-3812: Module details page needs to be refreshed for the "View on portal" link to show-up.

aprajshekhar:
This PR:

1. Raises event when the publish/unpublish happens (in versions component)
2. Handles the event in module display and assembly display componnets
3. If event is publish, calls the API endpoint to get the URL.
4. If the response is not empty, set the URL
5. If the response is empty, construct the url on the client side.

### DIFF
--- a/pantheon-bundle/frontend/src/app/assemblyDisplay.tsx
+++ b/pantheon-bundle/frontend/src/app/assemblyDisplay.tsx
@@ -24,6 +24,10 @@ export interface IAssemblyDisplayState {
         variant: string
         variantUUID: string
         versionValue: string
+        portalHostUrl: string
+        productUrlFragment: string
+        versionUrlFragment: string
+        locale: string
 }
 
 class AssemblyDisplay extends Component<any, IAssemblyDisplayState> {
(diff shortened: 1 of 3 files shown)


rednitish (dismissed):
LGTM

pantheon-bundle/frontend/src/app/assemblyDisplay.tsx
```diff
@@ -271,10 +281,33 @@ class AssemblyDisplay extends Component<any, IAssemblyDisplayState> {
...
 
     private onPublishEvent = () => {
-        this.getVersionUUID(this.props.location.pathname)
-        setTimeout(()=> {
-            this.getPortalUrl(this.props.location.pathname.substring(PathPrefixes.MODULE_PATH_PREFIX.length), this.state.variant)
-        }, 500)
+        // try to get the url from api only if the component has been published
+        if (this.state.releaseUpdateDate.trim() !== ""
+        && this.state.releaseUpdateDate.length >= 15 ?
+            new Intl.DateTimeFormat("en-GB", { year: "numeric", month: "long", day: "numeric" }).format(new Date(this.state.releaseUpdateDate)) : "--") {
```
rednitish, line 287: Observational Comment:
Do we need to use date in this way ?

aprajshekhar: Thanks for the review. Once CCS-3860 is done, the whole publish event itself would change.

aprajshekhar: Anyways updated to simplify the logic.

rednitish (approved):
Voila, suggestions well taken. changes LGTM

PR #462, opened 2020-10-30
CCS-3862-title_fix

xdavidson:
This PR fixes the fix title issue in assembly api

### DIFF
--- a/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/assembly/AssemblyVariantJsonServlet.java
+++ b/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/assembly/AssemblyVariantJsonServlet.java
@@ -17,6 +17,7 @@ import com.redhat.pantheon.model.module.ModuleVariant;
 import com.redhat.pantheon.model.module.ModuleVersion;
 import com.redhat.pantheon.servlet.AbstractJsonSingleQueryServlet;
 import com.redhat.pantheon.servlet.ServletUtils;
+import com.redhat.pantheon.servlet.util.ServletHelper;
 import com.redhat.pantheon.servlet.util.SlingPathSuffix;
 import org.apache.sling.api.SlingHttpServletRequest;
 import org.apache.sling.api.resource.Resource;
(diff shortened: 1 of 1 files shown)


pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/assembly/AssemblyVariantJsonServlet.java
```diff
@@ -201,7 +202,10 @@ protected boolean isValidResource(@Nonnull SlingHttpServletRequest request, @Non
                         .variants().get()
                         .canonicalVariant().get();
                 moduleMap.put("canonical_uuid", canonical.uuid().get());
-                moduleMap.put("title", page.title().get());
+                // Get the current title from module instead of using the stored value in the content node
+                // When module title is modified, the value stored in content node becomes stale.
+                // Sometime, the title stored in content node shows filename instead of title.
+                moduleMap.put("title", ServletHelper.getModuleTitleFromUuid(canonical));
```
carlosmunoz, line 208: I won't block this, but I'll note that the name of the method is not good: it says `fromUuid` when in fact it doesn't accept a uuid, but a `ModuleVariant`.

carlosmunoz (approved):
Approved with caveats. Hopefully we can clean this up in the near future.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/assembly/AssemblyVariantJsonServlet.java
benradey, line 208: I see what you're doing here - it seems like your thought is "page.title() doesn't always give me the information I need, so let me get it from somewhere else."

Here's the thing though... page.title() is exactly what we should be using, so if that's broken, then we need to fix it.

page.title() is populated in AssemblyVersion.consumeTableOfContents(). It looks like it's coded to use the draft title metadata field, and if the draft doesn't exist, it falls back to filename.

ServletHelper.getModuleTitleFromUuid() has logic that is similar - it pulls the draft metadata title, then tries released, then falls back to blank.

Since that second case is where you want to go, I would suggest fixing consumeTableOfContents to call latest() instead of draft() in its current logic. I think that would strike a good balance. What do you think?

benradey (changes requested):

PR #467, opened 2020-11-10
Fix publish/updated dates CCS-3877 CCS-3879

benradey:
Several parts to this PR:

- Start recording dates on metadata properly
- Reintroduce copying metadata from released version onto draft version
- Refactor version upload servlets to share common code

### DIFF
--- a/pantheon-bundle/src/main/java/com/redhat/pantheon/model/document/DocumentMetadata.java
+++ b/pantheon-bundle/src/main/java/com/redhat/pantheon/model/document/DocumentMetadata.java
@@ -34,9 +34,20 @@ public interface DocumentMetadata extends WorkspaceChild {
     @Named("jcr:createdBy")
     Field<String> createdBy();
 
+    /**
+     * This represents the date that a document was published *most recently* and is not carried forward from version
+     * to version (since a document version should record its own publish date). Contract this with 'dateFirstPublished'
+     * which deals with the document's *first ever* publish date.
+     * @return
+     */
     @Named("pant:datePublished")
     Field<Calendar> datePublished();
 
+    /**
+     * This represents the date that a document was published *for the first time ever* and is carried forward from
+     * version to version. Contrast this with 'datePublished' which deals with the *most recent* publish date.
+     * @return
+     */
     @Named("pant:dateFirstPublished")
     Field<Calendar> dateFirstPublished();
 
--- a/pantheon-bundle/src/main/java/com/redhat/pantheon/model/document/DocumentMetadata.java
+++ b/pantheon-bundle/src/main/java/com/redhat/pantheon/model/document/DocumentMetadata.java
@@ -37,6 +37,9 @@ public interface DocumentMetadata extends WorkspaceChild {
     @Named("pant:datePublished")
     Field<Calendar> datePublished();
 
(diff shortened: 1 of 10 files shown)


pantheon-bundle/src/main/java/com/redhat/pantheon/model/document/DocumentMetadata.java
```diff
@@ -37,6 +37,9 @@
     @Named("pant:datePublished")
     Field<Calendar> datePublished();
 
+    @Named("pant:dateFirstPublished")
```
xdavidson, line 40: what's the difference between pant:datePublished and pant:dateFirstPublished?

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/util/VersionUploadOperation.java
```diff
@@ -0,0 +1,161 @@
...
+
+    private static final Set<String> METADATA_COPY_EXCLUDES = Collections.unmodifiableSet(
+            new HashSet<>(
+                    Arrays.asList(
+                            "jcr:description",
+                            "jcr:lastModified",
+                            "jcr:primaryType",
+                            "jcr:title",
+                            "pant:dateUploaded",
+                            "pant:datePublished"
```
xdavidson, line 49: should we consider the following two to be added to this list?
"pant:headline"
"pant:abstract"

pantheon-bundle/src/main/java/com/redhat/pantheon/model/document/DocumentMetadata.java
benradey: Great question. datePublished is specific to that particular version and is not carried forward to any new versions. dateFirstPublished is populated only once, on first-ever publish only, and then carried forward from version to version.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/util/VersionUploadOperation.java
benradey: Probably! If those are both populated automatically from the doc, then it doesn't make sense to carry them forward.

benradey:
@xdavidson I added another commit to address your comments.

pantheon-bundle/src/main/java/com/redhat/pantheon/model/document/DocumentMetadata.java
xdavidson: Thank you for the info! how are we going to handle the modules/assemblies published in prod today that don't have pant:dateFirstPublished?

benradey: Another good question. :)

I have the logic set up in AsciidoctorService.java:258 so that the 'datePublished' value is used for both the "updated" and "published" injected attributes. Then immediately after that, if the 'firstPublished' value exists, it overwrites the "published" injected attribute.

So all of our docs today will rely on the fallback logic. Over time it will become less necessary, but the bottom line is that today's docs will be handled as best as possible with what we have.

xdavidson (approved):

PR #469, opened 2020-11-16
Ccs 3997 product version form

grantsydney:


pantheon-bundle/frontend/src/app/product.tsx
```diff
@@ -94,8 +98,24 @@ class Product extends Component<any, IState> {
...
+                                        label="Version URL Fragment:"
+                                        isRequired={true}
+                                        fieldId="version_url_fragment"
+                                >
+                                    <TextInput id="new_version_url_fragment" type="text" placeholder="Version URL Fragment" onChange={this.handleUrlInputChange} value={this.state.versionUrlFragment} />
+                                </FormGroup>
+                                <br />
                                 <ActionGroup>
-                                    <Button aria-label="Creates a new Product Name with Description specified." onClick={this.saveProduct}>Save</Button>
+                                    <Button type='submit' aria-label="Creates a new Product Name with Description specified." onClick={e => this.saveProduct(e)}>Save</Button>
```
xdavidson, line 118: we don't need the lambda function for onClick event or the Button type. I recommend:

<Button aria-label="Creates a new Product Name with Description specified." onClick={this.saveProduct}>Save</Button>

pantheon-bundle/frontend/src/app/product.tsx
```diff
@@ -26,18 +28,20 @@ class Product extends Component<any, IState> {
...
-
+    
     // render method transforms the react components into DOM nodes for the browser.
     public render() {
         return (
             <React.Fragment>
                 {/* Bullseye makes sure everyhting is in the middle */}
                 <Bullseye>
-                    <Form>
+                    <Form onSubmit={this.saveProduct} method='POST'>
```
xdavidson, line 44: I don't believe we don't believe we the changes on line 44.

pantheon-bundle/frontend/src/app/product.tsx
```diff
@@ -125,8 +145,52 @@ class Product extends Component<any, IState> {
...
+        if (/^[-.\w]+$/.test(versionUrlFragment)) {
+            this.setState({ versionUrlFragment, isUrlFragmentValid: true })
+        } else {
+            this.setState({ isUrlFragmentValid: false})
+        }
+    }
+
+    private saveVersion = () => {
+        // confirming form state is valid
+        if (this.state.versionName === "" || this.state.productUrlFragment === "") {
```
xdavidson, line 162: line 162 does not check if users can enter an empty string in Version Name field. It would be better if we do something like this:
if (this.state.versionName.trim() === "" || this.state.versionUrlFragment === "")

xdavidson (approved):

PR #484, opened 2020-12-04
Search Beta fixes2

xdavidson:
This PR fixes a jackrabbit exception when a repository name contains space

### DIFF
--- a/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/ModuleListingServlet.java
+++ b/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/ModuleListingServlet.java
@@ -100,13 +100,11 @@ public class ModuleListingServlet extends AbstractJsonQueryServlet {
 
         StringBuilder queryBuilder = null;
         if (repoParam != null) {
-            // encode if repo name contains space
+            // encode if repo name contains space.
+            // https://www.mail-archive.com/users@jackrabbit.apache.org/msg05782.html
             String repos = Arrays.stream(repoParam)
                     .map( repo -> {
-                        if (repo.contains(" ")) {
-                            return ISO9075.encode(repo);
-                        }
-                        return repo;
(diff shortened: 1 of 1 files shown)


pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/ModuleListingServlet.java
```diff
@@ -99,9 +100,15 @@ protected String getQuery(SlingHttpServletRequest request) {
...
         StringBuilder queryBuilder = null;
         if (repoParam != null) {
-            log.info("[" + ModuleListingServlet.class.getSimpleName() + "] contentTypeParam: " + contentTypeParam);
-            String repos = String.join(" | ", repoParam);
-            log.info("[" + ModuleListingServlet.class.getSimpleName() + "] repos: " + repos);
+            // encode if repo name contains space
+            String repos = Arrays.stream(repoParam)
+                    .map( repo -> {
+                        if (repo.contains(" ")) {
+                            return ISO9075.encode(repo);
```
aprajshekhar, line 107: Just one query - what is the intent behind using ISO9875.encode rather than UTF-8?

xdavidson: jackrabbit uses ISO9075.encode to handle the spaces.
https://www.mail-archive.com/users@jackrabbit.apache.org/msg05782.html

benradey: That is an excellent find, bravo for being able to dig up this relatively obscure internet artifact.

Would you mind adding this link as a comment in the source code?

Also, do we need the `if (repo.contains(" "))` statement? Can we just encode the string unconditionally?

benradey (approved):
Code looks good. I didn't test locally, I assume Lisa did her due diligence. Thanks Lisa!

PR #485, opened 2020-12-04
CCS-3908 Rework API so the body is what's inside of the cp-documentation tag

benradey:


benradey:
@wesruv @grantsydney Right now, if the code doesn't find the 'doc-content' ID on any element, it returns the entire body. Is that OK or do we prefer something else?

wesruv (approved):

wesruv:
Looks good to me

PR #491, opened 2020-12-21
CCS-3956: Make python uploader work as a library.

aprajshekhar:
This PR
- Makes required minimal changes to the uploader so that 
  -  It can be called as a library.
  - Progress and result can be saved in a in-memory datastructure and optionally published to redis.
- The changed uploader is in uploader-service directory so that current workflow is not impacted.
- In the future more cosmetic changes may be applied as service is created.


rednitish (approved):
LGTM

PR #494, opened 2020-12-29
CCS-3774: Rewrite the git2pantheon go app in Python.

aprajshekhar:
This PR:

1. Adds a new python based service that provides the same endpoints as that of current Git2Pantheon with the following changes:
    - The response status for /clone endpoint will return 202 with the key that could be used to extract current status using /status endpoint.
2. Adds a new endpoint /status that provides brief current status.
3. Adds Swagger for the endpoints.
4. Updated the progress DS to be more informative.
5. Adds logging of values passed to the uploader.

### DIFF
--- a/uploader-service/git2pantheon/__init__.py
+++ b/uploader-service/git2pantheon/__init__.py
@@ -1,0 +1,58 @@
+import logging
+import os
+
+from flask import Flask, json
+from werkzeug.exceptions import HTTPException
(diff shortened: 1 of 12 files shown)


aprajshekhar:
https://github.com/redhataccess/pantheon/pull/495 contains the changes in this PR. Hence, closing this PR.

PR #497, opened 2021-01-11
Ccs 3993 sitemap

xdavidson:
This PR introduces sitemap for the following resources:

- assembly (/api/sitemap/assembly.sitemap.xml)
- module (/api/sitemap/module.sitemap.xml)

### DIFF
--- a/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/sitemap/SiteMapServlet.java
+++ b/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/sitemap/SiteMapServlet.java
@@ -46,39 +46,25 @@ public class SiteMapServlet extends SlingAllMethodsServlet {
 
     private static final String RESOURCE_ROOT = "/content/repositories";
 
-    private Set<Resource> getAsset(Resource resource, String documentVariantResourceType, String documentVersionResourceType, Set<Resource> assets) {
+    private Set<Resource> getAsset(Resource resource, String documentVersionResourceType) {
 
-        for (Resource r1 : resource.getChildren()) {
-            // Use Resource Filter Stream to limit memory consumption and path traversal
-            ResourceFilterStream rfs = r1.adaptTo(ResourceFilterStream.class);
+        // Use Resource Filter Stream to limit memory consumption and path traversal
+        ResourceFilterStream rfs = resource.adaptTo(ResourceFilterStream.class);
 
-            assets.addAll(rfs.setBranchSelector("[sling:resourceType] == '" + documentVariantResourceType + "'")
-                    .setChildSelector("[released/sling:resourceType] == '" + documentVersionResourceType + "'")
-                    .stream()
-                    .collect(Collectors.toSet())
-                    );
-
-            if (r1.hasChildren()) {
-                getAsset(r1, documentVariantResourceType, documentVersionResourceType, assets);
-            }
-        }
-        return assets;
+        return rfs.setChildSelector("[released/sling:resourceType] == '" + documentVersionResourceType + "'")
+                        .stream()
+                        .collect(Collectors.toSet());
     }
 
 
     @Override
     protected void doGet(SlingHttpServletRequest request, SlingHttpServletResponse response) throws IOException {
-
-        String documentVariantResourceType = "";
         String documentVersionResourceType = "";
         Resource resource = request.getResourceResolver().getResource(RESOURCE_ROOT);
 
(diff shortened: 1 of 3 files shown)


pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/sitemap/SiteMapServlet.java
```diff
@@ -0,0 +1,172 @@
...
+            ResourceFilterStream rfs = r1.adaptTo(ResourceFilterStream.class);
+
+            assets.addAll(rfs.setBranchSelector("[sling:resourceType] == '" + documentVariantResourceType + "'")
+                    .setChildSelector("[released/sling:resourceType] == '" + documentVersionResourceType + "'")
+                    .stream()
+                    .collect(Collectors.toList()));
+            if (r1.hasChildren()) {
+                getAsset(r1, documentVariantResourceType, documentVersionResourceType, assets);
+            }
+        }
```
benradey, line 67: Is there a more efficient way to gather all children? Even if you are adapting to a stream, which is a cool technique that I have never seen before, this is still a loop that performs a database query for every single node in the tree. That could be a very expensive operation.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/sitemap/SiteMapServlet.java
```diff
@@ -0,0 +1,172 @@
...
+        DocumentVariant documentVariant = null;
+        Date dateModified = null;
+
+        // Process external url
+        if (System.getenv(PORTAL_URL) != null) {
+            if (resource.isResourceType(RESOURCE_TYPE_ASSEMBLYVARIANT)) {
+                documentVariant = resource.adaptTo(AssemblyVariant.class);
+            } else if (resource.isResourceType(RESOURCE_TYPE_MODULEVARIANT)) {
+                documentVariant = resource.adaptTo(ModuleVariant.class);
+            }
```
benradey, line 141: Couldn't you get rid of the 'if' statement and just use `resource.adaptTo(DocumentVariant.class)`?

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/sitemap/SiteMapServlet.java
```diff
@@ -0,0 +1,172 @@
...
+            if (documentVariant != null && documentVariant.released().isPresent()) {
+                locPath = new CustomerPortalUrlUuidProvider().generateUrlString(documentVariant);
```
benradey, line 143: You could probably re-use the same CustomerPortalUrlUuidProvider instance, but I don't think creating a new one is particularly expensive either, so this is pretty minor.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/sitemap/SiteMapServlet.java
xdavidson: I am using Resource Filter Stream service because is way more efficient than a database query. 
https://sling.apache.org/documentation/bundles/resource-filter.html

I also have a branch that uses one sql2 query to get the released assets, which is very slower on dev/qa/stage. It reminds me of the search home page. I have to wait for a while to see the display.  I'll deploy both branch on ohc so that you can see the difference in performance gain using Resource Filter Stream.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/sitemap/SiteMapServlet.java
xdavidson: done

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/sitemap/SiteMapServlet.java
xdavidson: updated

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/sitemap/SiteMapServlet.java
xdavidson: Ben, good call on the loop. It's removed.

benradey (approved):
Looks great, @xdavidson! I'm glad the loop can be removed without any change in functionality. You found a great technique here and I bet we can apply it to other areas of the application as well. Good work. :)

PR #498, opened 2021-01-15
CCS-4064: add additional search beta classname

grantsydney:


pantheon-bundle/frontend/src/app/__snapshots__/searchBeta.test.tsx.snap
```diff
@@ -310,6 +310,7 @@ exports[`SearchBeta tests should render default Search component 1`] = `
               toggleTextExpanded=""
             >
               <SearchInput
+                className="filters-drawer filters-drawer__repo-search "
```
wesruv, line 313: The parent is `filters-drawer` (the component wrapper) so this one only needs `filters-drawer__repo-search`, indicating that it's the `repo-search` part of `filters-drawer`.

You want repetition with modifiers (e.g. `--by-repo`), but not elements (e.g. `__repo-search`). Modifiers are saying it's a special variation of the normal styles, so we want both to apply the base styles AND then the special styles.

Having `filters-drawer` and `filters-drawer__repo-search` on the same element is saying this ONE thing is TWO parts of the same thing, the component wrapper AND a part of the component, which we don't want to do.

pantheon-bundle/frontend/src/app/searchBeta.tsx
```diff
@@ -138,6 +138,7 @@ class SearchBeta extends Component<IAppState, ISearchState> {
               value={this.state.repoFilterValue}
               onChange={this.onChangeRepoFilter}
               onClear={(evt) => this.onChangeRepoFilter("", evt)}
+              className='filters-drawer filters-drawer__repo-search '
```
wesruv, line 141: Same as above.

wesruv (changes requested):

wesruv (approved):
Poifect!

PR #500, opened 2021-01-18
CCS-4105: add react testing library and some tests

grantsydney:


wesruv (dismissed):

pantheon-bundle/frontend/package.json
```diff
@@ -22,6 +22,7 @@
     "clean": "rimraf dist"
   },
   "devDependencies": {
+    "@testing-library/react": "^10.1.0",
```
xdavidson, line 25: The latest version is 11.2.3. I wonder why we don't use the latest?

grantsydney: thanks for catching. this was the version installed by default. Updated the pr

xdavidson (approved):

PR #501, opened 2021-01-20
Minor change to the "Configuring your repository" assembly.

adahms:
Interestingly, this one-line change resolves an error running the AsciiDoctor command. Because this assembly does not render correctly in Pantheon V2 today, and this is the only difference I can see between this and other assemblies, my guess is that this will resolve that issue as well.


adahms:
@benradey - another one-liner if you're open to merging this.

I promise to keep these to a minimum, and this PR should resolve a rather curious AsciiDoctor error we're encountering on the CLI.

benradey (approved):

benradey:
@adahms At least these don't require effort to review and are pretty mechanical, so no worries :)

PR #502, opened 2021-01-22
CCS-3917 End User Docs

benradey:
Adding a couple classes to support rendering the special case of EUD docs, as well as tweaking some existing classes to help support EUD docs as well (for example, disabling xref processing). The logic checks to see if newer documents have been uploaded to /content/docs in the JCR, otherwise it falls back to the static help docs that were present when the pantheon bundle was built that are located in /content/staticdocs.


xdavidson (approved):
LGTM

PR #504, opened 2021-01-26
Update assembly-pantheon-help.adoc

pwright:
be consistent in the use of modules (and help support use of https://chrome.google.com/webstore/detail/asciidoctorjs-live-previe/iaalpfgpbocpdfblpnhhgllgbdbchmia)

### DIFF
--- a/pantheon-bundle/src/main/resources/SLING-INF/content/docs/assemblies/assembly-pantheon-help.adoc
+++ b/pantheon-bundle/src/main/resources/SLING-INF/content/docs/assemblies/assembly-pantheon-help.adoc
@@ -9,7 +9,7 @@ include::../attributes.adoc[]
 
 Welcome to {ProductShortName} help.
 
-include::modules/ref_prerequisites.adoc[leveloffset=+1]
+include::../modules/ref_prerequisites.adoc[leveloffset=+1]
 
 // Link to Overview of Pantheon
 


pwright:
If this PR is merged, you can view the whole guide (using the chrome extension above) at https://raw.githubusercontent.com/pwright/pantheon/patch-1/pantheon-bundle/src/main/resources/SLING-INF/content/docs/assemblies/assembly-pantheon-help.adoc

stoobie:
Hey @pwright Thanks for the contribution!

If I understand correctly, the goal in this PR is to enable previewing the Pantheon 2 documentation using a Google Chrome extension. I'm guessing that the extension doesn't support the use of symlinks in file paths, which would explain why you changed the include file path. Is that correct?

So first of all, be aware that because of Pantheon 2's current lack of support for nested assemblies, we are about to merge (PR 533)[https://github.com/redhataccess/pantheon/pull/533] which will severely impact this PR (504). You might want to take a look at that one.

Secondly, it would be helpful (if you decide to update this PR after we merge 533) to include some text describing asciidoctorjs-live-preview and where to find info on installing and using it. I think that might actually make the most sense to appear in the README.

Thirdly, be aware that the EUD is slated to be released and viewable from within Pantheon in the next release of Pantheon, which I am given to understand will be within a few weeks. Once it is viewable from within Pantheon, do you still see benefit to this PR?

pwright:
My reading of the related PR (533) is that we're going 'All in' wrt symlinks.
If true, feel free to close this PR as it relies on not using symlinks.

viewable in pantheon requires vpn access, and will only show latest published version. This PR would allow viewing of all branches, versions, forks, etc (that's the point of it)

pwright:
@stoobie You made some good points above, I'll follow up with #541

stoobie:
> @stoobie You made some good points above, I'll follow up with #541

Great, @pwright , looks good. I'm looking forward to trying it out! It sounds like a decent solution.

stoobie:
@pwright So now that #533 is merged, and you opened https://github.com/redhataccess/pantheon/issues/541, is this current PR still necessary?

pwright:
@stoobie  you're correct, closing

PR #505, opened 2021-01-29
PDF generation

carlosmunoz:
This is a preview for PDF generation of pantheon content using the asciidoctor-pdf library.

To access a basic PDF render of any document just access its full path in the document using the pdf extension. For example:

`GET http://<server>/path/to/document.adoc.pdf?draft=true` (note the draft parameter is being used)

**Known issues and other comments**
- Images are not being inserted into the documents
- Stylesheets are out-of-the-box
- The pdf file is being generated on every request


jmarrero (changes requested):
this PR is not ready :hugs:

PR #508, opened 2021-02-02
CCS-3860 Return the Portal URL from Publish API

rednitish:
At present the publish API[1] returns 200 OK with no body. Due to this, on the UI a separate call has to be made to retrieve the Portal URL.
If the publish API returns the portal url as part response body it will reduce no. of calls  at UI side.


xdavidson (approved):
LGTM

PR #513, opened 2021-02-12
CCS-3917 Tweaking symlink logic

benradey:
In troubleshooting an EUD issue, I discovered that the way that sling handles the combination of "working symlink to parent folder" and "missing ultimate file target" is unintuitive and our code hadn't previously handled that scenario well. This is a hopeful attempt to fix that.


xdavidson (approved):
LGTM

PR #522, opened 2021-02-25
CCS-4208 Fixing symlinks

benradey:
Our symlink logic introduces double-slashes into resource paths, and that was interfering with our check to see if the resource that was returned was the same one that we requested.


xdavidson (approved):
LGTM

PR #526, opened 2021-03-02
CCS-4212-add-delete-filter-chip-functionality

grantsydney:


xdavidson (approved):
Tested locally. Filter is working as expected.

PR #529, opened 2021-03-05
Add an initial revision of a semantic vanity url servlet

carlosmunoz:
This is a very early release for an API that allows the fetching of module **data** by using a more semantic approach to urls.

Once a module has been fully configured and released, this servlet allows to get its information via the following url patterns:

`http://localhost:8181/api/module/vanity.json/<locale>/<product_url>/<version_url>/<module_variant_uuid>

or the similar

`http://localhost:8181/api/module/vanity.json/<locale>/<product_url>/<version_url>/<module_vanity_url>

this may result in API urls like this:

`http://localhost:8181/api/module/vanity.json/en-us/rhel/8.0/configuring_wired_connection_using_control_center`

The url pattern can be changed but this would allow Pantheon to offer the content based on the configured metadata values, and as a result would offer repeatable urls for content.

### DIFF
--- a/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/VanityVariantJsonServlet.java
+++ b/pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/VanityVariantJsonServlet.java
@@ -1,0 +1,143 @@
+package com.redhat.pantheon.servlet.module;
+
+import com.ibm.icu.util.ULocale;
+import com.redhat.pantheon.jcr.JcrQueryHelper;
+import com.redhat.pantheon.model.api.SlingModels;
+import com.redhat.pantheon.model.module.ModuleVariant;
+import com.redhat.pantheon.model.module.ModuleVersion;
+import com.redhat.pantheon.servlet.MapConverters;
+import com.redhat.pantheon.servlet.util.SlingPathSuffix;
+import org.apache.sling.api.SlingHttpServletRequest;
+import org.apache.sling.api.SlingHttpServletResponse;
+import org.apache.sling.api.servlets.SlingSafeMethodsServlet;
+import org.apache.sling.servlets.annotations.SlingServletPaths;
+import org.jetbrains.annotations.NotNull;
+import org.osgi.framework.Constants;
+import org.osgi.service.component.annotations.Component;
+
+import javax.jcr.RepositoryException;
+import javax.servlet.Servlet;
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+import java.util.Map;
+import java.util.Optional;
+
+import static com.redhat.pantheon.servlet.ServletUtils.writeAsJson;
+
+/**
+ * An API endpoint for finding module information using a vanity url format which takes into account the content's
+ * product metadata, and locale. The format is as follows:<br><br>
+ *
+ * '/api/module/vanity.json/{locale}/{product}/{version}/{vanityUrlFragment}'<br><br>
+ *
+ * It will also resolve content with the following format: <br><br>
+ *
+ * '/api/module/vanity.json/{locale}/{product}/{version}/{moduleUUID}'
+ *
+ * <br>
+ * @see VariantJsonServlet for a servlet returning the same information
+ * @author Carlos Munoz
+ */
+@Component(
+        service = Servlet.class,
+        property = {
+                Constants.SERVICE_DESCRIPTION + "=Servlet to facilitate GET operation which accepts several path parameters to fetch module variant data",
+                Constants.SERVICE_VENDOR + "=Red Hat Content Tooling team"
+        })
+@SlingServletPaths(value = "/api/module/vanity")
+// /api/module/vanity.json/{locale}/{productLabel}/{versionLabel}/{vanityUrl}
+public class VanityVariantJsonServlet extends SlingSafeMethodsServlet {
+
+    private final SlingPathSuffix suffix = new SlingPathSuffix("/{locale}/{productLabel}/{versionLabel}/{vanityUrl}");
+
+    @Override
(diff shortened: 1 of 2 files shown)


pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/VanityVariantJsonServlet.java
```diff
@@ -0,0 +1,143 @@
...
+        service = Servlet.class,
+        property = {
+                Constants.SERVICE_DESCRIPTION + "=Servlet to facilitate GET operation which accepts several path parameters to fetch module variant data",
+                Constants.SERVICE_VENDOR + "=Red Hat Content Tooling team"
+        })
+@SlingServletPaths(value = "/api/module/vanity")
+// /api/module/vanity.json/{locale}/{productLabel}/{versionLabel}/{vanityUrl}
+public class VanityVariantJsonServlet extends SlingSafeMethodsServlet {
+
+    private final SlingPathSuffix suffix = new SlingPathSuffix("/{locale}/{productLabel}/{versionLabel}/{vanityUrl}");
```
xdavidson, line 52: the customer portal view_uri is constructed with product urlFragment and version urlFragment. not labels. Should we follow the same convention and use productUrlFragment and versionUrlFragment?

carlosmunoz: Indeed. If you look at the code it is in fact filtering by url fragment. The strings are just pattern placeholders to be able to extract the right values from the url. I will change them to reflect that they are url fragments and not label.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/VanityVariantJsonServlet.java
```diff
@@ -0,0 +1,143 @@
...
+        String productLabel = params.get("productLabel");
+        String versionLabel = params.get("versionLabel");
+        String vanityUrl = params.get("vanityUrl");
+
+        JcrQueryHelper queryHelper = new JcrQueryHelper(request.getResourceResolver());
+        try {
+
+            // Find the metadata with the right vanity url,
+            Optional<ModuleVersion> moduleVersion = queryHelper.query(
+                    "select * from [pant:moduleVersion] as v where v.[metadata/urlFragment] = '/" + vanityUrl + "'")
```
benradey, line 67: A couple thoughts...
- Why are you using Module classes instead of Document classes (then it would hopefully work for Modules and Assemblies)?
- I see you're querying for the vanity url at the location we keep it today, which is on the metadata node. That does work, but I think I'd like to see a system where we manage vanity URLs (and redirects, etc) as separate entities in the system (perhaps located at /content/redirects in the JCR, something along those lines?) that contain pointers to the UUIDs of the DocumentVariants (would need to nail down what makes the most sense there) that they redirect to. In that world, this query would obviously need to be refactored. Regardless, the overall concept here is still solid.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/VanityVariantJsonServlet.java
```diff
@@ -0,0 +1,143 @@
...
+ * @see VariantJsonServlet for a servlet returning the same information
+ * @author Carlos Munoz
+ */
+@Component(
+        service = Servlet.class,
+        property = {
+                Constants.SERVICE_DESCRIPTION + "=Servlet to facilitate GET operation which accepts several path parameters to fetch module variant data",
+                Constants.SERVICE_VENDOR + "=Red Hat Content Tooling team"
+        })
+@SlingServletPaths(value = "/api/module/vanity")
```
benradey, line 48: Interesting that you register this servlet by path. I'm wondering if that doesn't overburden Drupal... Drupal would have to be smart enough to know whether the URL it was handling was a vanity URL or not and then call the correct Pantheon endpoint accordingly. That might be tough. I wonder if it doesn't make more sense for us to try and implement something more like a DefaultGetServlet for Sling that could handle the whole spectrum of requests and internally forward them to servlets appropriately.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/VanityVariantJsonServlet.java
```diff
@@ -0,0 +1,143 @@
...
+                    // and the right product
+                    .filter(modVer -> {
+                        try {
+                            return modVer.metadata().get().productVersion().getReference().getProduct().urlFragment().get().equals(productLabel);
+                        } catch (RepositoryException e) {
+                            throw new RuntimeException(e);
+                        }
+                    })
+                    // There should be 1 at most, but get the first if there are more
+                    .findFirst();
```
benradey, line 92: In today's world, there could definitely be more than one. There's nothing stopping me from adding a vanity URL of "/cats" to every document in the system. My idea above of managing redirects as separate entities neatly addresses this problem.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/module/VanityVariantJsonServlet.java
```diff
@@ -0,0 +1,143 @@
...
+
+            if(!moduleVersion.isPresent()) {
+                // try to find by variant uuid, instead of vanity url
+                moduleVersion = findModuleByUuid(queryHelper, productLabel, versionLabel, locale, vanityUrl);
```
benradey, line 96: I'm wondering why this "vanity" servlet is also accepting the UUID - perhaps you envision this new servlet as "the" one-and-only servlet that Drupal would call, and it handles anything you throw at it? Can you clarify if that's what you intended?

benradey (commented):
Carlos, I like this initial stab at vanity URLs - but there are some big questions that I touched on in my comments and I'd like us to have a conversation about our high-level approach/architecture before we start committing code for any functionality here. Regardless, thanks for this submission, it will definitely help add some context to our conversations.

PR #530, opened 2021-03-05
CCS-4256-filter-by-repo-cleanup

grantsydney:
add scroll to repo list in filter by repo section so repos don't extend to the bottom of search page
fix bug that cause checked repo to uncheck when remove filter input


xdavidson (approved):
LGTM

PR #532, opened 2021-03-08
CCS 3861 , get url by converting response to json and parsing path

rednitish:
UI changes for adapting url fragment from backend call, and


aprajshekhar (approved):
Thanks for making the changes that we discussed. LGTM

PR #543, opened 2021-03-16
CCS-4249-SSO_enhancement_revert

xdavidson:
Revert login and logout enhancement. It breaks the session.


rednitish (approved):
Approving, It is revert of last merge.

PR #548, opened 2021-03-25
CCS-4285: disabling module and assembly sections

grantsydney:


pantheon-bundle/frontend/src/app/search.tsx
```diff
@@ -558,6 +560,9 @@ class Search extends Component<IAppState, ISearchState> {
 
   // methods for bulk operation
   private getdocumentsSelected = (documentsSelected) => {
+    if(this.state.repositoriesSelected.length == 0){
```
xdavidson, line 563: can we do 

if(this.state.repositoriesSelected.length === 0){ ...}

xdavidson (approved):

PR #549, opened 2021-03-25
Ccs 4269 bugfixes2

xdavidson:
This PR checks if users try to apply the same product/version to multiple repositories

### DIFF
--- a/pantheon-bundle/frontend/src/app/search.tsx
+++ b/pantheon-bundle/frontend/src/app/search.tsx
@@ -309,11 +309,6 @@ class Search extends Component<IAppState, ISearchState> {
       </React.Fragment>
     );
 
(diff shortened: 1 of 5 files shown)


pantheon-bundle/frontend/src/app/search.tsx
```diff
@@ -302,6 +309,11 @@ class Search extends Component<IAppState, ISearchState> {
       </React.Fragment>
     );
 
+    const editMetadataAlert = (
```
grantsydney, line 312: Is this ever used? If not we should delete

xdavidson: removed!

grantsydney (approved):

PR #551, opened 2021-03-29
 (CCS-4273) Create validation result Node(s) in JCR

rednitish:
Validate Xrefs
Create validation result Node(s) in JCR


pantheon-bundle/src/main/java/com/redhat/pantheon/asciidoctor/AsciidoctorService.java
```diff
@@ -324,6 +322,17 @@ private String buildDocument(@Nonnull Document base, @Nonnull Locale locale, @No
                 content.append(xrefProcessor.preprocess(rawContent));
 
                 html = asciidoctor.convert(content.toString(), ob.get());
+                Violations violations = new XrefValidator(html).validate();
+                Validations validationResult = documentVersion.validations().getOrCreate();
```
aprajshekhar, line 326: This can be a part of of utility method of a Helper class, for example `ValidationDataOps` such that the Violations can be passed and  the validation nodes will be returned.

pantheon-bundle/src/main/java/com/redhat/pantheon/asciidoctor/AsciidoctorService.java
```diff
@@ -324,6 +322,17 @@ private String buildDocument(@Nonnull Document base, @Nonnull Locale locale, @No
...
+                Validation validation = validationResult.page(validationResult.getValueMap().size()).getOrCreate();
+                validation.validationCategory().set("xref");
+                if(violations.hasViolations()){
+                    validation.message().set("Not a valid Xref");
```
aprajshekhar, line 330: The Validator itself should set these messages.

pantheon-bundle/src/main/java/com/redhat/pantheon/asciidoctor/AsciidoctorService.java
rednitish: @aprajshekhar Apart from Violations, we need to pass the context (in our case documentVersion) to create and return Validation node.

pantheon-bundle/src/main/java/com/redhat/pantheon/model/document/Validation.java
```diff
@@ -0,0 +1,27 @@
+package com.redhat.pantheon.model.document;
+
+import com.redhat.pantheon.model.api.Field;
+import com.redhat.pantheon.model.workspace.WorkspaceChild;
+import org.apache.jackrabbit.JcrConstants;
+
+import javax.inject.Named;
+import java.util.Calendar;
+
+
```
xdavidson, line 10: Should we add an annotation for @jcr:PrimaryType(pant:Validation) ?

xdavidson: of course you need define pant:validation in resources/SLING_INF/nodetypes/nodetypes.cnd first

pantheon-bundle/src/main/java/com/redhat/pantheon/asciidoctor/AsciidoctorService.java
```diff
@@ -324,6 +322,17 @@ private String buildDocument(@Nonnull Document base, @Nonnull Locale locale, @No
...
+                    validation.validator().set(violations.get("Not a valid Xref").getDetails());
+                }else{
+                    validation.message().set("valid Xref");
```
xdavidson, line 333: I think we should only store the messages when something is not valid.

pantheon-bundle/src/main/java/com/redhat/pantheon/model/document/Validations.java
```diff
@@ -0,0 +1,10 @@
+package com.redhat.pantheon.model.document;
+
+import com.redhat.pantheon.model.api.Child;
+import com.redhat.pantheon.model.workspace.WorkspaceChild;
+
+public interface Validations extends WorkspaceChild {
```
xdavidson, line 6: I wonder if we could use the annontation on line 5 to define the primaryType for Validatiions node
@JcrPrimaryType("sling:OrderedFolder")

pantheon-bundle/src/main/java/com/redhat/pantheon/validation/validators/XrefValidator.java
```diff
@@ -0,0 +1,96 @@
...
+            return violations;
+        }
+        return violations.add("Not a valid Xref",
+                new ErrorDetails().add("invalid Cross reference(s) exists in the document"));
+    }
+
+    private boolean isValidXref() {
+        try {
+                Document doc = Jsoup.parse(content);
+                Elements resultLinks = doc.select("h2");
```
xdavidson, line 65: this only validates links in a <h2> tag, currently they for _addtional_resources section. 
I would like to see us validate filepath-based xref. please see comments in the jira ticket.

rednitish: @xdavidson among list of all links, we are checking xrefs, [XrefValidationHelper().getObjectsToValidate()] which we have collected while XrefProcessing.

pantheon-bundle/src/main/java/com/redhat/pantheon/validation/helper/XrefValidationHelper.java
```diff
@@ -0,0 +1,21 @@
+package com.redhat.pantheon.validation.helper;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+
+public class XrefValidationHelper {
```
aprajshekhar, line 9: I am liking this approach.

pantheon-bundle/src/main/java/com/redhat/pantheon/validation/validators/XrefValidator.java
xdavidson: you're right. the part I'm little concerned is the "h2" selector used for resultLinks. It seems by doing so, you're only checking the xrefs inside that selector and overlook the xrefs that happen to be outside of the "h2".

xdavidson (approved):
Looks good to for the first pass.

PR #559, opened 2021-04-05
CCS:42711: some bulk publish updates

grantsydney:
- enable check icon to be added or removed based on publish/unpublish operation
- user cannot publish titles from multiple repos. alert message will display and publish/unpublish buttons are disabled
- locale and variant form data are now appended once, not for each title being published


xdavidson (approved):
Tested locally. LGTM

PR #564, opened 2021-04-06
bulk operation button double click fix

grantsydney:
previously when a user closed a publish/unpublish modal, and then tried to click another bulk operation button, they would have to click twice in order for the modal to appear. Now the button just needs one click


xdavidson (approved):
LGTM

PR #566, opened 2021-04-07
Xref validations

rednitish:
Validation Node should be created in case xrefs present in document is invalid. i.e. xyz.adoc (logic update )


aprajshekhar:
LGTM

aprajshekhar (approved):

PR #567, opened 2021-04-07
Xref validations

rednitish:
xref : create new sibling nodes in case of more than one xrefs failing in single document


rednitish:
> This PR introduces a new child for validations instead of listing a number node with pant:message for xref validation error/warning. I am not sure the purpose of this new child node....
> 
> "validations": {
> "jcr:created": "Wed Apr 07 2021 14:44:48 GMT-0400",
> "jcr:createdBy": "pantheon",
> "jcr:primaryType": "sling:OrderedFolder",
> "validation_type": {
> "jcr:created": "Wed Apr 07 2021 14:44:48 GMT-0400",
> "jcr:createdBy": "pantheon",
> "jcr:primaryType": "sling:OrderedFolder"
> }
> }

That's intentional,
The idea is to have a separate distinguish validation node, through validation type i.e. in case, we have xref validation with other validations like Not null or link or document validation, we can separate them out. Although we have another field inside to check validation type. If you don't agree with me, We can ofcourse go ahead and revert to original structure.

xdavidson:
I see what you are doing. Adding a direct node under validations just for xrefs is a good idea. I do prefer the node name to be xrefs instead of calling git "validation_type".

rednitish:
> I see what you are doing. Adding a direct node under validations just for xrefs is a good idea. I do prefer the node name to be "xrefs" instead of calling it "validation_type".

Thanks Lisa, for pointing that out, I have updated the node name to xref. Please take a look.

xdavidson (approved):
LGTM

PR #576, opened 2021-04-19
 Xref Validations

rednitish:
1.) Delete xref nodes, if xref is reuploaded
				2.) Update Anchor based xrefs validatation 
				3.) Clone validation nodes from dependent modules to assembly


pantheon-bundle/src/main/java/com/redhat/pantheon/validation/helper/XrefValidationHelper.java
```diff
@@ -8,14 +8,19 @@
 
 public class XrefValidationHelper {
 
-    private static Map<String, List<String>> xRefs = new HashMap<>();
+    private static List<String> xRefs;
```
aprajshekhar, line 11: can't initialization be done here?

pantheon-bundle/src/main/java/com/redhat/pantheon/validation/helper/XrefValidationHelper.java
```diff
@@ -8,14 +8,19 @@
...
-    public static List<String> getObjectsToValidate(String uuid) {
-        return xRefs.get(uuid);
+    public static List<String> getObjectsToValidate() {
+        return xRefs;
     }
 
-    public static void setObjectsToValidate(String uuid, List<String> objectsToValidate) {
-        if(objectsToValidate.size()>0)
-            xRefs.put(uuid, objectsToValidate);
+    public static void initList() {
```
aprajshekhar, line 17: This might not be required if initialization and declaration are done together.

pantheon-bundle/src/main/java/com/redhat/pantheon/validation/helper/XrefValidationHelper.java
rednitish: @aprajshekhar in that case, if we have re upload scenarios, we have multiple duplicate elements corresponding to document, and validation nodes gets duplicated.

pantheon-bundle/src/main/java/com/redhat/pantheon/validation/helper/XrefValidationHelper.java
rednitish: we require list to initialize each time, to clear previous memory if any, for document,

xdavidson (approved):

PR #579, opened 2021-04-22
CCS-4351-bugfixes_NullPointerException

xdavidson:
This PR fixes the null pointer exception in XrefValidationHelper

### DIFF
--- a/pantheon-bundle/src/main/java/com/redhat/pantheon/auth/keycloak/KeycloakFilter.java
+++ b/pantheon-bundle/src/main/java/com/redhat/pantheon/auth/keycloak/KeycloakFilter.java
@@ -50,7 +50,6 @@ public class KeycloakFilter extends KeycloakOIDCFilter implements Filter {
     private static final Logger log = LoggerFactory.getLogger(KeycloakFilter.class.getName());
     private static final String KARAF_ETC = "karaf.etc";
     private static final String KEYCLOAKOIDCFILTER_CONFIG_FILE_NAME = "keycloak.json";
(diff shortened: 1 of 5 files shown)


benradey:
@xdavidson This is a good first step but I'm a bit worried about all the other fields and methods in the class being `static`. I think that's why we got away with the tests passing when they shouldn't have, and that could also occasionally bite us in non-obvious ways where 2+ threads and handling xrefs at the same time.

Could you remove all the `static` keywords from the class? Unfortunately I think that might take some refactoring in some other classes as well, but hopefully it would be minor. Let me know if you don't have the capacity, perhaps either I or @rednitish could lend a hand there.

xdavidson:
> @xdavidson This is a good first step but I'm a bit worried about all the other fields and methods in the class being `static`. I think that's why we got away with the tests passing when they shouldn't have, and that could also occasionally bite us in non-obvious ways where 2+ threads and handling xrefs at the same time.
> 
> Could you remove all the `static` keywords from the class? Unfortunately I think that might take some refactoring in some other classes as well, but hopefully it would be minor. Let me know if you don't have the capacity, perhaps either I or @rednitish could lend a hand there.

Thanks ben for the feedback! @rednitish is on PTO for a couple of weeks. I am helping him to get the xrefValidation for the upcoming release which is scheduled for next week. 

Your contribution is highly appreciated if you want to help us to get this out for the release.

benradey:
Looks perfect Lisa, thank you for putting in the time to fix this properly.

benradey (approved):

PR #582, opened 2021-04-23
Change text in Pv2 Help menu link from *Help* to *User guide*

theashiot:


theashiot:
Hi @benradey,

can you please review the changes?

best,

ashwin

pantheon-bundle/frontend/src/app/components/Chrome/Header/User.tsx
```diff
@@ -27,7 +27,7 @@ class User extends Component<IAppState, IState> {
 
     public render() {
         const dropdownItems = [
-            <DropdownItem key="help" href="/pantheon/docs/assemblies/assembly-pantheon-help.html" target="_blank">Help</DropdownItem>,
+            <DropdownItem key="User guide" href="/pantheon/docs/assemblies/assembly-pantheon-help.html" target="_blank">Help</DropdownItem>,
```
benradey, line 30:
```suggestion
            <DropdownItem key="userguide" href="/pantheon/docs/assemblies/assembly-pantheon-help.html" target="_blank">User Guide</DropdownItem>,
```
@theashiot Close! But the `key` attribute is used internally by React and is not user-facing. Given that, renaming it doesn't hurt, and is probably good practice to make it match closely with its intention (although I did remove the space for safety's sake given that this is used by the framework, but it probably didn't make a difference). The human-facing label that I think you're going after is the text contained within the tag itself, which comes later in this line.

I suggested a change; let me know if that aligns with your goal. Thanks!

benradey (changes requested):
@theashiot Thanks for your submission but I do have a tweak to suggest - let me know what you think!

theashiot:
Thanks @benradey, I applied your suggestion!

best,

ashwin

stoobie (dismissed):
LGTM!

stoobie:
@redhataccess/pantheon-developers From our side, this is ready for merge. If there's anything we need to do to correct the errors, please provide guidance.

benradey:
@stoobie @theashiot Drat! Yes, our frontend tests noticed the text is no longer what is expected and is flagging that as an error. I didn't think about that.

This is easy to fix is you have a local development environment set up, but I assume that you did not set up a development environment just to make this simple change, is that a safe assumption?

benradey:
@stoobie @theashiot In order to update what the tests expect, please update this file:

https://github.com/redhataccess/pantheon/blob/0948ecdedc75ff806391c409c3230106ca88b55d/pantheon-bundle/frontend/src/app/components/Chrome/Header/__snapshots__/User.test.tsx.snap#L12

Change the value on line 12 from "Help" to match your new preference. Then the test should pass.

There is an automated way to do that if you have a local dev setup, but for now, this is the easiest fix. Thanks!

theashiot:
Thanks @benradey, done!

benradey (approved):
Looks great, thanks for the submission!!

PR #583, opened 2021-04-28
FCCEUD-6.1 Adding new module to pantheon2.yml

stoobie:
@redhataccess/pantheon-developers I forgot to add a new file to pantheon2.yml. This PR adds it.

If all looks well technically, please merge.


benradey (approved):
@stoobie Your change looks good and the import does pick up the new file, but I think there's still the issue with a different file not being included in the repository that we discussed yesterday. I'll merge this in anyway since it's a step in the right direction.

PR #584, opened 2021-04-29
FCCEUD-54.1 Updated pantheon2.yml with filename change

stoobie:
@redhataccess/pantheon-developers Please review and if all looks OK, please merge.


stoobie:
@benradey I think this should take care of any issue with pantheon2.yml. If not, let me know and let's fix them in this PR.

PR #589, opened 2021-05-10
Ccs 3895 git import messaging

grantsydney:
When user imports git repo, alert user if import has failed or was successful.
Screen Shot 2021-05-10 at 5 04 32 PM
Screen Shot 2021-05-10 at 5 05 14 PM


pantheon-bundle/frontend/src/app/search.tsx
```diff
@@ -24,6 +24,8 @@ import { IAppState } from "@app/app"
 import { BulkOperationMetadata } from "./bulkOperationMetadata";
 import { BulkOperationPublish } from "./BulkOperationPublish"
 import { PathPrefixes } from "./Constants";
+import GitImportAlert  from "@app/GitImportAlert"
```
xdavidson, line 27: Why do we import GitImportAlert in search?

PR #594, opened 2021-05-17
CCS-3895-git-import-messaging-update

grantsydney:
handle error when status key is returned for git import but no master branch exists


pantheon-bundle/frontend/src/app/contexts/GitImportContext.tsx
```diff
@@ -120,12 +120,23 @@ export function GitImportProvider({ children }: IGitImportProviderProps) {
...
+            let singleUpload
+            if (json.status === "done") {
+              singleUpload = {
+                id: Date.now(),
+                totalFiles: json.total_files_uploaded,
+                repoName: repository,
+                success: true
+              };
+            }
+            if (json.status === "error") {
```
xdavidson, line 132: can we change line 132 to else if?

xdavidson (approved):

PR #596, opened 2021-05-18
CCS-4183  Deliver Pre-Live URLs

benradey:


pantheon-bundle/frontend/src/app/contentDisplay.tsx
```diff
@@ -109,15 +111,32 @@ class ContentDisplay extends Component<any, IModuleDisplayState | IAssemblyDispl
...
                         {this.state.lastPublishDate.trim() !== "" && this.state.lastPublishDate !== "-"
                             && this.state.variantUUID !== ""
                             && this.state.portalUrl !== ""
+                            && this.state.portalUrlType === "LIVE"
                             && <span><a id="permanentURL" onClick={this.copyToClipboard} onMouseLeave={this.mouseLeave}>Copy permanent URL  <CopyIcon /></a></span>
                         }
+                        {this.state.variantUUID !== ""
+                            && this.state.portalUrl !== ""
+                            && this.state.portalUrlType === "PRELIVE"
+                            && <span><a id="permanentURL" onClick={this.copyToClipboard} onMouseLeave={this.mouseLeave}>Copy pre-live URL  <CopyIcon /></a></span>
```
xdavidson, line 133: Does the pre-live URL look different from the permanent URL?

xdavidson: when there are more than one urls to copy, i think we are supposed to use a dropdown:
https://marvelapp.com/prototype/70e251a/screen/67615882
https://www.patternfly.org/v4/components/dropdown

benradey: The pre-live URL would likely not look any different. The only difference is the wording of the label shown through the UI, to remind authors that the URL would change if they change the product metadata.

We don't need to use a dropdown because there's only one URL to copy. The code is surrounded by conditionals and only one conditional will evaluate to true at a time.

I think that addresses your concerns, but let me know if there's something I missed.

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/sitemap/SiteMapServlet.java
```diff
@@ -116,11 +117,12 @@ private void writeXML(Resource resource, XMLStreamWriter xmlStream, SlingHttpSer
...
         // Process external url
         if (System.getenv(PORTAL_URL) != null) {
-            UrlProvider provider = new CustomerPortalUrlUuidProvider();
             documentVariant = resource.adaptTo(DocumentVariant.class);
 
             if (documentVariant != null && documentVariant.released().isPresent()) {
-                locPath = provider.generateUrlString(documentVariant);
+                try {
+                    locPath = new CustomerPortalUrlUuidProvider(documentVariant).generateUrlString();
+                } catch (UrlException e) {}
```
xdavidson, line 125: can we at least log the message for the exception?

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/util/ServletHelper.java
```diff
@@ -169,8 +170,12 @@ public static void setAssemblyData(Resource resource, List<HashMap<String, Strin
         }
         if (assemblyVariant.released().isPresent() && System.getenv(PORTAL_URL) != null) {
             // Add Customer Portal view_uri
-            String view_uri = new CustomerPortalUrlUuidProvider().generateUrlString(assemblyVariant);
-            assemblyVariantDetails.put("view_uri", view_uri);
+            try {
+                String view_uri = new CustomerPortalUrlUuidProvider(assemblyVariant).generateUrlString();
+                assemblyVariantDetails.put("view_uri", view_uri);
+            } catch (UrlException e) {
+                // TODO - add error stuff here
```
xdavidson, line 177: please consider log the message

pantheon-bundle/frontend/src/app/contentDisplay.tsx
xdavidson: if only one url is displayed, then I'm fine without the dropdown.

benradey:
@xdavidson I adding logging statements, what do you think?

xdavidson (approved):

PR #606, opened 2021-06-04
CCS-4453 Initial script to populate drupal cache

benradey:


rednitish (commented):
Thanks Ben, It does the task beautifully and good for starting step. 
I have 2 inputs, let me know, if you agree or we can have separate ticket to target them.

1.  Once started, it took about 5 mins to complete the script. as request are independent, Can we make it multi thread, to have it completed faster?
2. Can we make more like standard python script with multi functions within it i.e.

def main():
    \t    getDrupalCache()
    \t   validateURLs()

if __name__ == "__main__":
     \t  main()

benradey:
@rednitish Thanks for your suggestions. I implemented your 2nd point, let me know what you think. For your 1st point, I agree that multithreading it would make it perform faster, but I don't think it's worth the effort. For as often as we'll run the script (probably once for the entire project), the amount of time potentially saved doesn't seem like it's worth the additional developer attention.

However, if we wind up having to do this exercise again at some unforeseen point in the future where we have thousands of published documents where the potential time savings is more substantial, then multithreading might be a very good idea in that scenario. So I would suggest making a ticket to capture your idea so we don't lose track of it. Does that sound reasonable to you?

rednitish (approved):
LGTM

PR #607, opened 2021-06-05
 feat: use first para as abstract

mikemckiernan:
Because writers update most ADOC files and specify
the first para as the abstract, just use the first
para as the abstract if it is not explicitly specified.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ sling-org-apache-sling-karaf-configs/.settings
 
 pantheon-karaf-dist/src/main/resources/etc/keystore/
 pantheon-karaf-dist/src/main/resources/etc/org.ops4j.pax.web.cfg
+.m2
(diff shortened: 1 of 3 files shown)


pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/ModuleListingServlet.java
```diff
@@ -227,7 +227,7 @@ protected String getQuery(SlingHttpServletRequest request) {
 
     @Override
     protected Map<String, Object> resourceToMap(Resource resource) {
-        Module module = resource.adaptTo(Module.class);
+        com.redhat.pantheon.model.module.Module module = resource.adaptTo(com.redhat.pantheon.model.module.Module.class);
```
xdavidson, line 230: Do we really need make this code change?

xdavidson (commented):
Thank you for your contribution on the abstract!

pantheon-bundle/src/main/java/com/redhat/pantheon/servlet/ModuleListingServlet.java
mikemckiernan: Thanks @xdavidson  and @carlosmunoz.  I changed to use Java 8 and reverted the change to this file.

I'm happy to make any other changes. Thanks again for the help.

xdavidson (approved):

PR #616, opened 2021-07-05
CCS-4453 - Improvement on existing code

rednitish:
Added support for providing environment as argument to help testers/devs have it working on other environments.

Added print usages instructions for better understanding of script.

### DIFF
--- a/scripts/populateDrupalCache.py
+++ b/scripts/populateDrupalCache.py
@@ -22,13 +22,19 @@ env = dict({
             'https://pantheon.corp.dev.redhat.com/api/sitemap/module.sitemap.xml'
         ]})
 
-headers = {'user-agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/50.0.2661.102 Safari/537.36'}
+proxies = {}
 
 def gather_urls(user_input):
     urls = []
     sitemaps = env.get(user_input)
+    if sitemaps is None:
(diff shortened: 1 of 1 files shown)


scripts/populateDrupalCache.py
```diff
@@ -2,16 +2,33 @@
...
+            'https://pantheon.corp.stage.redhat.com/api/sitemap/assembly.sitemap.xml',
+            'https://pantheon.corp.stage.redhat.com/api/sitemap/module.sitemap.xml'
+        ],
+        "dev" : [
+            'https://pantheon.corp.dev.redhat.com/api/sitemap/assembly.sitemap.xml',
+            'https://pantheon.corp.dev.redhat.com/api/sitemap/module.sitemap.xml'
+        ]})
 
-def gather_urls():
+headers = {'user-agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/50.0.2661.102 Safari/537.36'}
```
aprajshekhar, line 25: Query: Do we need headers?

aprajshekhar:
LGTM. Just one query.

scripts/populateDrupalCache.py
rednitish: @aprajshekhar It is not helping us, i kept to to get rid off forbidden 403. But on preprod, we are still getting it. So we can remove it.

aprajshekhar (approved):
LGTM

PR #619, opened 2021-07-14
CCS-4366-lock-down-basic-auth

xdavidson:
This PR introduces the follow change:
  * redirect users to sso on /starter.html


rednitish (approved):
LGTM

PR #622, opened 2021-08-10
CCS-4579: Allow xref-ids to wrap so they can be copied

wesruv:
https://issues.redhat.com/browse/CCS-4579

Allow xref id's to wrap so they can be copied. I'm not sure how to get this UI for testing, but once it is you should be able to see the entire xref and it should be selectable/copyable.

Should look like this:
image


xdavidson (approved):
Tested locally. The ccs fixes look good to me.

